PR #11, opened 2017-09-07
"workers" auto setup

uprt:
Added the feature of setting the number of workers for the application
to the number of CPU cores (if "workers" parameter == 0)


uprt:
As I can see, currently there is no default value for this parameter - if we make a json config without "workers" field, unitd will throw "Invalid configuration" error.

PR #18, opened 2017-09-08
Chinese Translate

tuzimoe:
Adding translation for Chinese


nshadrin:
Thank you for this contribution!

It is great to see the localization of the documentation available so quickly.

However, we are unable to place and track all the localized versions of the readme in the main repository right now. I hope this changes in future.

Today, we would be happy to link to the README_CN.md file or to your website.

Please let us know how do you want to see the link from the main README: to your github repo, or to some other website that hosts this readme?

tuzimoe:
Yeah, I would like to keep translate the README for the NGINX Unit.  
 I also will keep updating the readme in the future.   
The Chinese Doc will be available in https://github.com/tuzimoe/unit/blob/master/README_CN.md  
Cheers.

PR #22, opened 2017-09-09
Define include paths of CFLAGS for cgo

ReSTARTR:
This change enables to...

* `go get` and import from "github.com/nginx/unit/src/go/unit" without setting another GOPATH.
* build specific version of `nginx/unit` with a package management tool.

```go
package main

import (
	"fmt"
	"net/http"

	"github.com/nginx/unit/src/go/unit" // replace from "unit"
)

func main() {
	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		fmt.Fprintf(w, "Hello, Go")
	})
	unit.ListenAndServe(":8080", nil)
}
```

### DIFF
--- a/src/go/unit/unit.go
+++ b/src/go/unit/unit.go
@@ -6,6 +6,7 @@
 package unit
 
 /*
+#cgo CFLAGS: -I./../../ -I./../../../build
 #include "nxt_go_lib.h"
 */
 import "C"


mar0x:
Good attempt. The problem is this fix does not enables unit go package build 'out-of-the-box': headers files produced by ./configure, ./configure go, and 'make go' are required.
Current method to build the unit go package is the following commands sequence:
assume GOPATH is set
./configure
./configure go
make go-instal

ReSTARTR:
Wow, That's true...

Mmm... How about copying `build/*` and `src` in Makefile? 
(at https://github.com/nginx/unit/blob/0.1/auto/modules/go#L111-L116)
It keeps structures the same as this repository.

With it and https://github.com/nginx/unit/pull/22/commits/5e552b6364be7563a00bc5ebadb4b134fbdd5ee3, even if GOPATH is changed to another path, we might able to `make go-install`. And we might also use `go get`.

mar0x:
Even with copying generated files in Makefile, command sequence 'go get', 'go build' will not work. 2 configure calls required before build.

Adding generated files to repository is not a good idea.

mar0x:
Go module relies on files, generated by Unit configure script, so it cannot be obtained using convenient "go get" command.

PR #32, opened 2017-09-13
Update README.md

nginx-gists:
- In the "Configuration > Applications" and "Configuration > Listeners" sections, removing indentation from the code blocks. In the "Listeners" section, the "double-marking" of indentation plus enclosing backticks (```) was causing the backticks to appear in the output. This was not happening in the "Applications" section -- I don't understand why not, as the markup looked the same to me in both sections, but that is a separate issue.

- In "Configuration > Listeners" section, changing the hyperlink to reflect the change to the title of the target section:

     [JSON Specification for Listener and Application Objects](#json-specification-for-listener-and-application-objects)

   changes to

     [Listener and Application Objects](#listener-and-application-objects)

### DIFF
--- a/README.md
+++ b/README.md
@@ -469,19 +469,19 @@ This example runs 20 workers of the PHP application named **blogs** using the
 files found in the **/www/blogs/scripts** directory. The default launch file
 when the URL doesn't specify the PHP file is **index.php**.
 
-   ```
-   {
-        ...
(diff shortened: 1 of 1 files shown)


VBart:
Thank you. The documentation was moved to a separate repository and these issues were fixed.
https://github.com/nginx/unit-docs

PR #55, opened 2017-10-17
Initial support for Ruby Rack applications

montekki:
Hi all. Decided to send in the patchset so the code doesn't get uselessly lost. A WIP now, code is somewhat hairy, but workable towards production.

Things not done yet:

* correct ruby installation detection similar to php-config. Ruby has no such script, so code now defaults to 2.3.0 ubuntu build and may be configured with  ./configure ruby --lib-path=/usr/lib/x86_64-linux-gnu/
* support for different and antique ruby versions (say, 1.9.0)
* not all rack variables are set
* etc.

Tested with the following config file:

```javascript
{
    "listeners" : {
        "*:8300" : {
            "application" : "rack"
        }
    },
    "applications": {
        "rack" : {
            "type": "ruby",
            "workers" : 1,
            "rack_root" : "$HOME_PATH/unit/",
            "rack_script": "app.ru"
        }
    }
}

```

AB testing resulted in

```
Document Path:          /
Document Length:        0 bytes

Concurrency Level:      1000
Time taken for tests:   0.430 seconds
Complete requests:      10000
Failed requests:        0
Total transferred:      820000 bytes
HTML transferred:       0 bytes
Requests per second:    23279.15 [#/sec] (mean)
Time per request:       42.957 [ms] (mean)
Time per request:       0.043 [ms] (mean, across all concurrent requests)
Transfer rate:          1864.15 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    5   3.9      3      20
Processing:     3   32  69.9     11     406
Waiting:        2   30  70.3      8     405
Total:          6   37  71.7     20     416

```

Let's talk fixes and I will be adding them to this branch.

### DIFF
--- a/auto/help
+++ b/auto/help
@@ -45,4 +45,7 @@ cat << END
   go OPTIONS           configure Go module
                        run "./configure go --help" to see available options
 
+  ruby OPTIONS         configure Ruby module
+                       run "./configure ruby --help" to see available options
+
 END
--- a/auto/make
+++ b/auto/make
@@ -234,6 +234,9 @@ ${NXT_DAEMON}-uninstall:
 
 END
 
+if [ $NXT_RUBY_MODULE != NO ]; then
+    . auto/modules/ruby/make
+fi
 
 # Makefile.
 # *.dSYM is MacOSX Clang debug information.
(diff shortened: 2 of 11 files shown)


VBart:
Implemented in https://github.com/nginx/unit/commit/37051b6c15cce7d6ab01c50e1086f8ef0b34e93d.

Issue #91, opened 2018-04-02
Ruby Apps Using Bundler Fail Due to nxt_ruby Loading Rack


sshaw:
Often times one will have version X of Rack installed but their application depends on version Y. This dependency on Y is usually handled via [Bundler](http://bundler.io/). 

Give a `config.ru` file for an application that uses Bundler, one would execute:

```
bundle exec rackup config.ru
```

To load version Y only instead of: 

```
rackup config.ru
```

Which may load version X before the Bundler code in the application can load version Y. When Bundler does attempt to load version Y, the following fatal error is given: 

```
2018/04/01 22:03:04 [alert] 4648#18991245 Ruby: /Users/sshaw/.rvm/gems/ruby-2.3.5/gems/bundler-1.15.4/lib/bundler/runtime.rb:317:in `check_for_activated_spec!': You have already activated rack 2.0.3, but your Gemfile requires rack 1.6.4. Prepending `bundle exec` to your command may solve this. (Gem::LoadError)
```

This is a problem with Unit because it [requires Rack](https://github.com/nginx/unit/blob/856826f535b662085fc563a02df2388d287ff374/src/ruby/nxt_ruby.c#L205) in order to create the application's entry point. As seen above, this version of Rack is not supported by the application it's trying to run.

sshaw:
May just want to check for a `Gemfile` in the application's home directory and do the bundler stuff prior to parsing `config.ru`. This should not affect the Bundler calls made by the application. 

I can put together a PR with this if you like.

lexborisov:
Hi @sshaw ,
I'm doing this task. You can send your PR and we will find the right solution.

Thanks!

sshaw:
@lexborisov checkout #93

jfqd:
Is there a solution for this issue? Every rack app has Gemfile, so is using bundler.

VBart:
@jfqd This ticket is a bit obsolete, since Unit uses bundler now. Could you elaborate more what's the issue is?

jfqd:
@VBart Thx for asking.

I am using rvm (Ruby Version Manager) and tried to push a config for a rails app (with Gemfile) on macOS (with homebrew):

```
curl -X PUT --unix-socket /usr/local/var/run/unit/control.sock http://localhost/config -d '{
    "listeners": {
        "*:3000": {
            "pass": "applications/myapp"
        }
    },
    "myapp": {
        "myapp": {
            "type": "ruby",
[...]

PR #93, opened 2018-04-06
Check for Gemfile in Ruby apps

sshaw:
RE #91. WIP. 

The code in `nxt_ruby.c` seems to prefer using Ruby C API calls over libc calls and calling methods via `rb_const_get` + `rb_funcall` instead of calling their C equivalents so this is what I've done. 

Additional code is needed as currently [this does not get executed](https://github.com/bundler/bundler/blob/257fb54da0003c3a67f6e7b3b5a242a4ba9c45cb/lib/bundler/setup.rb#L5). 

Possibly two options to fix: 

1. `chdir` do `script`'s root directory
1. Set the `BUNDLE_GEMFILE` environment variable 

I can investigate more later today. 

I will add tests once we agree on approach and coding style 😄

### DIFF
--- a/src/ruby/nxt_ruby.c
+++ b/src/ruby/nxt_ruby.c
@@ -30,6 +30,7 @@ static VALUE nxt_ruby_init_basic(VALUE arg);
 static nxt_int_t nxt_ruby_init_io(nxt_task_t *task);
 static VALUE nxt_ruby_rack_init(nxt_ruby_rack_init_t *rack_init);
 
+static VALUE nxt_ruby_require_bundler(VALUE arg);
 static VALUE nxt_ruby_require_rubygems(VALUE arg);
 static VALUE nxt_ruby_require_rack(VALUE arg);
 static VALUE nxt_ruby_rack_parse_script(VALUE ctx);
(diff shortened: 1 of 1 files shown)


src/ruby/nxt_ruby.c
```diff
@@ -191,6 +192,24 @@ nxt_ruby_init_io(nxt_task_t *task)
...
+nxt_ruby_bundler_project(nxt_str_t *script)
+{
+  VALUE file, dirname, path;
+
+  path = rb_str_new((const char *) script->start, (long) script->length);
+
+  file = rb_const_get(rb_cObject, rb_intern("File"));
+  dirname = rb_funcall(file, rb_intern("dirname"), 1, path);
+
+  rb_str_free(path);
```
sshaw: Is this necessary?

sshaw:
> In config we can set working_directory (chdir), some like a:

The server should be able to figure this out. For a user I think this is a bit annoying having to set this. One would only have to set this if `script` lived outside the working directory. I don't think this is a valid use case when it comes to Rack apps. 

It makes sense to add this to `nxt_ruby_rack_init_t`.

> I added little changes. Try this path:

I will apply the patch tomorrow but key thing is still missing: 

 `chdir` to `working_dir` or set the BUNDLE_GEMFILE. I don't know enough about unit to know what problems a `chdir` may cause, so I guess that leaves `BUNDLE_GEMFILE`.

sshaw:
And patch seg faults.

lexborisov:
I forgot to make `qrefresh` before export patch. I updated the patch.

We use `working_directory` for all modules. Actually, [`working_directory`](https://github.com/nginx/unit/blob/2f3b9d458328c7c96f91259c1892810c16cbccdd/src/nxt_application.c#L330) is the same as `chdir`.
I'll think about using the `BUNDLE_GEMFILE` environment variable.

romanm-odysii:
Is anything that prevents from this PR merged into master?

VBart:
@romanm-odysii Please, see the above discussion. There's no consensus about how to configure a bundler gemfile and the patch isn't complete.

PR #177, opened 2018-10-24
Only build PHP and Python on Raspbian

mpmc:
Pretty much as title. The other langs aren't available OOTB on Raspbian.


defanator:
@mpmc thanks for your efforts. We don't consider Raspbian as one of primary platforms to support, so we'd like to avoid including any specific hacks around it in the base tree.

However, you always can build separate packages by simply running e.g. `make unit unit-php unit-python` inside `pkg/deb` and use the resulting binaries.

PR #211, opened 2019-01-15
Fedora >= 29 have python 3.7

remicollet:


remicollet:
Notice: make rpm still fail in unit-go build... but this is another issue #214

defanator:
@remicollet this one has just been merged with a minor change to commit log, thanks!

PR #215, opened 2019-01-16
prefer system crypto policy

remicollet:
Mandatory for official repository
See
* https://fedoraproject.org/wiki/Changes/CryptoPolicy
* https://docs.fedoraproject.org/en-US/packaging-guidelines/CryptoPolicies/

Perhaps better to have a configure option for this (like the --with-system-ciphers of PHP°

### DIFF
--- a/src/nxt_openssl.c
+++ b/src/nxt_openssl.c
@@ -295,7 +295,7 @@ nxt_openssl_server_init(nxt_task_t *task, nxt_mp_t *mp,
     nxt_tls_init_t *tls_init, nxt_bool_t last)
 {
     SSL_CTX                *ctx;
-    const char             *ciphers, *ca_certificate;
+    const char             *ca_certificate;
     nxt_tls_conf_t         *conf;
     STACK_OF(X509_NAME)    *list;
     nxt_tls_bundle_conf_t  *bundle;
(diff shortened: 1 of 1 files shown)


remicollet:
To summarize, it is easier for a sysadmin to alter a configuration than to change an hardcoded value in sources ;)

remicollet:
More explanation about why this change is needed
https://www.redhat.com/en/blog/consistent-security-crypto-policies-red-hat-enterprise-linux-8

src/nxt_openssl.c
```diff
@@ -303,13 +303,13 @@ nxt_openssl_server_init(nxt_task_t *task, nxt_tls_conf_t *conf)
...
+    if (conf->ciphers) { /* else use system crypto policy */
+        if (SSL_CTX_set_cipher_list(ctx, conf->ciphers) == 0) {
+            nxt_openssl_log_error(task, NXT_LOG_ALERT,
                               "SSL_CTX_set_cipher_list(\"%s\") failed",
-                              ciphers);
-        goto fail;
+                              conf->ciphers);
+            goto fail;
+        }
     }
```
alejandro-colomar, line 313: Hi Remi!

Please detail in the commit message the mechanism by which system crypto policy is being used in the "else" case.  Is it by not calling `SSL_CTX_set_cipher_list()`, or how?

alejandro-colomar:
If this PR is still necessary, please rebase it against the master branch.

remicollet:
Rebased

src/nxt_openssl.c
remicollet: If you don't call `SSL_CTX_set_cipher_list`, then you use system default. The goal of this PR

alejandro-colomar:
Looks good to me.  Please include something like `"If you don't call SSL_CTX_set_cipher_list, then you use system default. The goal of this PR"` in the commit message, and sign the patch (signed-off-by), and I'll merge it.

Thanks!

alejandro-colomar:
BTW, it would also be nice if you add a changelog line into `docs/changes.xml`.

alejandro-colomar:
Great.  I'll push it in a moment.

Thanks!

alejandro-colomar:
Merged.

Issue #109, opened 2018-04-19
16#48 *11027 shutdown(14, 2) failed (107: Transport endpoint is not connected)


urba1n:
With Unit Python, I get the following messages at almost each application call :
_Apr 19 16:29:48 unit_as - 2018/04/19 16:29:48 [info] 16#66 *199 shutdown(14, 2) failed (107: Transport endpoint is not connected)
Apr 19 16:29:48 unit_as - 2018/04/19 16:29:48 [info] 16#44 *160 shutdown(14, 2) failed (107: Transport endpoint is not connected)
Apr 19 16:29:48 unit_as - 2018/04/19 16:29:48 [info] 16#64 *202 shutdown(15, 2) failed (107: Transport endpoint is not connected)
Apr 19 16:29:48 unit_as - 2018/04/19 16:29:48 [info] 16#69 *203 shutdown(14, 2) failed (107: Transport endpoint is not connected)
Apr 19 16:29:49 unit_as - 2018/04/19 16:29:49 [info] 16#71 *211 shutdown(15, 2) failed (107: Transport endpoint is not connected)
Apr 19 16:29:49 unit_as - 2018/04/19 16:29:49 [info] 16#68 *206 shutdown(15, 2) failed (107: Transport endpoint is not connected)
Apr 19 16:29:49 unit_as - 2018/04/19 16:29:49 [info] 16#71 *205 shutdown(14, 2) failed (107: Transport endpoint is not connected)_

igorsysoev:
This means that client already closed connection.
We should change log level to debug for this error in shutdown() or close() syscalls.

PR #227, opened 2019-03-04
Change log level to debug for socket shutdown issue #109

urba1n:
Log level set to debug when shutting down a socket that has already been closed by client connexion end.


VBart:
This PR on Github has been forgotten, while the suggested change was independently committed in: https://hg.nginx.org/unit/rev/6cf42ea4b90e

Sorry about that.

PR #298, opened 2019-08-26
switch to __BYTE_ORDER__ (and other underscored endianness macros)

tiago4orion:
The macros `BYTE_ORDER`, `BIG_ENDIAN` and `LITTLE_ENDIAN` are not defined on Solaris.

This patch change them to `__BYTE_ORDER__`, `__ORDER_BIG_ENDIAN__` and `__ORDER_LITTLE_ENDIAN__`. 

These flags are defined on:

  - Linux
  - OSX
  - Solaris
  - FreeBSD

Closes #297 

Signed-off-by: Tiago Natel de Moura <tiago.moura@nginx.com>

PR #299, opened 2019-08-26
OpenSolaris: Fix sendfile detection and use O_NONBLOCK instead of FIONBIO

tiago4orion:
The configure script was defining `NXT_HAVE_LINUX_SENDFILE` for solaris, but the linux sendfile implementation uses the linux-only `MSG_MORE` flag (on send(2)) instead of `TCP_CORK` sockopt. I didn't test on Solaris 11 but on OpenSolaris.

Second bug is that unit was getting an error when setting the signals file descriptor to nonblocking. The error was: `Inappropriate ioctl for device`
I arranged the ifdefs to use fnctl and O_NONBLOCK on solaris. 

I don't know about the performance impact of second change but maybe it's good to make it compile and run for now. I'm willing to do some tests with Solaris Zones soon.

This change depends on #298

### DIFF
--- a/auto/sendfile
+++ b/auto/sendfile
@@ -10,29 +10,28 @@ NXT_HAVE_SOLARIS_SENDFILEV=NO
 NXT_HAVE_AIX_SEND_FILE=NO
 NXT_HAVE_HPUX_SENDFILE=NO
 
+# Solaris 8 sendfilev().
 
-# Linux sendfile().
+NXT_LIBSENDFILE=
 
-nxt_feature="Linux sendfile()"
-nxt_feature_name=NXT_HAVE_LINUX_SENDFILE
+nxt_feature="Solaris sendfilev()"
+nxt_feature_name=NXT_HAVE_SOLARIS_SENDFILEV
+nxt_feature_libs="-lsendfile"
 nxt_feature_test="#include <sys/sendfile.h>
-                  #include <sys/types.h>
-                  #include <sys/socket.h>
 
-                  int main() {
-                      off_t  offset;
+                    int main() {
+                        size_t              sent;
+                        struct sendfilevec  vec;
 
-                      sendfile(-1, -1, &offset, 0);
-                      send(-1, NULL, 0, MSG_MORE);
-
-                      return 0;
-                  }"
+                        sendfilev(-1, &vec, 0, &sent);
+                        return 0;
+                    }"
 . auto/feature
 
 if [ $nxt_found = yes ]; then
-    NXT_HAVE_LINUX_SENDFILE=YES
+    NXT_HAVE_SOLARIS_SENDFILEV=YES
+    NXT_LIBSENDFILE=$nxt_feature_libs
 fi
-
 
 if [ $nxt_found = no ]; then
     # FreeBSD sendfile().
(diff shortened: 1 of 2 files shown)


src/nxt_file.c
```diff
@@ -312,7 +312,7 @@ nxt_file_rename(nxt_file_name_t *old_name, nxt_file_name_t *new_name)
  * Linux 2.6 does not use BKL.
  */
 
-#if (NXT_HAVE_FIONBIO)
+#if defined(NXT_HAVE_FIONBIO) && !defined(NXT_SOLARIS)
```
VBart: ```#if (NXT_HAVE_FIONBIO) && !(NXT_SOLARIS)```
All `NXT_*` macros should be cheked this way as we want to have an ability to unset them with `-DNXT_HAVE_FIONBIO=0`.

src/nxt_file.c
```diff
@@ -312,7 +312,7 @@ nxt_file_rename(nxt_file_name_t *old_name, nxt_file_name_t *new_name)
  * Linux 2.6 does not use BKL.
```
VBart: The comment here needs to be updated.

auto/sendfile
```diff
@@ -16,11 +16,15 @@ NXT_HAVE_HPUX_SENDFILE=NO
...
 nxt_feature_name=NXT_HAVE_LINUX_SENDFILE
 nxt_feature_test="#include <sys/sendfile.h>
+                  #include <sys/types.h>
+                  #include <sys/socket.h>
 
                   int main() {
                       off_t  offset;
 
                       sendfile(-1, -1, &offset, 0);
+                      send(-1, NULL, 0, MSG_MORE);
```
VBart: This test will be passing again as soon as Solaris will get MSG_MORE support. In general it's not a good idea to workaround the test with checking something unrelated.

It's better to reorder the tests. We want to use sendfilev() where it's available, then FreeBSD sendfile() and only then Linux (as it has the less powerful interface).

tiago4orion: Ok, I thought it was the other way around as linux sendfile is tested first.

src/nxt_file.c
tiago4orion: What do you mean? Document the behavior on newer kernels?

VBart: I meant that we should add to this comment a line about why `&& !(NXT_SOLARIS)` has been added.

tiago4orion: I updated, take a look.

auto/sendfile
```diff
@@ -156,3 +131,24 @@ if [ $nxt_found = no ]; then
         NXT_HAVE_HPUX_SENDFILE=YES
     fi
 fi
+
+# Linux sendfile().
```
VBart: This comment similar to other tests in this file should be inside the `if [ $nxt_found = no ]; then` block.

src/nxt_file.c
```diff
@@ -310,9 +310,13 @@ nxt_file_rename(nxt_file_name_t *old_name, nxt_file_name_t *new_name)
...
  * Linux 2.4 uses BKL for ioctl() and fcntl(F_SETFL).
  * Linux 2.6 does not use BKL.
+ * 
+ * On Solaris, the FIONBIO ioctl command is not allowed on a signalfd
+ * file descriptor. Ref.: 
+ * https://github.com/illumos/illumos-gate/blob/master/usr/src/uts/common/io/signalfd.c#L623
  */
 
-#if (NXT_HAVE_FIONBIO)
+#if (NXT_HAVE_FIONBIO) && !(NXT_SOLARIS)
```
VBart: I've just talked to Igor, and it seems that this problem is caused by the ordering in `https://github.com/nginx/unit/blob/master/src/nxt_service.c`. Because regular Solaris system doesn't support epoll. It's expected that Solaris event port will be used instead. All these problems seem specific to illumos.

tiago4orion: Ok, I just got a proper Solaris 11.4 box and it seems this indeed is not needed there. But the other changes apply (the byte_order needs work and also the rearrange of the sendfile detection).

tiago4orion:
I'm closing this for now. If any user complains I can work on this again.

PR #289, opened 2019-08-16
Add namespace isolation

tiago4orion:
Adds an `isolation` field to app config for namespace process and sandbox related settings.

- [x] isolation field
- [x] build detect ns flags
- [x] switch to clone on Linux
- [x] uid/gid remapping
- [x] uid/gid config

The `isolation` field was designed to be system-dependent. This PR adds Linux (or clone)  isolation:

```json
{
    "type": "external",
    "executable": "/bin/app",
    "isolation": {
        "namespaces": {
            "user": true,
            "mount": true
        }
    }
}
```
The list of Linux namespaces supported are: `user`, `pid`, `mount`, `network`, `ipc`, `uts`, `cgroup`.

All namespaces other than `user` requires root (or sys capability) if used alone. You can set `user` together with any other ns flags in order to use isolation from an unprivileged Unit process. 

The isolation config for linux improves a lot the security of a Unit web app deployment. For example, If an attacker exploits a PHP vulnerability, he/she can just see the namespaces configured for the app. If it's completed jailed, then the only way to reach the host namespace is by exploiting the Linux kernel.

*For now, every app needs to share the ipc namespace with Unit. This is required for the shared memory configuration.* That means the `ipc` flag is blocked.

In addition, this PR also adds a config for uid/gid remap. When using `user` namespace, you can map a range of user ids in the container namespace to a range of ids in the parent namespace. Then, for example, you can create a container from an unprivileged user and map this user to uid 0 (root) inside the container. The uid 0 inside the container has the full set of capabilities on that namespace but its capabilities in the parent are restricted to that of the parent process uid.
The same applies for gid remap. The example below shows how to configure it:

```json
{
    "type": "external",
    "executable": "/bin/app",
    "isolation": {
        "namespaces": {
            "user": true
        },
        "uidmap": [
            {
                "containerID": 0,
                "hostID": 1000,
                "size": 1
            }
        ],
        "gidmap": [
            {
                "containerID": 0,
                "hostID": 1000,
                "size": 1
            }
        ]
    }
}
```

If no map is set, a default one binding root in the container to the current euid of the unit process is used.

In order for users to get real advantages of these settings, we need to add other general configs as "mounts" and "rootfs" fields to properly chroot the app in its own root filesystem.

Operating systems and architectures tested:

**Ubuntu 18.04 (x86_64)**
```
[...]

PR #301, opened 2019-08-27
Add capability check

tiago4orion:
Add capability check for the ability to change user and group (setuid and setgid). 

There's a specific implementation for operating systems that support some kind of fine-grained control for process permissions.

- Linux capability (portable across v1, v2 and v3)
- Solaris privileges

Other operating systems require root.

In order to make possible to run Unit (with some limitation) even if root or privileges is not an option, we allow running it with --user=$USER. This is the default if Unit starts without capability. In this mode, Unit starts workers but not try to change user. When running applications, it error if passing a "user" field (or "group") with a value different than the current process user.

Tested on:

- Linux Ubuntu 18.04 (64-bits)
- Linux Debian trusty 32-bits
- OpenSolaris / OpenIndiana 5.11


src/nxt_capability.c
```diff
@@ -0,0 +1,139 @@
...
+    return NXT_OK;
+}
+
+void
+nxt_capability_log_hint(nxt_task_t *task)
+{
+    nxt_log(task, NXT_LOG_INFO, "hint: In order to give the right capabilities"
+        " you can run Unit as root or use /sbin/setcap, eg.: "
+        "# setcap cap_setuid,cap_setgid=+ep %s",
+        *nxt_process_argv);
```
tiago4orion: I'm not happy with this long message. Can we just point to the documentation URL where we document this better to each OS? Something like:

```c
    nxt_log(task, NXT_LOG_INFO, "hint: https://unit.nginx.org/installation/#privs");
```

tiago4orion:
Below are some usage examples:

```
vagrant@ubuntu-bionic:/vagrant/src/unit$ echo $USER
vagrant
vagrant@ubuntu-bionic:/vagrant/src/unit$ id
uid=1000(vagrant) gid=1000(vagrant) groups=1000(vagrant)
```
Unprivileged:
```
vagrant@ubuntu-bionic:/vagrant/src/unit$ ./build/unitd --no-daemon
2019/08/27 12:42:47 [notice] 15661#15661 Unit is running unprivileged, then it cannot use arbitrary user and group. Using "vagrant" and "vagrant" for user and group, respectively.
2019/08/27 12:42:47 [info] 15661#15661 hint: In order to give the right capabilities you can run Unit as root or use /sbin/setcap, eg.: # setcap cap_setuid,cap_setgid=+ep ./build/unitd
2019/08/27 12:42:47 [notice] 15661#15661 initgroups(vagrant, 1000) failed (1: Operation not permitted), ignored
2019/08/27 12:42:47 [info] 15661#15661 unit started
^C
```
Non-root but with right privileges:
```
vagrant@ubuntu-bionic:/vagrant/src/unit$ sudo setcap cap_setuid,cap_setgid=+ep build/unitd
vagrant@ubuntu-bionic:/vagrant/src/unit$ /build/unitd --no-daemon
2019/08/27 12:47:47 [info] 15699#15699 unit started
```

Root:
```
vagrant@ubuntu-bionic:/vagrant/src/unit$ sudo ./build/unitd --no-daemon 
2019/08/27 12:44:24 [info] 15676#15676 unit started
^C
vagrant@ubuntu-bionic:/vagrant/src/unit$ 
```

tiago4orion:
@VBart I tested the Solaris privileges on Solaris 11.4 (after changes to build) and it works. But as we're not going to support it for now, do you want me to drop Solaris privileges support here?

src/nxt_conf_validation.c
```diff
@@ -1250,72 +1243,6 @@ nxt_conf_vldt_array_iterator(nxt_conf_validation_t *vldt,
     }
 }
 
-
```
tiago4orion: @VBart I removed the user/group validation at conf load-time because this will not be possible with isolation anymore.

tiago4orion:
Closing as this is merged in #289 already

PR #307, opened 2019-09-05
Use socket credential passing (SCM_CREDENTIALS / SCM_CREDS)

i4ki:
This patch adds support for `SCM_CREDENTIALS/SCM_CREDS` cmsg to avoid having to pass the `pid` explicitly in the msg payload. Also, add some IPC validations on obvious places. We need to review all comms in the future.

Tested on Linux, FreeBSD and OSX.

PR #308, opened 2019-09-06
Refactor nxt_unit.c to reuse nxt_sendmsg and nxt_recvmsg

tiago4orion:
This refactor is needed to avoid a lot of duplications in the process credentials work (#307).
Now we can just process the `SCM_CREDENTIALS/SCM_CREDS/SCM_RIGHTS` cmsg headers at the same place, the file `nxt_sockmsg.c`.

Another benefit of these changes, is that now we can properly use scatter-gather buffers in the application side as well, as the previous implementation depended upon the portable Go `unix.Recvmsg/unix.Sendmsg` functions that have the limitation of just only 1 iovec buffer.

I removed the `port_recv` and `port_send` callbacks from the API because I think it complicates the matter, as the portable Recvmsg/Sendmsg from Go stdlib doesn't support Msghdrs. My first approach was just using the Go `syscall` package in order to bypass Go interface but this way we'll have to maintain sendmsg/recvmsg on each arch and operating system.

### DIFF
--- a/src/nxt_sockmsg.c
+++ b/src/nxt_sockmsg.c
@@ -1,0 +1,145 @@
+/*
+ * Copyright (C) Igor Sysoev
+ * Copyright (C) NGINX, Inc.
+ */
+
+#include <nxt_main.h>
+#include "nxt_sockmsg.h"
+
+#if (NXT_HAVE_MSGHDR_MSG_CONTROL)
+
+/*
+ * Linux, FreeBSD, Solaris X/Open sockets,
+ * MacOSX, NetBSD, AIX, HP-UX X/Open sockets.
+ */
+
+ssize_t
+nxt_sendmsg(nxt_socket_t s, nxt_fd_t fd, nxt_iobuf_t *iob, nxt_uint_t niob)
+{
+    struct msghdr       msg;
+    union {
+        struct cmsghdr  cm;
+        char            space[CMSG_SPACE(sizeof(int))];
+    } cmsg;
+
+    msg.msg_name = NULL;
+    msg.msg_namelen = 0;
+    msg.msg_iov = iob;
+    msg.msg_iovlen = niob;
+    /* Flags are cleared just to suppress valgrind warning. */
+    msg.msg_flags = 0;
+
+    if (fd != -1) {
+        msg.msg_control = (caddr_t) &cmsg;
+        msg.msg_controllen = sizeof(cmsg);
+
+#if (NXT_VALGRIND)
+        nxt_memzero(&cmsg, sizeof(cmsg));
+#endif
+
+        cmsg.cm.cmsg_len = CMSG_LEN(sizeof(int));
+        cmsg.cm.cmsg_level = SOL_SOCKET;
+        cmsg.cm.cmsg_type = SCM_RIGHTS;
+
+        /*
+         * nxt_memcpy() is used instead of simple
+         *   *(int *) CMSG_DATA(&cmsg.cm) = fd;
+         * because GCC 4.4 with -O2/3/s optimization may issue a warning:
+         *   dereferencing type-punned pointer will break strict-aliasing rules
+         *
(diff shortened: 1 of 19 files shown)


src/nxt_unit.h
```diff
@@ -137,14 +137,9 @@ struct nxt_unit_callbacks_s {
...
-    ssize_t  (*port_send)(nxt_unit_ctx_t *, nxt_unit_port_id_t *port_id,
-                 const void *buf, size_t buf_size,
-                 const void *oob, size_t oob_size);
-
-    /* Receive data on port id. Optional. */
-    ssize_t  (*port_recv)(nxt_unit_ctx_t *, nxt_unit_port_id_t *port_id,
-                 void *buf, size_t buf_size, void *oob, size_t oob_size);
+    /* Lookup the port pair from the port id. Optional. */
+    void     (*lookup_port_pair)(nxt_unit_ctx_t *, nxt_unit_port_id_t *port_id,
+                                int *in_fd, int *out_fd);
```
tiago4orion: Maybe the prototype below is better:

```c
void (*lookup_port_pair)(nxt_unit_ctx_t *, nxt_unit_port_id_t *port_id, int pair[2]);
```

Let me know what you think.
@VBart @mar0x

src/nxt_unit.h
```diff
@@ -190,8 +185,8 @@ nxt_unit_ctx_t *nxt_unit_init(nxt_unit_init_t *);
  * from port socket should be initially processed by unit.  This function
  * may invoke other application-defined callback for message processing.
  */
-int nxt_unit_process_msg(nxt_unit_ctx_t *, nxt_unit_port_id_t *port_id,
-    void *buf, size_t buf_size, void *oob, size_t oob_size);
+int nxt_unit_process_msg(nxt_unit_ctx_t *ctx, nxt_port_msg_t *port_msg,
```
tiago4orion: I would like to rewrite this function as:

```
int nxt_unit_process_msg(nxt_unit_ctx_t *ctx, nxt_unit_recv_msg_t *msg, nxt_port_msg_t *port_msg)
```

But doing that requires exporting some types in `nxt_unit.h`. 
They are here: https://github.com/nginx/unit/blob/master/src/nxt_unit.c#L27

src/nxt_sockmsg.c
```diff
@@ -0,0 +1,145 @@
...
+        /* (*fd) = *(int *) CMSG_DATA(&cmsg.cm); */
+        nxt_memcpy(fd, CMSG_DATA(&cmsg.cm), sizeof(int));
+    }
+
+    return n;
+}
+
+#else
+
+/* Solaris 4.3BSD sockets. */
```
tiago4orion: @VBart Can I get rid of Solaris support here? I just copy-pasted these functions from `nxt_socketpair.c`.

src/nxt_unit.h
tiago4orion: I reverted these changes. My goal was just to handle messages in the same way as Unit does on its side:
https://github.com/nginx/unit/blob/master/src/nxt_port_socket.c#L619

But doing this requires some re-design in the unit lib.

src/nxt_unit.h
mar0x: I think we should avoid blocking in ListenAndServe() function.  In spite of the fact cgo function calls slower than pure Go function, excessive cgo calls looks better than locking the thread with a Unit wait message loop.
For code reuse, would be nice to investigate if it is possible to poll the fd in Go without reading, and then read and process message in C - this way we can reuse nxt_readmsg().

src/nxt_sockmsg.c
VBart: Yes, you can.

src/nxt_unit.h
tiago4orion: Ok, I'm closing this as it requires a better solution.
Thanks

PR #310, opened 2019-09-16
WIP: Add details to process startup errors

tiago4orion:
This PR improves the error messages for errors happening after worker starts but before the READY message is sent.

I think this is a must in order to have meaningful error messages for features like `chroot` and `isolation`.

Example:

```
$ curl --unix-socket control.unit.sock -XPUT http://localhost/config/applications/py --data-binary @app-python.json
{
	"error": "Failed to apply new configuration.",
	"detail": "chdir(/app) failed (2: No such file or directory)"
}
```

_Work in progress_

### DIFF
--- a/src/nxt_application.c
+++ b/src/nxt_application.c
@@ -17,7 +17,6 @@
 
 #include <glob.h>
 
-
 typedef struct {
     nxt_app_type_t  type;
     nxt_str_t       version;
@@ -35,7 +34,8 @@ static void nxt_discovery_quit(nxt_task_t *task, nxt_port_recv_msg_t *msg,
 static nxt_app_module_t *nxt_app_module_load(nxt_task_t *task,
     const char *name);
 static nxt_int_t nxt_app_set_environment(nxt_conf_value_t *environment);
-
+void nxt_app_error(nxt_task_t *task, nxt_str_t *err, u_char *buf,
+    const char *fmt, ...);
 
 static uint32_t  compat[] = {
     NXT_VERNUM, NXT_DEBUG,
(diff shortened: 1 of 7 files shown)


src/nxt_process.c
```diff
@@ -136,19 +138,25 @@ nxt_process_start(nxt_task_t *task, nxt_process_t *process)
...
     if (init->user_cred != NULL) {
         /*
          * Changing user credentials requires either root privileges
          * or CAP_SETUID and CAP_SETGID capabilities on Linux.
          */
         ret = nxt_user_cred_set(task, init->user_cred);
         if (ret != NXT_OK) {
-            goto fail;
+            err.start = (u_char *) "failed to change user credentials";
+            err.length = nxt_strlen(err.start);
```
VBart: nxt_str_set()

src/nxt_application.c
```diff
@@ -343,30 +378,45 @@ nxt_app_start(nxt_task_t *task, void *data)
...
 
-    return ret;
+fail:
+    buf = nxt_buf_mem_alloc(rt->mem_pool,
+                            err.length+1, 0);
+
+    if (nxt_fast_path(buf != NULL)) {
+        buf->mem.free = nxt_cpymem(buf->mem.free, err.start,
+                                   err.length);
+        *buf->mem.free++ = '\0';
```
VBart: Why is this needed? In the handler function you do the opposite: `nxt_buf_mem_used_size(&msg->buf->mem) - 1;`.

tiago4orion: Because I send a NULL terminated string in the message buffer.

Here: https://github.com/nginx/unit/blob/870e8e0f867603c552f10b9cb8f6941c1cf59f3e/src/nxt_controller.c#L1507
I convert it back to an `nxt_str_t` for detail. I thought that nxt_str_t doesn't need the NULL terminator. Is that correct?

tiago4orion: I can also just send the string bytes without NUL, as the buffer already carry the size. Is it better?

VBart: > maybe I can just send the string bytes without NUL, as the buffer already carry the size. If this was what you meant, I think you are right, much simpler.

Yeap.

PR #316, opened 2019-09-20
Debian init script fixes

thresheek:
The first patch fixes the following issues:

     - inability to stop unit daemon
     - default configuration from /etc/default/unit are not propagated to the daemon

The second patch is purely cosmetical change since we include the defaults a few lines later as well.

### DIFF
--- a/pkg/deb/debian/unit.init
+++ b/pkg/deb/debian/unit.init
@@ -16,8 +16,6 @@ DAEMON=/usr/sbin/unitd
 NAME=unit
 DESC=unitd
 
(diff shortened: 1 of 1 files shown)


defanator:
@thresheek LGTM

thresheek:
Merged in https://github.com/nginx/unit/commit/919230a21e9d5f2b6edcf99e1938707d487639a2 and https://github.com/nginx/unit/commit/5c6b01d76c329f4ddcc31378e4ec62123e804d01

Issue #369, opened 2020-01-10
[patch] http route match optimization


hongzhidao:
A patch for http route match optimization in the case of large hosts and uris.

hongzhidao:
Here's the patch related to route match optimization.
https://gist.github.com/hongzhidao/b3427a563748444a6002430e9213c5e1

1. The final structure format.
```
# (host, uri)
route->hosts[$host].uris[$uri][] = match;
# (host, nil)
route->hosts[$host][] = match;
# (nil, url)
route->uris[$uri][] = match;
# (nil, nil)
route->matches[] = match;
```

2. Introduced key functions
```
route:
   host_hash: nxt_lvlhsh_t
   uri_tree: nxt_radix_tree_t

nxt_http_route_matches_add(route, host, uri, match);
nxt_http_route_matches_find(route, host, uri);  // return matches
```

3. WIP:
group matches by host.  (done)
group matches by uri.

hongzhidao:
It seems `variables` is a nice solution instead of `radix tree`.

PR #404, opened 2020-03-01
Group http route matches by host.

hongzhidao:
Hi.
Take a look again at this improvement of host routing.
All the tests passed.


hongzhidao:
See: https://github.com/nginx/unit/issues/369

PR #425, opened 2020-04-17
Fix build with PHP 8.0.0-dev

remicollet:
Build OK and module load fine (not real test yet)


VBart:
Thanks. The fix was committed: https://github.com/nginx/unit/commit/140b81208e83569913aa81f964eb64e15940d897

PR #490, opened 2020-11-06
fix:configure java module in alpine linux not working issue

wujjpp:
Alpine linux using musl libc instread of glibc, if we try to build java module on alpine Linux, we will find it not working, The patch is for resolving the issue

### DIFF
--- a/auto/modules/java
+++ b/auto/modules/java
@@ -313,7 +313,7 @@ NXT_JAVA_LIBJVM="$NXT_JAVA_LIB_SERVER_PATH/libjvm.so"
 if [ "$NXT_SYSTEM" = "Darwin" ]; then
 NXT_JAVA_LIBC_DIR="/usr/lib"
 else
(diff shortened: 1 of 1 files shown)


auto/modules/java
```diff
@@ -313,7 +313,7 @@ NXT_JAVA_LIBJVM="$NXT_JAVA_LIB_SERVER_PATH/libjvm.so"
 if [ "$NXT_SYSTEM" = "Darwin" ]; then
 NXT_JAVA_LIBC_DIR="/usr/lib"
 else
-NXT_JAVA_LIBC_DIR=`ldd "$NXT_JAVA_LIBJVM" | grep libc.so | cut -d' ' -f3`
+NXT_JAVA_LIBC_DIR=`ldd "$NXT_JAVA_LIBJVM" | grep -E "libc.so|libc.musl-x86_64.so" | cut -d' ' -f3`
```
mar0x, line 316: I think `grep -F libc.` is strict enough for this purpose.

wujjpp: @mar0x YES, simple and strict enough

Issue #317, opened 2019-09-21
pipe is not a function. NodeJs


Alexufo:
I get error when use multer https://www.npmjs.com/package/multer
TypeError: request.req.pipe is not a function

I am right it is the same problem?
https://github.com/nginx/unit/issues/274

VBart:
Yes, #274 is the same.

Issue #274, opened 2019-07-05
node.js http-proxy-middleware seems not working


freerider7777:
I'm trying to use 
http-proxy-middleware
for some endpoints

and a simple example from here (replacing http->unit-http) does not work:
https://www.npmjs.com/package/http-proxy-middleware
(uses http-proxy inside)

app.js:

```
#!/usr/bin/env node

const {
  createServer,
  IncomingMessage,
  ServerResponse,
} = require('unit-http');

var proxy = require('http-proxy-middleware');

require('http').ServerResponse = ServerResponse;
require('http').IncomingMessage = IncomingMessage;

const express = require('express')
const app = express();


app.get('/', (req, res) => {
    console.log('request /'); // Print the HTML for the Google homepage.
	res.send('Hello, Unit!')
});
 // app.get('/authenticate', (req, res) => res.send('Hello, Unit (authenticate)!'))

app.use('/api', proxy({ target: 'http://www.example.org', changeOrigin: true }))
createServer(app).listen();

```

Error (when calling /api):

TypeError: (options.buffer || req).pipe is not a function
    at Array.stream (/modules/proxy/node_modules/http-proxy/lib/http-proxy/passes/web-incoming.js:168:29)
    at ProxyServer.<anonymous> (/modules/proxy/node_modules/http-proxy/lib/http-proxy/index.js:81:21)
    at middleware (/modules/proxy/node_modules/http-proxy-middleware/lib/index.js:46:13)
    at Layer.handle [as handle_request] (/modules/proxy/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/modules/proxy/node_modules/express/lib/router/index.js:317:13)
    at /modules/proxy/node_modules/express/lib/router/index.js:284:7
    at Function.process_params (/modules/proxy/node_modules/express/lib/router/index.js:335:12)
    at next (/modules/proxy/node_modules/express/lib/router/index.js:275:10)
    at expressInit (/modules/proxy/node_modules/express/lib/middleware/init.js:40:5)
    at Layer.handle [as handle_request] (/modules/proxy/node_modules/express/lib/router/layer.js:95:5)
[HPM] Error occurred while trying to proxy request /apiapi from localhost:8080 to http://www.example.org (ECONNRESET) (https://nodejs.org/api/errors.html#errors_common_system_errors)
2019/07/05 12:01:10 [notice] 1#1 process 1181 exited with code 0

freerider7777:
Is there a suggested workaround?

[...]

PR #502, opened 2020-11-24
Stream implementation  for nodejs module

wujjpp:
# Stream implementation for nodejs

For fixing #317 and #274.

features:

- Pass [request](https://github.com/request/request)'s response to ServerResponse
- Using [multer](https://github.com/expressjs/multer) for uploading file.
- Create web proxy by `http-proxy-middleware`

## 1. conf.json

```json
{
  "listeners": {
    "*:9003": {
      "pass": "applications/node"
    },
  },

  "applications": {
    "node": {
      "type": "external",
      "working_directory": "/usr/local/apps/node/",
      "executable": "app.js",
      "processes": 1
    }
  }
}
```

## 2. package.json

```json
{
  "name": "app",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "dependencies": {
    "express": "^4.17.1",
    "http-proxy-middleware": "^1.0.6",
    "multer": "^1.4.2",
    "request": "^2.88.2"
  }
}
```

## 3. nodejs server app

```javascript
#!/usr/bin/env /usr/local/nodejs/bin/node

process.env.NODE_ENV='production'

const {
          createServer,
          IncomingMessage,
          ServerResponse,
} = require('unit-http')

const http = require('http')
const { createProxyMiddleware } = require('http-proxy-middleware');

http.ServerResponse = ServerResponse
http.IncomingMessage = IncomingMessage

const express = require('express')
const request = require('request')
const fs = require('fs')
const path = require('path')

// configure multer
const multer  = require('multer')
const upload = multer({ dest: 'uploads/' })

const app = express()

app.get('/', (req, res) => res.send('Hello, Node On Unit!'))

// for testing "Writable for ServerResponse"
app.get('/proxy', (req, res, next) => {
  try {
    let path = req.query.url
    request
      .get(encodeURI(path))
      .on('error', function (err) {
        console.log(err)
         next(err)
      })
      .pipe(res)
  } catch (err) {
    next(err)
  }
})

// for testing "Readable for ServerRequest"
app.post('/upload', upload.single('uploadedFile'), function (req, res, next) {  
  const oldFileName = path.join(process.cwd(), req.file.path)
  const newFileName = path.join(process.cwd(), 'uploads', req.file.originalname)
  // rename tmp file
  fs.renameSync(oldFileName, newFileName)
  // send uploaded file back to client
  res.sendFile(newFileName)
})

// for testing web proxing served by `http-proxy-middleware`
app.get('/api', createProxyMiddleware({ target: 'http://nginx.org', changeOrigin: true }))

createServer(app).listen()
```

## 4. testing

Test Writable for ServerResponse

```shell
curl http://127.0.0.1:9003/proxy?url=http://nginx.org/
```

Test Readable for ServerRequest

```shell
curl -F "uploadedFile=@/root/unit-1.21.0.tar.gz" --output download.tar.gz http://127.0.0.1:9003/upload
```

Test proxing by `http-proxy-middleware`

```shell
curl http://127.0.0.1:9003/api/
```


src/nodejs/unit-http/http_server.js
```diff
@@ -480,10 +479,10 @@ Server.prototype.emit_request = function (req, res) {
         this.emit("request", req, res);
     }
 
-    process.nextTick(() => {
+    setTimeout(() => {
```
mar0x, line 482: Please explain why this change is important.

wujjpp: @mar0x In proxy mode, process.nextTick's callback doesn't execute, and the request not ended before timeout.

wujjpp: @mar0x In proxy mode, process.nextTick's callback execute too early, it cause the request not ended until request timeout.
You can simulate the issue by calling "curl http://127.0.0.1:9003/api/", in this case, if we emit finish event in process.nextTick's callback, our request will not ended until request time, but if we change `process.nextTick(cb)` to `setTimeout(cb ,0)`, everything goes fine.

wujjpp:
change setTimeout to setImmediate

src/nodejs/unit-http/http_server.js
```diff
@@ -11,10 +11,13 @@ const util = require('util');
...
 const Socket = require('./socket');
 const WebSocketFrame = require('./websocket_frame');
-
+const stream = require('stream');
+const Writable = stream.Writable;
+const Readable = stream.Readable;
 
 function ServerResponse(req) {
     EventEmitter.call(this);
+    Writable.call(this);
```
mar0x, line 20: Writable already inherits Stream, which, in turns, inherits EventEmitter. Is it still required to inherit EventEmitter explicitly?

src/nodejs/unit-http/http_server.js
```diff
@@ -331,21 +335,22 @@ ServerResponse.prototype.end = function end(chunk, encoding, callback) {
...
         this.finished = true;
+        this.emit('finish');
     }
 
     return this;
 };
 
 function ServerRequest(server, socket) {
     EventEmitter.call(this);
+    Readable.call(this);
```
mar0x, line 348: Readable already inherits Stream, which, in turns, inherits EventEmitter. Is it still required to inherit EventEmitter explicitly?
Currently ServerRequest overrides 'listeners' and 'resume' methods as a stubs. 'listeners' defined in EventEmitter and 'resume' - in Readable.
As soon as ServerRequest is inherited from Readable, it should have '_read' method for paused reading mode.

wujjpp: @mar0x '_read' for Readable is the best way

src/nodejs/unit-http/http_server.js
wujjpp: @mar0x inherit EventEmitter is useless, stream.Writable is already inherited

src/nodejs/unit-http/http_server.js
mar0x: '_read' func can be implemented in JS: all the data already in ServerRequest._data buffer.

wujjpp: @mar0x Thanks for your replay. 

For controlling executing sequence,  I have change all `process.nextTick(cb)` to `setImmediate(cb)` in `http_server.js`, it works as expected.

Also, I have compared the performance difference between `process.nextTck(cb)` and `setImmediate(cb)` in this case by `ab`, it seems no difference. Everything goes fine.

Follow your advice, I have add logger for recording event triggerd and inspecting `_data`'s in `ServerRequest.prototype.on`
image

The logger output after call `curl http://127.0.0.1:9003/api`
image

Unfortunately, the `_data`'s length is `0`. So add `_read` func on ServerRequest.prototype directly for feeding data may be not work

APPENDIX:

`udp-logger-server.js` for collecting log

```javascript
const dgram = require('dgram')
const server = dgram.createSocket('udp4')
server.bind(5678)

server.on('listening', function () {
  console.log('udp server linstening 5678.') // eslint-disable-line
})

server.on('message', function (msg, rinfo) {
  const message = msg.toString()
  console.log(`${new Date().getTime()}: ${message}`) // eslint-disable-line
})

server.on('error', err => {
  console.log('some error on udp server.', err.message) // eslint-disable-line
  server.close()
})
```

`logger.js` for send log in http_server.js

```javascript
const dgram = require('dgram')
const client = dgram.createSocket('udp4')

client.on('close', function () {
  console.log('udp client closed.') // eslint-disable-line
})

client.on('error', function () {
  console.log('some error on udp client.')  // eslint-disable-line
})

const send = message => {
  client.send(message, 0, message.length, 5678, '127.0.0.1')
}

module.exports.info = send

send('hello world from nodejs')
```

src/nodejs/unit-http/http_server.js
```diff
@@ -283,15 +283,15 @@ ServerResponse.prototype._writeBody = function (chunk, encoding, callback) {
     if (typeof callback === 'function') {
         /*
          * The callback must be called only when response.write() caller
-         * completes.  process.nextTick() postpones the callback execution.
+         * completes.  setImmediate() postpones the callback execution.
```
mar0x, line 284: Have you review this patch yourself before pushing?

src/nodejs/unit-http/http_server.js
```diff
@@ -283,15 +283,15 @@ ServerResponse.prototype._writeBody = function (chunk, encoding, callback) {
...
          *
-         * process.nextTick() is not technically part of the event loop.
+         * setImmediate() is not technically part of the event loop.
          * Instead, the nextTickQueue will be processed after the current
          * operation completes, regardless of the current phase of
-         * the event loop.  All callbacks passed to process.nextTick()
+         * the event loop.  All callbacks passed to setImmediate()
          * will be resolved before the event loop continues.
          */
-        process.nextTick(callback);
```
mar0x, line 290: Why do you think write callback should be executed in next event loop phase?

src/nodejs/unit-http/http_server.js
mar0x: Please do not made all the changes in this pull request. I don't see reason to blindly change `process.nextTick()` for `setImmediate()`.

mar0x: `_data.length` is `0` for GET requests because there is no request body.

wujjpp: @mar0x Thanks for your reply.

I will create new PR and explaining the reason for each changes

mar0x:
@wujjpp I've prepared the branch with minimal set of changes and '_read' implementation in https://github.com/mar0x/unit/tree/nodejs-request-stream . Please have a look and test it on installation.

wujjpp:
@mar0x coll for `_read`, let's try it

wujjpp:
@mar0x Test results:

1. `curl http://127.0.0.1:9003/proxy?url=http://nginx.org/`

> result: 
> ```
> status code: `200`
> headers: `correct`
> body: `empty`
> ```

2. `curl -F "uploadedFile=@/root/unit-1.21.0.tar.gz" --output download.tar.gz http://127.0.0.1:9003/upload`
> result
image

3.  `curl http://127.0.0.1:9003/api/`
> result
> request is not ended until timeout

mar0x:
@wujjpp Thanks for your testing! It seems, I've fixed the errors you reported earlier. Please pull review and test the branch (https://github.com/mar0x/unit/tree/nodejs-request-stream) again.

wujjpp:
@mar0x Perfect, it seems all functions are working fine by single call.

using `push(null)` instread of trigger 'finish' to notify `the next stream`

But in my opnion, the following changes are necessary:
1. In func `ServerResponse.prototype._writeBody`, we should change `process.next(callback)` to `setImmediate(callback)`
2. In func `ServerRequest.prototype._read`, we should wrap with `setImmediate`:
```javascript
ServerRequest.prototype._read = function _read(n) {
    setImmediate(() => {
        const b = this._request_read(n);

        if (b != null) {
            this.push(b);
        }

        if (!this._pushed_eofchunk && (b == null || b.length < n)) {
            this._pushed_eofchunk = true;
            this.push(null);
        }

    })
};
```

Simulate & Reason:

1. Start benchmark by following cmd
```shell
ab -c 1000 -n 500000 http://127.0.0.1:9003/
```
2. During benchmarking, if we execute one of following cmd
```shell
curl http://127.0.0.1:9003/proxy?url=http://nginx.org/
curl -F "uploadedFile=@/root/unit-1.21.0.tar.gz" --output download.tar.gz http://127.0.0.1:9003/upload
curl http://127.0.0.1:9003/upload
```
You will find the request will not ended until benchmark ended.

The root cause is `_read` funtion make event loop stay in "poll" pharse(even if the request body is empty), there is no chance to execute other pharse's callback, may be this issue occurs in sending a huge body to client(not test yet).

In real world, two scenarios will cause this issue:
1. Client post a huge body (`_read` func `starve` the other event loop pharse )
2. Server send a huge body (`_writeBody` func `starve` the other event loop pharse )

For resolving this issue, please considering made the mentioned changes to `http_server.js`

mar0x:
> In func ServerResponse.prototype._writeBody, we should change from process.next(callback) to setImmediate(callback)

Why do you think callback needs to be postponed for next loop iteration? This callback is not the function which handles _huge body_, but the notification about operation complete. If the callback produces more data, it can use setImmediate() inside.

> In func ServerRequest.prototype._read, we should wrap with setImmediate

There is no blocking calls inside current _read implementation and I don't see any reason to postpone pushing the ready chunks to handlers.

> The root cause is _read funtion make event loop stay in "poll" pharse(even if the request body is empty), there is no chance to execute other pharse's callback, may be this issue occurs in sending a huge body to client(not test yet).

Again, _request_read() does not _wait_ the data from client, it returns the data which is already delivered to application (this is the way Unit currently handles request body).

setImmediate() is a powerful function, but it's not a cure-all and it's not _free_ (as well as process.nextTick). To use it you should have a strong argumentation not just "It makes all tests pass" :)

wujjpp:
@mar0x Please review the difference by the following 2 videos

[Before](https://www.youtube.com/watch?v=1rALc5dFBmw&list=PLXQl3pjJTr9Gigt-18Kmq3sDpjMIh4Z30&index=2)
[After](https://www.youtube.com/watch?v=ta7Gx-INIYo&list=PLXQl3pjJTr9Gigt-18Kmq3sDpjMIh4Z30&index=1)

Where is my FIFO?

mar0x:
@wujjpp Unfortunately, I'm unable to see the video because it has restricted access.

Application samples and sequence of commands is more useful. I will try to reproduce it.

Regarding FIFO, it is expected only for single connection - client expects to get responses in same order as requests sent. It is hard to predict or guarantee the order of responses in different connections.

mar0x:
@wujjpp I've reproduced the issue. I confirm there is a difference in app.js behaviour running just with nodejs and under Unit. I will investigate it and think how to fix.

Thanks for your testing and interesting in Unit!

mar0x:
@wujjpp It seems I've found the root of this issue: it's a result of (bad) optimisation in our libunit.  The function which processes request messages from Unit may block.  The issue fixed in preview branch (https://github.com/mar0x/unit/tree/nodejs-request-stream).  Please pull it and test in your installation.

wujjpp:
@mar0x That's great,  all functions are working fine now. Thanks for your efforts.

PR #517, opened 2020-12-31
defines nxt_cpu_pause for ARM64

devnexen:


alejandro-colomar:
Hello David!

Would you mind adding some information to the commit message pointing to documentation of `__asm__("yield")` and maybe describing it a bit?  Why `"yield"`?  I don't know about ARM64 asm, and would like to be able to review the patch.  Thanks!

Cheers,

Alex

ac000:
This [SO](https://stackoverflow.com/questions/70810121/why-does-hintspin-loop-use-isb-on-aarch64) suggests `isb` may be a better instruction to use on aarch64...

devnexen:
Yes it is indeed, my knowledge of arm64 had grown since :-)

ac000 (approved):
LGTM. Thanks.

Just needed some tweaking to fit with current style.

ac000:
This was merged as https://github.com/nginx/unit/commit/dfdf948f899bda9120750d28c1be32f255ed941f

Sorry this took so long!

Issue #160, opened 2018-08-13
Support for .user.ini files (PHP)


avkarenow:
Hello,

.user.ini file is good for users without access to server configuration and it should allow changes only variables with PHP_INI_PERDIR/PHP_INI_ALL flags.

Now my users are using .user.ini to reconfigure their applications (php-fpm), so it will good to still have that possibility with Unit.

More about .user.ini: http://php.net/manual/en/configuration.file.per-user.php

Any chances for support for it?

markkrj:
I think this is #543?
Maybe add 'Fixes #160' in the pull request to auto close this when/if it gets merged
Btw, @mar0x do you plan to merge it?

mar0x:
Hello,
I'm afraid #543 is not ready for merge yet. I've commented the patch.

markkrj:
Regarding https://github.com/nginx/unit/pull/543/files#r714593715:
Thinking more about .user.ini, I don't know if you wanna allow live config this way. Maybe to follow what other PHP application servers (mostly apache2/mod_php and PHP-FPM) do and developers expect, but that implies checking the ttl on every request or at main loop and then checking and loading configs recursively (from app's root up to the requested php script) once the cache expires. That is the reason nginx does not have an .htaccess equivalent in the first place, but again, it is a matter of how closely you wanna follow PHP-FPM spec. It's an obvious performance trade-off.
Other option is to close this as `wont fix` and explain why, maybe with a benchmark, like [this](https://www.nginx.com/resources/wiki/start/topics/examples/likeapache-htaccess/).

markkrj:
It won't actually be updated on every request. TTL is checked on every request, and if expired, then it will be updated. But I'm mostly on side of performance. Just wanted to point out that it is a necessary trade-off **if** you wanna follow PHP-FPM spec...

VBart:
TTL cache may work good or not. First of all, the cache isn't shared between processes. As a result, if you have 50 processes, then it can be up to 50 scans for each entry even within TTL. If the processes configured to scale dynamically, then situation will be even worse as each process starts up with a clean cache.

At the same time, I guess that if we just scan all the directory tree for `.user.ini` files and load them once at the moment of creating/updating/reloading the PHP app in Unit, then that won't give any additional overhead in runtime and will work well for almost all users.  We can add it as an optional mode, that will work by default, unless a user wants the classic behaviour and turns it off in the Unit config.  What do you think?

markkrj:
[...]

PR #543, opened 2021-04-15
Support for .user.ini files (PHP: user_ini.filename/user_ini.cache_ttl)

avkarenow:
This patch adds support for .user.ini files in PHP (user_ini.filename and user_ini.cache_ttl).

Tested on FreeBSD 12.2 with PHP 5.6, 7.0-7.4, 8.0.

### DIFF
--- a/src/nxt_php_sapi.c
+++ b/src/nxt_php_sapi.c
@@ -65,6 +65,15 @@ typedef struct {
 } nxt_php_run_ctx_t;
 
 
+HashTable nxt_php_user_config_cache;
+
+
+typedef struct {
+    nxt_time_t  expires;
+    HashTable   *user_config;
+} user_config_cache_entry;
+
+
 #if NXT_PHP8
 typedef int (*nxt_php_disable_t)(const char *p, size_t size);
 #elif NXT_PHP7
@@ -137,6 +146,8 @@ static int nxt_php_unbuffered_write(const char *str, uint str_length TSRMLS_DC);
 static int nxt_php_read_post(char *buffer, uint count_bytes TSRMLS_DC);
 #endif
 
+static void nxt_php_user_config_cache_entry_dtor(user_config_cache_entry *entry);
+static void nxt_php_activate_user_config(nxt_php_run_ctx_t ctx);
 
 #ifdef NXT_PHP7
 #if PHP_VERSION_ID < 70200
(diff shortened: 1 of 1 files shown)


src/nxt_php_sapi.c
```diff
@@ -65,6 +65,15 @@ typedef struct {
 } nxt_php_run_ctx_t;
 
 
+HashTable nxt_php_user_config_cache;
+
+
+typedef struct {
+    nxt_time_t  expires;
+    HashTable   *user_config;
```
mar0x, line 73: `user_config` lifetime exactly the same as the whole `user_config_cache_entry`. Maybe it makes sense to store `HashTable` not the pointer?

avkarenow: I was inspired by implementations in uWSGI and PHP SAPIs (fpm and litespeed) which all are using the same structure. Maybe it's a good idea to keep it similar to other implementations?

888520ll19811209ll (approved):

888520ll19811209ll (approved):

src/nxt_php_sapi.c
mar0x: I think it is always good to have a reasonable explanation why pointer used instead of the structure regardless of inspiration source.

src/nxt_php_sapi.c
```diff
@@ -65,6 +65,15 @@ typedef struct {
...
+} user_config_cache_entry;
```
mar0x, line 74: According to our style, type name should be prefixed by `nxt_php_` and ends with `_t`. So, the correct type name here is `nxt_php_ user_config_cache_entry_t`.

src/nxt_php_sapi.c
```diff
@@ -65,6 +65,15 @@ typedef struct {
 } nxt_php_run_ctx_t;
 
 
+HashTable nxt_php_user_config_cache;
```
mar0x, line 68: This variable is `static`.

src/nxt_php_sapi.c
```diff
@@ -411,10 +422,63 @@ nxt_php_setup(nxt_task_t *task, nxt_process_t *process,
         nxt_php_set_options(task, value, ZEND_INI_USER);
     }
 
+    zend_hash_init(&nxt_php_user_config_cache, 0, NULL, (dtor_func_t) nxt_php_user_config_cache_entry_dtor, 1);
```
mar0x, line 425: Maximum code line length is 80. Please adjust.

src/nxt_php_sapi.c
```diff
@@ -411,10 +422,63 @@ nxt_php_setup(nxt_task_t *task, nxt_process_t *process,
...
+
     return NXT_OK;
 }
 
 
+static void
+nxt_php_activate_user_config(nxt_php_run_ctx_t ctx) {
```
mar0x, line 432: Is it required to pass `ctx` by value, not the pointer to it?
Function start brace should be on a new line.

src/nxt_php_sapi.c
```diff
@@ -411,10 +422,63 @@ nxt_php_setup(nxt_task_t *task, nxt_process_t *process,
...
+    char                     *ptr;
+    user_config_cache_entry  *new_entry, *entry;
+    nxt_time_t               request_time = (nxt_time_t) sapi_get_request_time();
+    char                     path[ctx.script_filename.length + 1];
```
mar0x, line 436: It is not allowed to use variable length arrays.
Assignment in a declaration is forbidden.
Type names should be sorted by type name length.

src/nxt_php_sapi.c
```diff
@@ -411,10 +422,63 @@ nxt_php_setup(nxt_task_t *task, nxt_process_t *process,
...
+
+    nxt_memcpy(path, ctx.script_filename.start, ctx.script_filename.length);
+
+#ifdef NXT_PHP7
+    if ((entry = zend_hash_str_find_ptr(&nxt_php_user_config_cache, path, ctx.script_filename.length)) == NULL) {
+#else /* PHP 5. */
+    if (zend_hash_find(&nxt_php_user_config_cache, path, ctx.script_filename.length + 1, (void **) &entry) == FAILURE) {
+#endif
```
mar0x, line 444: Maximum line length is 80 characters.

src/nxt_php_sapi.c
```diff
@@ -411,10 +422,63 @@ nxt_php_setup(nxt_task_t *task, nxt_process_t *process,
...
 
 
+static void
+nxt_php_activate_user_config(nxt_php_run_ctx_t ctx) {
+    char                     *ptr;
+    user_config_cache_entry  *new_entry, *entry;
+    nxt_time_t               request_time = (nxt_time_t) sapi_get_request_time();
+    char                     path[ctx.script_filename.length + 1];
+
+    nxt_memcpy(path, ctx.script_filename.start, ctx.script_filename.length);
```
mar0x, line 438: What if `ctx.script_filename.length == 0` ?

src/nxt_php_sapi.c
```diff
@@ -411,10 +422,63 @@ nxt_php_setup(nxt_task_t *task, nxt_process_t *process,
...
+        if (nxt_slow_path(new_entry->user_config == NULL)) {
+            return;
+        }
+
+        zend_hash_init(new_entry->user_config, 0, NULL, (dtor_func_t) config_zval_dtor, 1);
+#ifdef NXT_PHP7
+        entry = zend_hash_str_update_ptr(&nxt_php_user_config_cache, path, ctx.script_filename.length, new_entry);
+#else /* PHP 5. */
+        zend_hash_update(&nxt_php_user_config_cache, path, ctx.script_filename.length + 1, new_entry, sizeof(user_config_cache_entry), (void **) &entry);
+#endif
```
mar0x, line 461: Maximum line length is 80 characters.

src/nxt_php_sapi.c
```diff
@@ -411,10 +422,63 @@ nxt_php_setup(nxt_task_t *task, nxt_process_t *process,
...
+    if (request_time > entry->expires) {
+        zend_hash_clean(entry->user_config);
+        ptr = path + (ctx.root->length - 1);
+
+        while ((ptr = strchr(ptr, DEFAULT_SLASH)) != NULL) {
+            *ptr = 0;
+            php_parse_user_ini_file(path, PG(user_ini_filename), entry->user_config);
+            *ptr = '/';
+            ptr++;
+        }
```
mar0x, line 473: `DEFAULT_SLASH` macro used to find the slash, and constant `'/'` used to restore the value. It's a bit inconsistent.

src/nxt_php_sapi.c
```diff
@@ -1057,6 +1121,8 @@ nxt_php_execute(nxt_php_run_ctx_t *ctx, nxt_unit_request_t *r)
 
     SG(request_info).path_translated = NULL;
 
+    nxt_php_activate_user_config(*ctx);
```
mar0x, line 1124: `nxt_php_execute()` called for each request and new function is also called for each request. This is the main concern about this patch - possible performance degradation.

alejandro-colomar:
The design of this PR has serious performance issues.  See the discussion in <https://github.com/nginx/unit/issues/160>.  So I'm closing it.

However, the feature might be interesting, as mentioned by @VBart .  Feel free to open a new PR with a patch set that doesn't have the performance issues.

Cheers,
Alex

PR #595, opened 2021-11-14
Add "script_name" to Python configuration

### LINKED ISSUE #590: Feature Request: Support Separating `SCRIPT_NAME` and `PATH_INFO` in Python
I am requesting that Python module be updated to accept an optional `script-name` variable that is passed to WSGI applications in the `SCRIPT_NAME` environment variable, and that the `PATH_INFO` environment variable be updated to not start with the `script-name` path.

Currently, WSGI applications in Nginx Unit are passed the full request path in the `PATH_INFO` variable, even if the application can only be accessed on non-root paths. This differs from the behavior of Apache's `mod_wsgi`, which puts part of the path in `SCRIPT_NAME` and the remaining part in `PATH_INFO`.

The behavior of `mod_wsgi` means that a Django application only needs to be designed to handle its own URL routes, such as `/admin/`, but can be accessed using Apache at paths such as `/django-app/admin/` without any changes to the application.

Because Nginx Unit puts the full path in the `PATH_INFO` variable, a Django application accessible at a non-root path such as `/django` must be configured to strip `/django` from the path before determining how to route the request, and if Unit is reconfigured to place the application under `/app`, the Django application would also need to be updated.

OutOfFocus4:
Fixes #590

Allow configurations for Python applications to accept an optional `script_name` parameter. If set, the value must be a string beginning with `/`. If not set, Unit will behave as before.

The `script_name` is used in the following ways:

* In ASGI applications, the key-value pair `"root_path": "<script_name>"` is added to the `scope` dict
* In WSGI applications, the key-value pair `"SCRIPT_NAME": "<script_name>"` is added to the `environ` dict, and if the WSGI callable is called for a request whose path begins with `script_name`, the `PATH_INFO` in `environ` is set to all characters after those in `script_name`, or `/` is the request path is identical to `script_name`

### DIFF
--- a/src/nxt_application.h
+++ b/src/nxt_application.h
@@ -52,6 +52,7 @@ typedef struct {
     nxt_str_t                  module;
     char                       *callable;
     nxt_str_t                  protocol;
+    nxt_str_t                  script_name;
     uint32_t                   threads;
     uint32_t                   thread_stack_size;
     nxt_conf_value_t           *targets;
--- a/src/nxt_conf_validation.c
+++ b/src/nxt_conf_validation.c
@@ -127,6 +127,8 @@ static nxt_int_t nxt_conf_vldt_python_path_element(nxt_conf_validation_t *vldt,
     nxt_conf_value_t *value);
 static nxt_int_t nxt_conf_vldt_python_protocol(nxt_conf_validation_t *vldt,
     nxt_conf_value_t *value, void *data);
+static nxt_int_t nxt_conf_vldt_python_script_name(nxt_conf_validation_t * vldt,
+    nxt_conf_value_t *value, void *data);
 static nxt_int_t nxt_conf_vldt_threads(nxt_conf_validation_t *vldt,
     nxt_conf_value_t *value, void *data);
 static nxt_int_t nxt_conf_vldt_thread_stack_size(nxt_conf_validation_t *vldt,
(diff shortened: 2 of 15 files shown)


silane:
Great work!
I think this is a must-have feature. I hope this will be merged soon.

VBart:
Thanks for the pull request. But it seems that this approach with setting static "script_name" path will limit possible use cases. E.g. what if some app needs to be provided on multiple web-sites via different URIs?  Or even part of the app needs to be put on `/`, but other part should have some `/prefix/`?

OutOfFocus4:
I agree that this feature is limited right now, but it solves a real problem, and is purely additive; the applications you mention could simply not set `script_name` in the configuration and do the route handling in the application.

If your point is that Nginx Unit should be able to dynamically set the script name, then I agree that this pull request doesn't go far enough. However, I am not knowledgeable enough in the inner workings of Nginx Unit  to do something like that, and I haven't seen anyone with the required knowledge indicate they want to add such a feature.

I believe that a static `script_name` is a good first step, and later the Python configurations can be updated so (for example) `script_name` could be either a static string or an object mapping TLDs to script names, or mapping regular expressions to script names.

silane:
I think most of the use cases of `SCRIPT_NAME` and `root_path` is appending static prefix such as `/api/v1`. Dynamic `script_name` is surely more flexible, but event just static one helps a lot.

OutOfFocus4:
I found a bug in the code. I'm withdrawing this pull request until I fix it.

OutOfFocus4:
For anyone wondering, I *thought* I found a memory error in most of my changes, rendering the pull request a security nightmare. Not knowing how long it would take to fix, I decided to close the pull request.

Turns out, the "bug" didn’t actually exist, and I was misunderstanding an error message I was receiving while testing updates to my pull request.

Anyway, I opened a new pull request at #596

PR #596, opened 2021-11-16
Add "script_name" to Python configuration

### LINKED ISSUE #590: Feature Request: Support Separating `SCRIPT_NAME` and `PATH_INFO` in Python

OutOfFocus4:
Updated version of #595, fixes #590

Allow configurations for Python applications to accept an optional `script_name` parameter. If set, the value must be *either* a string beginning with `/`, or an array of such strings. If not set, Unit will behave as before.

The `script_name` is used in the following ways:
1. If `script_name` is a string and a request is handled whose path begins with the value of `script_name`:
	* In ASGI applications, the key-value pair `"root_path": "<script_name>"` is added to the `scope` dict
	* In WSGI applications, the key-value pair `"SCRIPT_NAME": "<script_name>"` is added to the `environ` dict, and the `PATH_INFO` in `environ` is set to all characters after those in `script_name`, or `/` if the request path is identical to `script_name`
2. If `script_name` is an array of string, each it is iterated for each request, checking if the request path begins with any of its member strings. If a `match` is found:
	* In ASGI applications, the key-value pair `"root_path": "<match>"` is added to the `scope` dict
	* In WSGI applications, the key-value pair `"SCRIPT_NAME": "<match>"` is added to the `environ` dict, and the `PATH_INFO` in `environ` is set to all characters after those in `match`, or `/` if the request path is identical to `match`

Array elements are checked in the order they appear in the configuration JSON.

The `script_name` configuration variable accepts an empty string or empty array (but not an array of empty strings). In this case, it is ignored.


OutOfFocus4:
@VBart, I think this solves the use cases you brought up in the original pull request:

> Thanks for the pull request. But it seems that this approach with setting static "script_name" path will limit possible use cases. E.g. what if some app needs to be provided on multiple web-sites via different URIs? Or even part of the app needs to be put on `/`, but other part should have some `/prefix/`?

1. In the case of serving different paths for different websites, all of those paths could be in the `script_name` array.
2. If both `/` and `/prefix` need to route to the same application, settings `script_name` to the string `/prefix` would be sufficient. Requests to `/` would not set the `SCRIPT_NAME` variable in `environ`, while requests to `/prefix` would remove `/prefix` from the request path and set the `SCRIPT_NAME` before calling the WSGI application.

What do you think? Are there any features I need to add before this pull request can be merged?

VBart:
@OutOfFocus4 Using array here doesn't look convenient, as you have to iterate over array to find the right prefix and a user of array has to think about the ordering, how to avoid collisions.  Also there are measures need to be made to separate these apps, e.g. on `example.com` there need to be prefix `/pref1/`, but on `example.org`, the prefix need to be `/pref2/`. Because they configured in the same array and applied sequentially to the same request, there's a possibility, that `/pref2/` will be served for `.com` or vice versa `/pref1/` will be served for `.org`, if it won't be properly filtered before.

We already have a way how to separate apps in one process, called `targets`: https://unit.nginx.org/configuration/#configuration-python-targets

I wonder, why did you avoid allowing configuring `script_name` in `targets` and rather went with that cumbersome array approach?

Also note the "[fastcgi_split_path_info](https://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_split_path_info)" directive in nginx, which gives wide flexibility of setting `SCRIPT_NAME` and `PATH_INFO`.

We also should think about a better name for the directive. Because `script_name` is just a mapping of CGI variable `SCRIPT_NAME` and has several problems:
1. It's not very intuitive if you aren't familiar with CGI variables and their interpretation in Python
2. It does more than just setting `SCRIPT_NAME`, while having the same name
3. More and more users use ASGI, and for them `root_path` more familiar

We have to carefully evaluate every feature and it's configuration. Because, once the decision is made and feature is implemented, then it's hard to change as users will start using it and any incompatible change will affect many setups.

I'd like to include this feature in Unit 1.27.0, but it needs to be polished as much as possible.

OutOfFocus4:
@VBart 

You are correct that `script_name` is a poor name for this directive given its intended uses. How about `path_prefix` or `application_prefix`?

It never occurred to me to use `targets`. What do you think of the following implementation?

* Both the top-level python application object and each member in the `targets` object is updated to optionally accept a regular expression called `path_prefix`. A valid configuration would only allow a top-level `path_prefix` if the `targets` object is empty. 
* When handling a request, the selected `target` is checked for a non-empty `path_prefix`. If `path_prefix` is a non-empty regular expression, it is compared to the request's `path`. If a match is found, the `SCRIPT_NAME` (if WSGI) or `root_path` (if ASGI) is set to the matching portion of portion of the path. Additionally, only portions not captured by the regular expression are used in the `PATH_INFO` for WSGI applications.

Thoughts?

OutOfFocus4:
@VBart

You are correct that `script_name` is a poor label for this directive, given its use cases. How about `path_prefix` or `application_prefix`?

It never occurred to me to use `targets`. What do you think of this?

- The top-level python application object and members of the `targets` object will accept a `path_prefix`.
    - `path_prefix` can be either a static string beginning with `"/"` or a regular expression beginning with `"^/"`. Static strings are internally converted to regular expressions.
    - `targets` with no `path_prefix` will use the top-level `path_prefix` if it exists.
    - For example, in the following application definition, the `path_prefix` of `foo` is `"/"`, and `"/bar"` for the `bar` application:
```json
{
    "applications": {
        "python-app": {
            "type": "python",
            "path": "/www/apps/python-app/",
            "path_prefix": "/bar",
            "targets": {
                "foo": {
                    "module": "foo.wsgi",
                    "callable": "foo",
                    "path_prefix": "/",
                },

                "bar": {
                    "module": "bar.wsgi",
                    "callable": "bar"
                }
            }
        }
    }
}
```

- When handling a request to a Python application, the target is checked for a non-empty `path_prefix`. If the target has a `path_prefix`, it is compared to the request's `path`. If a match is found, the matching portion is used as the `SCRIPT_NAME` (for WSGI applications) or `root_path` (for ASGI applications). Additionally, only the parts of the `path` that don't match the regular expression are used in a WSGI application's `PATH_INFO`. If the match covers the whole path, `PATH_INFO` is set to `"/"`.
- If neither the `targets` nor the top-level application use the `path_prefix` directive, requests are handled like in 1.26.0

VBart:
@OutOfFocus4
About the name:
- `application_prefix` when put inside `/applications/NAME/application_prefix` repeats "application" word again. All options inside `/applications/NAME/` already related to the application and there is no need to repeat it, otherwise we would add `applicaton_` prefix almost to any option there. Also it's a quite long option name.
- `path_prefix` has  undesirable interference with the `path` option, which is about file system path, while the discussed option is about URI parts. So, `uri_prefix` will look better than `path_prefix`, but I suggest to use just `prefix`.

Another idea: `uri_split`, or `prefix_split`? But maybe just `prefix` is good enough. What do you think?

Currently Unit doesn't inherit any options into `targets`, because it feels like better to specify explicitly each parameter, even if it requires some copy'n'paste, than deal with complexity of inheritance and having to check what targets will be affected when something is changed on the top-level. Currently Unit just prohibits specifying target parameters on the top-level when `targets` option is used. Maybe we will change this someday and allow also on the top-level with inheritance, if there will be such requests, but currently I suggest to follow the same validation rules as with other `targets` options.

Also I suggest more clean distinction of static value from a regexp one. I suggest to use the same marker as used in the `match` patterns: `~`.  When the value starts with `~`, then it will be interpreted as regexp.

Converting static strings into regexps isn't a good idea, it's better to do splitting in pure C, because:
- Regexp has complex escaping rules and you will have to follow such rules in conversion, e.g. escape all dots, as `.` in regexp has special meaning. So, the conversion may be even more complex than pure implementation in C.
- Unit can be built without PCRE dependency and hence without all the regexp functionality. It will be sad to lost any ability to set prefix in this case.

It will be a good idea to implement this option without regexp support first, and then as another pull request/patch enhance its functionality with regexps. You may even postpone this regexp enhancement unless there will be some feature requests about it. Thus by the time of implementation we will better understand the particular use case for regexp.

Also, it seems to me there's no need to require 2 capture groups for ASGI apps. The second one won't be used anyway.

OutOfFocus4:
@VBart
I updated the pull request. The changes to Unit are as follows:

1. Added the `prefix` directive to Python applications and targets. If set, it must be a string beginning with `"/"`.
1a. Like `module` and `callable`, `prefix` can only be used at the top-level if `targets` isn't used.
1b. Each target in `targets` has its own `prefix`; they are not shared or inherited in any way.
2. When a Python application handles a request, if a `prefix` exists for the target or top-level application and the request's `path` begins with the same characters as `prefix`, then:
2a. For WSGI applications, set `SCRIPT_NAME` to `prefix` and `PATH_INFO` to all characters in `path` after the characters in `prefix`, or `"/"` if `path` and `prefix` contain the same characters
2b. For ASGI applications, set `root_path` in the scope dictionary to `prefix`

VBart:
@OutOfFocus4 
What about these two configs:

1. ```json
    {
        "prefix": "/blog"
    }
    ```
2. ```json
    {
        "prefix": "/blog/"
    }
    ```
?

I think it will be more intuitive and less confusing if they will work equally. Also, looking at your code I guess the first configuration for a request like: `GET /blogger/something` will set `SCRIPT_NAME` as `/blog` and `PATH_INFO` as `ger/something`, which doesn't look right.

I suggest to add ending `/` to `prefix` at the configuration stage if there's no one.

OutOfFocus4:
@VBart
Thanks for catching that!

I updated the code to fix this. When setting up `nxt_py_targets`, any target with a `prefix` not ending in `/` will have `/` appended to the `prefix`.

When comparing the target's `prefix` to a request `path`, the process is now:
1. Check if `path` is longer than or as long as `prefix` and the characters in `prefix` are at the beginning of `path`
2. Check if `prefix` is one character longer than `path` and the characters in `path` are at the beginning of `prefix`

If (1) is true:
* For ASGI applications, set `root_path` to `prefix` minus the trailing `/`
* If WSGI, set `SCRIPT_NAME` `prefix` except for the trailing `/` and set `PATH_INFO` to all characters in `path` after those in `prefix`, but with a leading `/`.

If (2) is true:
* For ASGI applications, set `root_path` to `prefix` minus the trailing `/`
* For WSGI applications, set `SCRIPT_NAME` to `prefix` except for the trailing `/` and set `PATH_INFO` to `/`.

A `prefix` of `"/"` is pointless, since it makes `SCRIPT_NAME` and `root_path` the empty string and doesn't affect `PATH_INFO`.

OutOfFocus4:
> Note that according to CGI spec, `PATH_INFO` should either start with a `/` or be empty: https://datatracker.ietf.org/doc/html/rfc3875#section-4.1.5

I just noticed this part of your response, and I updated the pull request again.

For WSGI applications, if `path` is equal to `prefix` without a trailing `/`, `PATH_INFO` is set to the empty string rather than `/`.

OutOfFocus4:
@VBart
What do you think? Is there anything I should change?

VBart:
I'll review the code later, probably next week. Currently is busy with 1.26.1 bugfix release.

src/python/nxt_python_wsgi.c
```diff
@@ -604,8 +612,29 @@ nxt_python_get_environ(nxt_python_ctx_t *pctx)
...
+                                   (char*)prefix.start, i - 1));
+            RC(nxt_python_add_char(pctx, nxt_py_path_info_str, path, 0));
+        }
+        else {
+            RC(nxt_python_add_char(pctx, nxt_py_path_info_str, path, vl));
+        }
+    } else {
+        RC(nxt_python_add_sptr(pctx, nxt_py_path_info_str, &r->path,
+                               r->path_length));
+    }
```
VBart, line 637: It seems that if you do the opposite: always remove "/" from the "prefix" value, then code will look much simpler.

Assuming that "prefix" is always without "/" at the end:
```c
    path = nxt_unit_sptr_get(&r->path);
    path_len = r->path_length;

    if (prefix->length > 0
        && ((path_len > prefix->length && path[prefix->length] == '/')
            || path_len == prefix->length)
        && nxt_memcmp(prefix->start, path, prefix->length) == 0)
    {
        RC(nxt_python_add_char(pctx, nxt_py_script_name,
                               prefix->start, prefix->length));

        path += prefix->length;
        path_len -= prefix->length;
    }

    RC(nxt_python_add_char(pctx, nxt_py_path_info_str, path, path_len));
```

src/python/nxt_python_wsgi.c
```diff
@@ -692,10 +721,18 @@ static int
 nxt_python_add_sptr(nxt_python_ctx_t *pctx, PyObject *name,
     nxt_unit_sptr_t *sptr, uint32_t size)
 {
-    char      *src;
-    PyObject  *value;
+    char *src;
 
     src = nxt_unit_sptr_get(sptr);
+    return nxt_python_add_char(pctx, name, src, size);
```
VBart, line 727: Two style issues: 1. variable declaration needs a double space between name and type; 2. missing empty line before return value.

I actually suggest to write it as one line (there's no need in an intermediate variable):
```c
return nxt_python_add_char(pctx, name, nxt_unit_sptr_get(sptr), size);
```

src/python/nxt_python_wsgi.c
```diff
@@ -692,10 +721,18 @@ static int
...
+}
+
+
+static int
+nxt_python_add_char(nxt_python_ctx_t *pctx, PyObject *name,
+    char *src, uint32_t size)
+{
+    PyObject *value;
```
VBart, line 735: Style issue: double space is needed between the type name and variable name.

src/python/nxt_python_asgi.c
```diff
@@ -603,6 +604,27 @@ nxt_py_asgi_create_http_scope(nxt_unit_request_info_t *req)
...
+    if (prefix.length > 0) {
+        i = prefix.length;
+        target_length = r->path_length;
+        p = nxt_unit_sptr_get(&r->path);
+        target = (char*) prefix.start;
+        if (
+            (target_length >= i && nxt_strncmp(target, p, i) == 0) ||
+            // Request path matches prefix but doesn't end with "/"
+            ((target_length + 1) == i && nxt_strncmp(target, p, i - 1) == 0)
+        ) {
```
VBart, line 616: nxt_strncmp() is supposed to be used with C-strings (null-terminated). Since it's not the case, nxt_memcmp() should be used instead.

src/python/nxt_python_asgi.c
```diff
@@ -603,6 +604,27 @@ nxt_py_asgi_create_http_scope(nxt_unit_request_info_t *req)
...
+            v = PyString_FromStringAndSize((char*) target, i - 1);
+            if (nxt_slow_path(v == NULL)) {
+                nxt_unit_req_alert(req, "Python failed to create a 'root_path'"
+                                        " string");
```
VBart, line 620: Style issue: when splitting long log message the space should be left at the end of previous line, rather than at the beginning of the next line.

src/python/nxt_python_asgi.c
```diff
@@ -603,6 +604,27 @@ nxt_py_asgi_create_http_scope(nxt_unit_request_info_t *req)
...
 
+    if (prefix.length > 0) {
+        i = prefix.length;
+        target_length = r->path_length;
+        p = nxt_unit_sptr_get(&r->path);
+        target = (char*) prefix.start;
+        if (
+            (target_length >= i && nxt_strncmp(target, p, i) == 0) ||
+            // Request path matches prefix but doesn't end with "/"
+            ((target_length + 1) == i && nxt_strncmp(target, p, i - 1) == 0)
```
VBart, line 615: Multiple style issues with this "if" statement.
```c
        if ((target_length >= i && nxt_strncmp(target, p, i) == 0)
            || ((target_length + 1) == i && nxt_strncmp(target, p, i - 1) == 0))
        {
```

src/nxt_conf_validation.c
```diff
@@ -1780,6 +1795,27 @@ nxt_conf_vldt_python_protocol(nxt_conf_validation_t *vldt,
                                      "\"wsgi\" or \"asgi\".");
 }
 
+static nxt_int_t
```
VBart, line 1798: Style issue: double empty lines between function declarations are needed.

src/python/nxt_python_wsgi.c
```diff
@@ -172,7 +177,6 @@ static nxt_python_proto_t  nxt_py_wsgi_proto = {
     .done           = nxt_python_wsgi_done,
 };
 
-
```
VBart, line 175: Unrelated change, that breaks style here.

OutOfFocus4:
@VBart 

Thanks for the feedback!

I updated the code. What do you think?

OutOfFocus4:
@andrey-zelenkov
@VBart
I applied the patch. Anything else?

andrey-zelenkov:
@OutOfFocus4 Please apply few more diff's from my last patch:

```
diff --git a/test/python/targets/asgi.py b/test/python/targets/asgi.py
--- a/test/python/targets/asgi.py
+++ b/test/python/targets/asgi.py
@@ -32,8 +32,7 @@ async def application_prefix(scope, rece
             'headers': [
                 (b'content-length', b'0'),
                 (b'prefix', scope.get('root_path', 'NULL').encode()),
-                (b'request-uri', scope['path'].encode()),
-            ]
+            ],
         }
     )
 
diff --git a/test/test_python_targets.py b/test/test_python_targets.py
--- a/test/test_python_targets.py
+++ b/test/test_python_targets.py
@@ -54,13 +54,13 @@ class TestPythonTargets(TestApplicationP
                 "listeners": {"*:7080": {"pass": "routes"}},
                 "routes": [
                     {
-                        "match": {"uri": ["/app", "/app/*"]},
+                        "match": {"uri": "/app*"},
                         "action": {"pass": "applications/targets/app"},
                     },
                     {
                         "match": {"uri": "*"},
                         "action": {"pass": "applications/targets/catchall"},
-                    }
+                    },
                 ],
                 "applications": {
                     "targets": {
@@ -82,26 +82,20 @@ class TestPythonTargets(TestApplicationP
                             },
                         },
                     }
-                }
+                },
             }
         )
 
-        resp = self.get(url='/app')
-        assert resp['status'] == 200
-        assert resp['body'] == '/app '
-
-        resp = self.get(url='/app/rest/user/')
-        assert resp['status'] == 200
-        assert resp['body'] == '/app /rest/user/'
+        def check_prefix(url, body):
+            resp = self.get(url=url)
+            assert resp['status'] == 200
+            assert resp['body'] == body
 
-        resp = self.get(url='/catchall')
-        assert resp['status'] == 200
-        assert resp['body'] == 'No Script Name /catchall'
-
-        resp = self.get(url='/apis')
-        assert resp['status'] == 200
-        assert resp['body'] == 'No Script Name /apis'
-
-        resp = self.get(url='/api/users/')
-        assert resp['status'] == 200
-        assert resp['body'] == '/api /users/'
+        check_prefix('/app', '/app ')
+        check_prefix('/app/', '/app /')
+        check_prefix('/app/rest/user/', '/app /rest/user/')
+        check_prefix('/catchall', 'No Script Name /catchall')
+        check_prefix('/api', '/api ')
+        check_prefix('/api/', '/api /')
+        check_prefix('/apis', 'No Script Name /apis')
+        check_prefix('/api/users/', '/api /users/')
```

OutOfFocus4:
@andrey-zelenkov
Sorry about that. All of your changes should now be applied to the code.

andrey-zelenkov (approved):

OutOfFocus4:
Merged recent commits into my fork to ensure no conflicts.

@VBart & @andrey-zelenkov
Are there any other steps I need to take?

OutOfFocus4:
Merged recent commits into my fork to ensure no conflicts.

OutOfFocus4:
@tippexs @alejandro-colomar @artemkonev
Can someone please look at this?

alejandro-colomar:
> @tippexs @alejandro-colomar @artemkonev Can someone please look at this?

Yes.  I'll take a look soon.

tippexs:
@OutOfFocus4 looked at this again. We hab lately a discussion about the script name and did a lot of research lately. The SCRIPT_NAME can be set using the environment variable as well. That means you can simply set the value if needed. Will that be a fix for you in general? 

Timo

OutOfFocus4:
@tippexs, I'm afraid it is not a matter of "simply" setting an environment variable.

A developer can set a `SCRIPT_NAME` environment variable, but that variable, along with most other environment variables, will not be part of the `environment` dictionary passed to the application, so they must manually set the script prefix in their WSGI applications.

In addition, environment variables are set in the `application` object, so if a WSGI application is made available at multiple paths using `targets`, all of those `targets` will share a `SCRIPT_NAME` value.

Please let me know if I'm incorrect in my analysis.

tippexs:
Hi @OutOfFocus4 - Thanks for making that clear. I was checking that again and you are right. Environment variables are not included in the Python applications context.

Regarding the naming of the new variable and where to put it. So basically `SCRIPT_NAME` as well as some other configuration options names are not defined by Unit. We can just pick the names up and let the user configure it. Telling the user whats available and for what it can be used is something we can / should cover with our documentation or at least link to the right python / python frameworks docs.

We already have PHP Options to configure the PHP app context using an `options` object in the configuration syntax.
Here is the PHP example.
```
{
    "type": "php",
    "processes": 20,
    "root": "/www/blogs/scripts/",
    "user": "www-blogs",
    "group": "www-blogs",
    "options": {
        "file": "/etc/php.ini",
        "admin": {
            "memory_limit": "256M",
            "variables_order": "EGPCS"
        },

        "user": {
            "display_errors": "0"
        }
    }
}
```
So, what do you think about create an `options` object in the target like
```
{
    "applications": {
        "python-app": {
            "type": "python",
            "path": "/www/apps/python-app/",
            "targets": {
                "foo": {
                    "module": "foo.wsgi",
                    "callable": "foo",
                    "options":  {
                       "script_name": "/foo/"
                     }
                },

                "bar": {
                    "module": "bar.wsgi",
                    "callable": "bar"
                }
            }
        }
    }
}
```
This means we would give users the oportunity to define the runtime configuration params as they want. `SCRIPT_NAME` is one of them but what if they want to change some others? With this we are more flexible for future changes.

Sorry for stressing this topic about naming and stuff again but given we can bring that down to a final solution we can ship this enhancement within the next releases.

OutOfFocus4:
@tippexs I have no problem changing the name to `script_name` or `script_prefix` or `script_alias` or something else, nor do I have an issue with moving it to be inside an `options` object.

However, the `SCRIPT_NAME` is more than just a variable in the `environ` dict; it is used to strip path information from the request path to aid routing.

For example, when I develop a Django application, I can run it on a test server and go to `http://localhost:8000/tickets/` to test the application properly routes and renders the view at `/tickets/`. In production, however, that application might be under the path `/help-desk/`, and the `tickets` view would therefore be located at `/help-desk/tickets/`.

Currently, unit will send the full path `/help-desk/tickets/` to the application, which the application won't be able to properly route. With this pull request, if the `prefix` (or `script_name` or `script_prefix` or `script_alias`) is set to `/help-desk`, then `/help-desk` will be stripped from the beginning of the path, and the application will properly route the request to the view for `/tickets/`.

tippexs:
@OutOfFocus4 thanks for answering so quickly. I have just reviewed the python envrion documentation
https://peps.python.org/pep-3333/#environ-variables

and now it is clear. The `options` object does NOT make sense here. Sorry for the convusion. As `SCRIPT_NAME` is the last missing envrion according to the spec.

@alejandro-colomar and I are fine with `prefix`. We will share our findings internally tomorrow and update this right after. @alejandro-colomar will review your PR once more and let you know any concerns.

Fyi, at the moment we are working on a similiar thing with Ruby/Rails. As Ruby follows the same Spec then python it should work the exact same way. But this is out of this discussion.
Timo

OutOfFocus4:
@tippexs @hongzhidao @alejandro-colomar 
I haven’t heard anything in a while. Do I need to make any changes to this?

src/nxt_conf_validation.c
```diff
@@ -1786,6 +1801,28 @@ nxt_conf_vldt_python_protocol(nxt_conf_validation_t *vldt,
...
+    nxt_str_t  prefix;
+
+    if (nxt_conf_type(value) != NXT_CONF_STRING) {
+        return nxt_conf_vldt_error(vldt, "The \"prefix\" must be a string "
+                                   "beginning with \"/\".");
+    }
+
+    nxt_conf_get_string(value, &prefix);
+
+    if (prefix.length < 1 || prefix.start[0] != '/') {
```
alejandro-colomar, line 1817: You could use the following:

```c
$ grepc nxt_strchr_start
./src/nxt_string.h:155:
#define nxt_strchr_start(s, c)                                                \
    (((s)->length != 0) && ((s)->start[0] == c))
```

So:

```c
if (!nxt_strchr_start(&prefix, '/')) {
```

src/nxt_conf_validation.c
```diff
@@ -1786,6 +1801,28 @@ nxt_conf_vldt_python_protocol(nxt_conf_validation_t *vldt,
...
 
+static nxt_int_t
+nxt_conf_vldt_python_prefix(nxt_conf_validation_t *vldt,
+    nxt_conf_value_t *value, void *data)
+{
+    nxt_str_t  prefix;
+
+    if (nxt_conf_type(value) != NXT_CONF_STRING) {
+        return nxt_conf_vldt_error(vldt, "The \"prefix\" must be a string "
+                                   "beginning with \"/\".");
```
alejandro-colomar, line 1812: I think I prefer specifying the character as `'/'`, instead of `\"/\"`, for consistency with C's syntax for referring to strings and characters.

src/python/nxt_python.c
```diff
@@ -373,6 +375,33 @@ nxt_python_set_target(nxt_task_t *task, nxt_python_target_t *target,
         goto fail;
     }
 
+    value = nxt_conf_get_object_member(conf, &prefix_str, NULL);
```
alejandro-colomar, line 378: Maybe this whole code block could go into a separate helper function?  This function is already growing more than I'd like, I think.

src/python/nxt_python.c
```diff
@@ -677,7 +707,14 @@ nxt_python_atexit(void)
 
     if (nxt_py_targets != NULL) {
         for (i = 0; i < nxt_py_targets->count; i++) {
-            Py_XDECREF(nxt_py_targets->target[i].application);
+            target = &nxt_py_targets->target[i];
+
+            Py_XDECREF(target->application);
+            Py_XDECREF(target->py_prefix);
+
+            if (target->prefix.length > 0) {
```
alejandro-colomar, line 715: Do we really need this check?  What does the string contain when the length is 0?  I'd expect a NULL pointer, which can be passed to *free(3)* safely.  Or is it maybe an uninitialized pointer?  Please check.

src/python/nxt_python_asgi.c
```diff
@@ -610,6 +613,17 @@ nxt_py_asgi_create_http_scope(nxt_unit_request_info_t *req)
         return NULL;
     }
 
+    prefix = app_target->prefix;
+    path_length = r->path_length;
+    p = nxt_unit_sptr_get(&r->path);
+    if (prefix.length > 0
+        && ((path_length > prefix.length && p[prefix.length] == '/')
+            || path_length == prefix.length)
```
alejandro-colomar, line 621: Maybe you could combine this test and the `nxt_memcmp()` one with the following:

```c
$ grepc nxt_str_start
./src/nxt_string.h:147:
#define nxt_str_start(s, p, _length)                                          \
    (((s)->length >= _length) && (nxt_memcmp((s)->start, p, _length) == 0))
```
So something like:
```c
nxt_str_start(p, prefix, prefix.start, prefix.length)
```
might help (but maybe not, and maybe it just complicates it more).  Please check.

test/python/prefix/wsgi.py
```diff
@@ -0,0 +1,10 @@
+def application(environ, start_response):
+    start_response(
+        '200',
+        [
+            ('Content-Length', '0'),
+            ('Script-Name', environ.get('SCRIPT_NAME', 'NULL')),
+            ('Path-Info', environ['PATH_INFO']),
```
alejandro-colomar, line 7: I'd like to understand better what is the relation between prefix and SCRIPT_NAME and PATH_INFO.  Could you please explain it to me?

alejandro-colomar:
> @tippexs @hongzhidao @alejandro-colomar I haven’t heard anything in a while. Do I need to make any changes to this?

Hi!

Sorry, I was doing other things and let this aside for a too long time.  I left some review before I go on vacation.  I'll be back next month.

Cheers,
Alex

test/python/prefix/wsgi.py
OutOfFocus4: From the [WSGI specification](https://peps.python.org/pep-3333/#environ-variables):
> `SCRIPT_NAME`
> The initial portion of the request URL’s “path” that corresponds to the application object, so that the application knows its virtual “location”. This may be an empty string, if the application corresponds to the “root” of the server.

> `PATH_INFO`
> The remainder of the request URL’s “path”, designating the virtual “location” of the request’s target within the application. This may be an empty string, if the request URL targets the application root and does not have a trailing slash.

As an example, say I have a server with a help desk application. The help desk application is the only application, so all HTTP requests to it are routed through that application. A request to `/tickets/20/` will result in the application being called with an empty `SCRIPT_NAME` and a `PATH_INFO` of `/tickets/20/`.

If I someday wish to add a separate customer survey application to the same server, I may update the application server configuration to route requests to `/help` and its descendants to the help desk, and requests to `/survey` and its descendants to the survey application.

In this case, a request to `/help/tickets/20/` will result in the help desk application being called with a `SCRIPT_NAME` of `/help` and a `PATH_INFO` of `/tickets/20/`. The routing logic doesn't need to be changed, and the `SCRIPT_NAME` can be used during server-side HTML rendering so links to `/tickets/all/` are now links to `/help/tickets/all/`.

src/nxt_conf_validation.c
OutOfFocus4: The value of `"prefix"` is set in JSON, where only double-quotes are allowed. I was more worried about consistency with the language of the configuration (JSON) than the language of the program validating the configuration (C).

I can change it if you want.

alejandro-colomar: Hmm, I don't have a strong opinion here.  You opinion makes sense too.  I didn't think of JSON, because as it doesn't represent a full JSON string, but rather one character of a JSON string...  But whatever makes more sense to you is good.

So, you decide.

src/nxt_application.h
```diff
@@ -49,8 +49,6 @@ typedef struct {
 typedef struct {
     char                       *home;
     nxt_conf_value_t           *path;
-    nxt_str_t                  module;
```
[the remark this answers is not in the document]
alejandro-colomar: Makes sense.  Please remove it in a separate commit (it's fine if the commit is in the same PR, but then put it _before_ the actual commit that implements the prefix thing).

alejandro-colomar:
What did change in the last force fush?  Due to the lot changes in the master branch, the "Compare" button doesn't show nicely the difference in this PR.

src/python/nxt_python_asgi.c
OutOfFocus4: `nxt_str_start` won't work, because `p` is a `char *`, not an `nxt_str_t`.

OutOfFocus4:
> What did change in the last force fush? Due to the lot changes in the master branch, the "Compare" button doesn't show nicely the difference in this PR.

Sorry about that. I replaced the check `(prefix.length < 1 || prefix.start[0] != '/')` with `(!nxt_strchr_start(&prefix, '/'))` like you suggested and removed more unused code I found.

My most recent push splits the unused code removal from the rest of the PR.

src/nxt_application.h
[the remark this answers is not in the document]
OutOfFocus4: I have split the PR into two commits.

src/nxt_conf_validation.c
OutOfFocus4: I think I will keep it as-is. We shouldn't assume that the developers using this application are all familiar with C and the difference between double- and single-quotes.

src/python/nxt_python.c
OutOfFocus4: After consulting Stack Overflow, it seems the result of calling `malloc` with a size of 0 will return either `NULL` or a pointer that should never be dereferenced, depending on the implementation. Either way, the result is always safe to pass to `free`, so I removed the check.

src/python/nxt_python.c
OutOfFocus4: Is it really a "helper" function if it is only called once from a single place?

alejandro-colomar: It's not a bad thing, IMHO.  If we know it's only called at one place and inconditionally, we can make it `[[gnu::always_inline]]` to help the compiler (in case we can prove it doesn't inline it on its own).

We also get the opportunity to name the whole section of code, which might increase readability.  Unless you consider that moving it into a separate function would hurt readability because of loosing context.

alejandro-colomar: It would also simplify error handling.

alejandro-colomar: I'm not entirely happy about this:

```c
                nxt_free(prefix);
                goto fail;
```

alejandro-colomar:
The changes look good to me.  `Reviewed-by: Alejandro Colomar <alx@nginx.com>`

However, I have the small concerns about reorganizing the code that I mentioned in comments (which don't affect behavior at all).

@andrey-zelenkov and @artemkonev please have a look in case you want to comment anything about the tests or the interface.

src/python/nxt_python.c
OutOfFocus4: I moved the code into a separate function. All the tests pass, but please let me know if you notice any problems with my changes.

src/python/nxt_python.c
```diff
@@ -390,6 +396,45 @@ nxt_python_set_target(nxt_task_t *task, nxt_python_target_t *target,
...
+            nxt_alert(task, "Python failed to allocate target prefix "
+                            "string");
+            return NXT_ERROR;
+        }
+        nxt_memcpy(prefix, str.start, str.length);
+        target->prefix.start = prefix;
+    }
+
+    return NXT_OK;
+}
```
alejandro-colomar, line 435: It would be simpler to have:

```c
if (str.length == 0) {
    return NXT_OK;
}
```

src/python/nxt_python.c
alejandro-colomar: Thanks for the changes, and thanks for keeping them as a separate patch (it made the differential review much easier).  They look good (still can improve a little bit, but great job!).

alejandro-colomar:
Could you please add some information in the commit message?

You could add a condensed version of the discussion in this github page.  I'm interested in why the patch is necessary, and how it is expected to be used.

Also, if you could sign the patches, it would be great. <https://www.kernel.org/doc/html/latest/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin>

Thanks!
Alex

src/python/nxt_python.c
OutOfFocus4: Thanks for the tip. I updated the code.

OutOfFocus4:
> Could you please add some information in the commit message?

Would this be a new commit, or should I edit the commit message that currently reads `Add "prefix" to the Python configuration`?

> Also, if you could sign the patches, it would be great.

Future commits will be signed. I tried signing the current commits, but it failed due a misconfiguration on my end.

alejandro-colomar:
> > Could you please add some information in the commit message?
> 
> Would this be a new commit, or should I edit the commit message that currently reads `Add "prefix" to the Python configuration`?

Edit the commit message.

> 
> > Also, if you could sign the patches, it would be great.
> 
> Future commits will be signed. I tried signing the current commits, but it failed due a misconfiguration on my end.

Okay.  I can do that.  What name and email would you like to use?

Thanks!

alejandro-colomar:
Thanks, will merge later today probably!

Great job!

src/nxt_main_process.c
```diff
@@ -190,18 +190,6 @@ static nxt_conf_map_t  nxt_python_app_conf[] = {
...
-        NXT_CONF_MAP_STR,
-        offsetof(nxt_common_app_conf_t, u.python.module),
-    },
-
-    {
-        nxt_string("callable"),
-        NXT_CONF_MAP_CSTRZ,
-        offsetof(nxt_common_app_conf_t, u.python.callable),
-    },
-
```
alejandro-colomar, line 204: I didn't realize:  shouldn't this be removed in the first commit?  The one that removes unused code.

src/python/nxt_python_wsgi.c
```diff
@@ -692,10 +719,16 @@ static int
...
-    src = nxt_unit_sptr_get(sptr);
+static int
+nxt_python_add_char(nxt_python_ctx_t *pctx, PyObject *name,
+    char *src, uint32_t size)
+{
+    int       res;
+    PyObject  *value;
 
     value = PyString_FromStringAndSize(src, size);
     if (nxt_slow_path(value == NULL)) {
```
alejandro-colomar, line 734: This seems like an unrelated refactoring.  Please separate it into a commit that goes before the main change, to be able to better discern the real changes.

src/nxt_application.h
```diff
@@ -49,7 +49,6 @@ typedef struct {
 typedef struct {
     char                       *home;
     nxt_conf_value_t           *path;
-    char                       *callable;
```
alejandro-colomar, line 53: This seems to also need to go into the commit that removes unused code.  Please check if there's anything else that should go there.

alejandro-colomar:
Also, please squash commits 2 and 3, since commit 3 is just a refactoring of the newly introduced code.

src/nxt_application.h
OutOfFocus4: I reorganized the commits.

alejandro-colomar: Thanks!

alejandro-colomar:
Hi!

I already applied the first two commits.  I'll have another look at the remaining one.

alejandro-colomar:
Would you mind rebasing against the master branch?

src/nxt_application.h
alejandro-colomar, line 52: This should go to a separate commit that removes dead code.

alejandro-colomar:
@OutOfFocus4 Would you please add a changelog line (in <docs/changes.xml>)?

@artemkonev  The code changes in this patch look good to me.  Would you mind having a look at it to check the changes to the configuration make sense to you?

src/nxt_application.h
OutOfFocus4: I created a new commit for removing code

alejandro-colomar:
At commit 5f52174: `Reviewed-by: Alejandro Colomar <alx@nginx.com>`

alejandro-colomar (approved):

artemkonev (approved):
LGTM, I'll proceed to amend the docs some time later.
BTW, I suggest adding @andrey-zelenkov to the loop to take a look at the tests.

alejandro-colomar:
> LGTM, I'll proceed to amend the docs some time later.

Thanks!

> BTW, I suggest adding @andrey-zelenkov to the loop to take a look at the tests.

AFAIK he already reviewed them (this PR is approved by him).  However, it might be good if he can just confirm.  @andrey-zelenkov

alejandro-colomar:
Merged.

Issue #597, opened 2021-11-18
OpenSSL 3.0 Compatibility


remicollet:
AS this will be the default version in Fedora 36 and RHEL 9 (and Centos and other clones)

For now:

```
$ make
cc -c -pipe -fPIC -fvisibility=hidden -O -W -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wmissing-prototypes -Werror -g   -I src -I build -I/opt/openssl/include  \
                      \
                     \
-o build/src/nxt_openssl.o \
-MMD -MF build/src/nxt_openssl.dep -MT build/src/nxt_openssl.o \
src/nxt_openssl.c
src/nxt_openssl.c: In function 'nxt_tls_ticket_key_callback':
src/nxt_openssl.c:820:5: error: 'HMAC_Init_ex' is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
  820 |     if (HMAC_Init_ex(hctx, ticket[i].hmac_key, ticket[i].size, digest, NULL)
      |     ^~
In file included from /opt/openssl/include/openssl/ssl.h:37,
                 from src/nxt_openssl.c:9:
/opt/openssl/include/openssl/hmac.h:43:27: note: declared here
   43 | OSSL_DEPRECATEDIN_3_0 int HMAC_Init_ex(HMAC_CTX *ctx, const void *key, int len,
      |                           ^~~~~~~~~~~~
src/nxt_openssl.c: In function 'nxt_openssl_copy_error':
src/nxt_openssl.c:1784:9: error: 'ERR_get_error_line_data' is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
 1784 |         err = ERR_get_error_line_data(NULL, NULL, &data, &flags);
      |         ^~~
In file included from src/nxt_openssl.c:11:
/opt/openssl/include/openssl/err.h:413:15: note: declared here
  413 | unsigned long ERR_get_error_line_data(const char **file, int *line,
      |               ^~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
make: *** [build/Makefile:1375: build/src/nxt_openssl.o] Error 1

```

ac000:
This has been handled in commit 5665838b680c

PR #598, opened 2021-11-18
Fix build with OpenSSL 3

remicollet:
See #597 

1st commit drop usage of deprecated ERR_get_error_line_data

2nd commit is a workaround waiting for callback refactoring (move from `SSL_CTX_set_tlsext_ticket_key_cb` to `SSL_CTX_set_tlsext_ticket_key_evp_cb`)

### DIFF
--- a/src/nxt_openssl.c
+++ b/src/nxt_openssl.c
@@ -1777,7 +1777,11 @@ nxt_openssl_copy_error(u_char *p, u_char *end)
     clear = 0;
 
     for ( ;; ) {
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+        err = ERR_get_error_all(NULL, NULL, NULL, &data, &flags);
+#else
         err = ERR_get_error_line_data(NULL, NULL, &data, &flags);
+#endif
         if (err == 0) {
             break;
         }


src/nxt_openssl.c
```diff
@@ -1781,7 +1785,11 @@ nxt_openssl_copy_error(u_char *p, u_char *end)
     clear = 0;
 
     for ( ;; ) {
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
```
remicollet, line 1788: This can even be used with older version (tested with 1.1.1)

alejandro-colomar: Yes; please use it for all cases and remove the conditional.

That also allows removing the pragma, I guess.

src/nxt_openssl.c
```diff
@@ -14,6 +14,10 @@
 #include <openssl/bio.h>
 #include <openssl/evp.h>
 
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+/* TODO removed when SSL_CTX_set_tlsext_ticket_key_cb updated to SSL_CTX_set_tlsext_ticket_key_evp_cb */
```
alejandro-colomar, line 18: What needs to be done exactly?

Is the replacement difficult?

alejandro-colomar: I'd prefer adding -Wno-error=deprecated-declarations to the build system, so that we get a reminder every time we compile that the warning is to be fixed.

alejandro-colomar:
For patch 1: `Reviewed-by: Alejandro Colomar <alx@nginx.com>`

src/nxt_openssl.c
andypost: The [migration guide](https://www.openssl.org/docs/man3.0/man7/migration_guide.html) said there's few variants


> ERR_load_*(), ERR_func_error_string(), ERR_get_error_line(), ERR_get_error_line_data(), ERR_get_state()
>> OpenSSL now loads error strings automatically so these functions are not needed.

> Applications should use [ERR_get_error_all(3)](https://www.openssl.org/docs/man3.0/man3/ERR_get_error_all.html), or pick information with ERR_peek functions and finish off with getting the error code by using [ERR_get_error(3)](https://www.openssl.org/docs/man3.0/man3/ERR_get_error.html).

src/nxt_openssl.c
ac000: We don't currently get the OpenSSL 3.0 deprecation warnings due to https://github.com/nginx/unit/commit/5665838b680cdc06d6eb83107e4d46db6cfae0c4

alejandro-colomar:
@remicollet 
Please remove the preprocessor conditionals (since it works with old OpenSSL versions too) and the pragma (since we don't get the warning, as Andrew reported).
Also, please sign the patch.

Thanks!

src/nxt_openssl.c
remicollet: Sorry, my mistake, this is really openssl 3 only, see

https://www.openssl.org/docs/man3.0/man3/ERR_get_error_line_data.html   (ERR_get_error_all is there)
https://www.openssl.org/docs/man1.1.1/man3/ERR_get_error_line_data.html (ERR_get_error_all not available)

remicollet:
rebased, pragma removed, and signed

alejandro-colomar:
Merged.  Thanks!

Issue #643, opened 2022-02-13
unix socket not closed on SIGTERM or SIGQUIT


echolimazulu:
___
### - All users of the unix socket as a listener
___

Hello Nginx Team and Contributors,

When a **SIGTERM** or a **SIGQUIT** signals are received, the unitd daemon does not close the existing `unix:` socket listener. On a subsequent restart of unitd with a configuration to use the same `unix:` socket, it becomes impossible due to a previously unclosed socket.

**Also note:** 
`"32: Broken pipe"` when closing processes.

**Log:**
```
# unitd
2022/02/13 12:20:13 [info] 31931#31931 unit 1.26.1 started

# ls -l /var/run/unit.sock
srw-rw-rw- 1 root root 0 Feb 13 12:20 /var/run/unit.sock

# kill -15 $(cat /var/run/unit.pid) && cat /var/log/unit.log
...
2022/02/13 12:21:43 [notice] 31958#31958 process 31960 exited with code 0
2022/02/13 12:21:43 [notice] 31958#31958 process 31961 exited with code 0
2022/02/13 12:21:43 [notice] 31962#31962 app process (isolated 31963) exited with code 0
2022/02/13 12:21:43 [alert] 31962#31962 sendmsg(13, -1, -1, 2) failed (32: Broken pipe)
2022/02/13 12:21:43 [notice] 31958#31958 process 31962 exited with code 0

# ls /var/run/unit.sock
srw-rw-rw- 1 root root 0 Feb 13 12:21 /var/run/unit.sock

# unitd
2022/02/13 12:22:35 [info] 31975#31975 unit 1.26.1 started

# cat /var/log/unit.log
...
2022/02/13 12:22:35 [alert] 31976#31976 bind(\"unix:/var/run/unit.sock\") failed (98: Address already in use)
2022/02/13 12:22:35 [alert] 31979#31979 failed to apply new conf
2022/02/13 12:22:35 [alert] 31978#31978 failed to apply previous configuration
```

ac000:
Since ccaad38bc529897218a1e41c7e8704566695d1a9 Unix Domain Sockets are properly unlink(2)'d at shutdown.

PR #649, opened 2022-02-17
solving the unix socket reuse issue

echolimazulu:
This patch solves the problem of reusing unix sockets.

**Changes:**
- added removal of socket file before "bind" operation. necessary in case of unexpected termination of the program and subsequent restart without errors;
- added removal of a socket after it is closed in case of errors when creating a socket;
- minor optimizations and fixes.

Issue https://github.com/nginx/unit/issues/643:
`bind(\"unix:/var/run/unit.sock\") failed (98: Address already in use)`

**Notice:**
This PR does not address the issue of deleting sockets when exiting a program.

### DIFF
--- a/src/nxt_main_process.c
+++ b/src/nxt_main_process.c
@@ -1,5 +1,6 @@
 
 /*
+ * Copyright (C) Evgenii Sokolov
  * Copyright (C) Igor Sysoev
  * Copyright (C) NGINX, Inc.
  */
@@ -1096,18 +1097,20 @@ nxt_main_listening_socket(nxt_sockaddr_t *sa, nxt_listening_socket_t *ls)
 {
     nxt_err_t         err;
     nxt_socket_t      s;
+    sa_family_t       family;
 
     const socklen_t   length = sizeof(int);
     static const int  enable = 1;
 
-    s = socket(sa->u.sockaddr.sa_family, sa->type, 0);
+    family = sa->u.sockaddr.sa_family;
+    s = socket(family, sa->type, 0);
 
     if (nxt_slow_path(s == -1)) {
         err = nxt_errno;
 
 #if (NXT_INET6)
 
-        if (err == EAFNOSUPPORT && sa->u.sockaddr.sa_family == AF_INET6) {
+        if (err == EAFNOSUPPORT && family == AF_INET6) {
             ls->error = NXT_SOCKET_ERROR_NOINET6;
         }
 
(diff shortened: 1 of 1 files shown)


echolimazulu:
@mar0x, Good night! 
I will be glad to your review, questions and comments. Thanks.

mar0x:
Hello,
I'm afraid this approach is unsafe. Imagine one's "mistake" to configure listener's unix socket `/usr/sbin/unitd` or `/etc/passwd`... Another bad scenario is removing socket created by other running Unit instance.

Before removing file, Unit need to make sure it's a unix domain socket and it is abandoned now. To achieve this, it is possible to use file locks, try to connect to socket or something else.

VBart:
There's already a function that does all the necessary checks in order to avoid various unix socket collisions: https://hg.nginx.org/unit/rev/0a8840921fd0, but at the moment it only used for the control socket.

mar0x:
> There's already a function that does all the necessary checks in order to avoid various unix socket collisions: https://hg.nginx.org/unit/rev/0a8840921fd0, but at the moment it's only used for the control socket.

Correct. But the control socket creation procedure does not check the file is a socket (which is Ok for control socket). Perhaps for other listener sockets we want to avoid regular files removal to create unix socket, removing old sockets is Ok.

echolimazulu:
> Hello, I'm afraid this approach is unsafe. Imagine one's "mistake" to configure listener's unix socket `/usr/sbin/unitd` or `/etc/passwd`... Another bad scenario is removing socket created by other running Unit instance.
> 
> Before removing file, Unit need to make sure it's a unix domain socket and it is abandoned now. To achieve this, it is possible to use file locks, try to connect to socket or something else. 

**Good morning** @mar0x,
Accepted. 

Totally agree with you. I have taken note of all your comments. I'm going to make the necessary corrections. Thank you for your comments, they help me provide a better solution to the problem.

> There's already a function that does all the necessary checks in order to avoid various unix socket collisions: https://hg.nginx.org/unit/rev/0a8840921fd0, but at the moment it's only used for the control socket.

**Good morning** @VBart, 
I'm glad you joined us. 

The problem of not being able to reopen a unix socket only affects the socket as a listener. What @mar0x mentioned does not apply to the control socket, but user mistake is minimized if it is used. From my point of view, listening sockets require more fine-grained control and maintenance.

VBart:
Anyway, I suggest to avoid code duplication and add the necessary check (i.e. checking file type after ECONNREFUSED) to the function mentioned above, then reuse it for listening sockets.

echolimazulu:
> Anyway, I suggest to avoid code duplication and add the necessary check (i.e. checking file type after ECONNREFUSED) to the function mentioned above, then reuse it for listening sockets.

@VBart Healthy thought. I already thought about this, but initially I thought that you were forking the code in these two different files on purpose. The labels in `nxt_listen_socket_create` will have to be reworked a little, as they currently look like potentially causing an error (`nxt_file_delete` before `nxt_socket_close` - without `return` in `listen_fail` they are executed sequentially), and passing the access mode to configure a socket with permissions other than `600`. I am in the process of solving this problem. If you have any other ideas please let me know. Everything you have suggested is really valuable ideas.

VBart:
@echolimazulu That split wasn't made by intention. The whole Unit project was forked out of some other project and as a result there are still many leftovers of old code originally intended for different purpose.

echolimazulu:
@VBart, @mar0x Good morning!

I closed this PR, as in the future I will open a new one, taking into account all the changes.

I have carefully studied and implemented all your suggestions, such as:
- checking for the necessary permissions when creating a socket file;
- check for file type matching socket when deleting;
- creating and checking for the presence of a lock file;
- use of the functional part of the program from listen_socket with the implementation of the error system when creating a socket from main_process (rpc);
- deleting socket/lock files when closing the SIGQUIT, SIGKILL program;
- deleting socket/lock files when changing the program configuration file;
- setting the correct values ​​of the necessary access rights to the specified files: controller, listeners, locks.

At the moment, I am testing the modified functionality. 

**The changes will affect the following list of files:**
- nxt_array.h / nxt_array.c
- nxt_controller.h / nxt_controller.c
- nxt_errno.h
- nxt_file.h / nxt_file.c
- nxt_listen_socket.h / nxt_listen_socket.c
- nxt_main_process.h / nxt_main_process.c
- nxt_port.h
- nxt_process.c
- nxt_router.h / nxt_router.c
- nxt_runtime.h / nxt_runtime.c
- nxt_sockaddr.h / nxt_sockaddr.c
- nxt_socket.h / nxt_socket.c
- other files in which the new functionality is implemented

Please also check out my updated and submitted PR (https://github.com/nginx/unit/pull/663) regarding the "Array" functionality, as this functionality will also be used by me in upcoming changes.

Thanks!

PR #655, opened 2022-03-02
Supporting UNIX sockets in address matching.

alejandro-colomar:


alejandro-colomar:
Simplified it to support any unix socket, by specifying the string `"unix"`.  A path shall not be specified.

src/nxt_http_route_addr.c
```diff
@@ -47,6 +47,16 @@ nxt_http_route_addr_pattern_parse(nxt_mp_t *mp,
 
     nxt_str_null(&port);
```
VBart, line 48: Nulling port variable has no sense in case of "unix" socket. This should be after the "unix" socket condition, not before.

alejandro-colomar: Oops.  Right.

VBart: It makes sense to move "unix" socket condition also above the `addr.length < 2` check. Because that check is directly related to the following condition with `addr.start[0] == '*' && addr.start[1] == ':'`. Keeping them together makes code self-explainable about why length should be not less than 2.

src/nxt_http_route.c
```diff
@@ -1947,6 +1947,14 @@ nxt_http_route_addr_pattern_match(nxt_http_route_addr_pattern_t *p,
         break;
 #endif
 
+#if (NXT_HAVE_UNIX_DOMAIN)
+    case AF_UNIX:
+
+        match = (base->addr_family == AF_UNIX
+                 || base ->addr_family == AF_UNSPEC);
```
VBart, line 1954: Hm... doesn't it look strange if `*:80` will match any of Unix domain sockets as well?  This is also a significant behaviour change as currently it doesn't match Unix sockets.  `AF_UNSPEC` was used to match both IPv4 and IPv6, but not Unix sockets.

Note also a surplus space after the `base` in `base ->addr_family`.

VBart:
You also need to add a feature to `changes.xml`.

src/nxt_http_route.c
alejandro-colomar: Ahhh, I was very confused here.  Will fix it in a moment.

alejandro-colomar:
> You also need to add a feature to `changes.xml`.

Done.  I still need to add a pytest.

docs/changes.xml
```diff
@@ -37,6 +37,12 @@ compatibility with GCC 12.
 </para>
 </change>
 
+<change type="feature">
+<para>
+supporting UNIX sockets in address matching
```
VBart, line 42: The dot at the end of sentence is missing.

Also I usually try to estimate subjectively the amount of users affected or importance of the feature/fix for them and sort relatively to that estimation. In this case, this feature seems more important than support for a not-yet-released compiler version (so it should go above).

Also 2, it makes sense to consult with Artem about the best wording. Commit log for devs, change log for users. The current wording is good enough, but I have some doubts that many users will understand what "address matching" is meant here. Maybe it's worth to describe specifically, that this is about address matching in routing and "client_ip" options.

alejandro-colomar: > The dot at the end of sentence is missing.

D'oh.

> 
> Also I usually try to estimate subjectively the amount of users affected or importance of the feature/fix for them and sort relatively to that estimation. In this case, this feature seems more important than support for a not-yet-released compiler version (so it should go above).

Makes sense.

> 
> Also 2, it makes sense to consult with Artem about the best wording. Commit log for devs, change log for users. The current wording is good enough, but I have some doubts that many users will understand what "address matching" is meant here. Maybe it's worth to describe specifically, that this is about address matching in routing and "client_ip" options.

Yes, and yes, I'll add those details regarding client_ip.

alejandro-colomar:
I'm getting EACCESS from unlink(2) :(

alejandro-colomar:
I broke the patch I added yesterday for unlinking sockets to a new separate PR.

Issue #653, opened 2022-02-23
Ruby 3.x module on Mac OS failing to build


travisbell:
Trying to compile the Ruby module on Mac OS and Ruby 3.1.1 (3.0.3 is the ~same) results in the following:

```
configuring Ruby module ...
ruby 3.1.1p18 (2022-02-18 revision 53f5fc4236) [x86_64-darwin21]
----------------------------------------
checking for Ruby library
In file included from build/autotest.c:2:
In file included from /Users/travisbell/.rbenv/versions/3.1.1/include/ruby-3.1.0/ruby.h:38:
In file included from /Users/travisbell/.rbenv/versions/3.1.1/include/ruby-3.1.0/ruby/ruby.h:25:
In file included from /Users/travisbell/.rbenv/versions/3.1.1/include/ruby-3.1.0/ruby/defines.h:73:
In file included from /Users/travisbell/.rbenv/versions/3.1.1/include/ruby-3.1.0/ruby/backward/2/attributes.h:42:
In file included from /Users/travisbell/.rbenv/versions/3.1.1/include/ruby-3.1.0/ruby/internal/attr/pure.h:25:
/Users/travisbell/.rbenv/versions/3.1.1/include/ruby-3.1.0/ruby/assert.h:132:1: error: '__declspec' attributes are not enabled; use '-fdeclspec' or '-fms-extensions' to enable support for __declspec attributes
RBIMPL_ATTR_NORETURN()
^
/Users/travisbell/.rbenv/versions/3.1.1/include/ruby-3.1.0/ruby/internal/attr/noreturn.h:29:33: note: expanded from macro 'RBIMPL_ATTR_NORETURN'
# define RBIMPL_ATTR_NORETURN() __declspec(noreturn)
                                ^
In file included from build/autotest.c:2:
In file included from /Users/travisbell/.rbenv/versions/3.1.1/include/ruby-3.1.0/ruby.h:38:
In file included from /Users/travisbell/.rbenv/versions/3.1.1/include/ruby-3.1.0/ruby/ruby.h:26:
In file included from /Users/travisbell/.rbenv/versions/3.1.1/include/ruby-3.1.0/ruby/internal/anyargs.h:77:
/Users/travisbell/.rbenv/versions/3.1.1/include/ruby-3.1.0/ruby/internal/intern/vm.h:383:1: error: '__declspec' attributes are not enabled; use '-fdeclspec' or '-fms-extensions' to enable support for __declspec attributes
[...]

PR #656, opened 2022-03-10
Workaround clang bug in __has_declspec_attribute()

alejandro-colomar:
Add -fdeclspec to NXT_RUBY_CFLAGS for Clang, if it's available.

Clang incorrectly reports 1 for __has_declspec_attribute(x) in
some cases, such as MacOS or Cygwin.  That causes ruby code to
break.  ruby added -fdeclspec to their CFLAGS in 2019 to
workaround this bug, since it enables __declspec() and therefore,
the compiler behavior matches what it reports.

Since we don't know what are all the architectures that trigger
the clang bug, let's add the flag for all of them (especially
since it should be harmless).

Add this workaround only at the time of configuring the ruby
module.  This way we don't clutter the global NXT_CFLAGS with an
unnecessary flag.

Link: unit bug <https://github.com/nginx/unit/issues/653>
Link: ruby bug <https://bugs.ruby-lang.org/issues/18616>
Link: LLVM bug <https://github.com/llvm/llvm-project/issues/49958>
Commit: LLVM: Add -fdeclspec `<d170c4b57a91adc74ca89c6d4af616a00323b12c>`
Commit: ruby: Use -fdeclspec `<0958e19ffb047781fe1506760c7cbd8d7fe74e57>`

### DIFF
--- a/auto/modules/ruby
+++ b/auto/modules/ruby
@@ -56,6 +56,7 @@ nxt_found=no
 
 if /bin/sh -c "$NXT_RUBY -v" >> $NXT_AUTOCONF_ERR 2>&1; then
 
+    NXT_RUBY_CFLAGS=
     NXT_RUBY_RUBYHDRDIR=`$NXT_RUBY -rrbconfig -e 'print RbConfig::CONFIG["rubyhdrdir"]'`
     NXT_RUBY_ARCHHDRDIR=`$NXT_RUBY -rrbconfig -e 'print RbConfig::CONFIG["rubyarchhdrdir"]'`
     NXT_RUBY_SITEDIR=`$NXT_RUBY -rrbconfig -e 'print RbConfig::CONFIG["sitedir"]'`
@@ -71,6 +72,31 @@ if /bin/sh -c "$NXT_RUBY -v" >> $NXT_AUTOCONF_ERR 2>&1; then
     NXT_RUBY_LIBSCONF=`$NXT_RUBY -rrbconfig -e 'print RbConfig::CONFIG["LIBS"]'`
     NXT_RUBY_LIBPATH=`$NXT_RUBY -rrbconfig -e 'print RbConfig::CONFIG["libdir"]'`
     NXT_RUBY_LIBS="-l$NXT_RUBY_LIBNAME $NXT_RUBY_LIBSCONF"
+
+    if [ $NXT_CC_NAME = clang ]; then
+        # Workaround Clang bug
+        nxt_feature="-fdeclspec"
+        nxt_feature_name=
+        nxt_feature_run=
+        nxt_feature_incs="-fdeclspec"
+        nxt_feature_libs=
+        nxt_feature_test="#include <stdlib.h>
+
+                          __declspec(noreturn) static void f(void);
+
+                          static void f(void) {
+                              exit(0);
+                          }
+
+                          int main(void) {
+                              f();
+                          }"
+        . auto/feature
+
+        if [ $nxt_found = yes ]; then
+            NXT_RUBY_CFLAGS="$NXT_RUBY_CFLAGS -fdeclspec"
+        fi
+    fi
 
     nxt_feature="Ruby library"
     nxt_feature_name=""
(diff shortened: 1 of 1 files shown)


auto/cc/test
```diff
@@ -156,6 +156,21 @@ case $NXT_CC_NAME in
...
+                         | sed 's/.*version [0-9]*\.\([0-9]*\).*/\1/'`
+
+        if [ "$nxt_cc_clang_major" -ge 4 ]; then
+            NXT_CFLAGS="$NXT_CFLAGS -fdeclspec"
+        else
+        if [ "$nxt_cc_clang_minor" -eq 3 ] && [ "$nxt_cc_clang_minor" -ge 8 ];
+        then
+            NXT_CFLAGS="$NXT_CFLAGS -fdeclspec"
+        fi
+        fi
```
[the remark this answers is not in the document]
alejandro-colomar: Hi Valentin,

Yes, I think in the long term we should actually test the flag, but since there's no script doing this in `auto/`, we will need to add a new one, which will be a bit more complex than this.

Do you agree that we should release this easy workaround now, and prepare a better solution later?

Thanks,

Alex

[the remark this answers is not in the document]
VBart: `auto/feature` does the test, since it compiles the code and `NXT_TEST_CFLAGS` can be used for adding an option during the test.

docs/changes.xml
```diff
@@ -63,6 +63,13 @@ object, notably with Falcon.
 </para>
 </change>
 
+<change type="bugfix">
+<para>
+an issue in Clang 3.8 and later made the ruby module fail to configure
+in some systems.
+</para>
+</change>
```
VBart, line 71: It doesn't look like a Unit bug at all, so listing it as a "bugfix" seems wrong. As a feature, it looks too insignificant (especially knowing that it can be workarounded in early versions too by manually adding of relevant flag). More like a sugar for package maintainers and enthusiasts who build from sources.

I suggest to not bloat changelog by such trivial improvement.

auto/cflags
```diff
@@ -0,0 +1,29 @@
...
+
+                      int main(void) {
+                          f();
+                      }"
+    . auto/feature
+
+    if [ $nxt_found = yes ]; then
+        NXT_CFLAGS="$NXT_CFLAGS -fdeclspec"
+    fi
+fi
```
VBart, line 29: Do we have a good reason to add it globally rather than only for building Ruby module (and therefore only check it when configuring Ruby module)?

alejandro-colomar: No good reason per se.

But are we still in time for appending to the `build/Makefile`'s `CFLAGS` at the time of configuring a module?

alejandro-colomar: No good reason per se.

But are we still in time to append to `build/Makefile`'s `CFLAGS` at the time of configuring modules?

VBart: @alejandro-colomar See `NXT_PERL_CFLAGS` for example. Makefile for Ruby module is generated in the same file where relevant configure tests happen (`auto/modules/ruby`), so it's easy to append additional flags.

VBart: As I understand, you don't need to append it to `build/Makefile`'s `CFLAGS` for building whole Unit, but only for building the module itself.

alejandro-colomar: > As I understand, you don't need to append it to `build/Makefile`'s `CFLAGS` for building whole Unit, but only for building the module itself.

You're right. I don't.

alejandro-colomar: Okay, done.  (Only tradeoff/regression wiith respect to the previous implementation (i.e., global `NXT_CFLAGS`), is that now we don't have the test inside a check that the compiler is clang).  But overall, I think it's better: it's restricted to a much smaller part of the build system.

alejandro-colomar: I haven't been able to test this yet due to a consistent inability to configure the ruby module (something related to the library path, but let me investigate that tomorrow).

VBart: >  we don't have the test inside a check that the compiler is clang

Hm.. but why can't we use this check in this case? `configure` for any module can't be called before `configure` for the whole Unit and `NXT_CC_NAME` is saved into the state file (see `auto/save`) and then loaded with `. $NXT_AUTOCONF_DATA`.

alejandro-colomar: Didn't know about `NXT_AUTOCONF_DATA`.  Thanks.

alejandro-colomar: BTW, I found that the problem to configure is only with Clang... again.

The issue this time is that it doesn't seem to understand `__attribute__((__error__ ("msg")))`.  Will probably open an issue with it.

alejandro-colomar:
@travisbell, can you check that this branch works for you?

Issue #578, opened 2021-09-10
HTTP redirect to HTTPS


olszakjakub:
Hi, is there any easy way to redirect HTTP traffic to HTTPS? I didn't find any official way of doing that and only this workaround: https://github.com/nginx/unit/issues/353. Am I missing something or isn't it an obvious thing?

VBart:
See here: https://unit.nginx.org/configuration/#instant-responses-redirects - isn't it what you're looking for?

olszakjakub:
That was it. Dunno how i missed it. Thanks.

jansaputra:
That doesn't really solve the problem completely, right? We can't do resource based redirects like in the workaround in https://github.com/nginx/unit/issues/353 . So for example, one often wants to redirect something like http://foo.com/abc/def to https://foo.com/abc/def . Is there any way to do that without writing a helper application?

alejandro-colomar:
`Add variables in "location"` didn't fully solve this issue, since we also needed `$request_uri`.  `$uri` is problematic with location, because its contents are decoded, while location expects encoded URIs.  I'm reopening, and linking against the patch adding `$request_uri`.

Footnote:  Do NOT use `$uri` in `"location"` unless you really understand what you're doing.

alejandro-colomar:
`Var: Added $request_uri (as in NGINX)` was merged.

PR #660, opened 2022-03-17
Add variables in "location"

alejandro-colomar:
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

location = ":/?#[]@!$&'()*+,;=", expect = ":/?#[]@!$&'()*+,;="

    def check_location(location, expect=None):
        if expect is None:
            expect = location
    
>       assert 'success' in self.conf(
            {"return": 301, "location": location}, 'routes/0/action'
        ), 'configure location'
E       AssertionError: configure location
E       assert 'success' in {'detail': 'Unknown variable "" in the "location" value.', 'error': 'Invalid configuration.'}
E        +  where {'detail': 'Unknown variable "" in the "location" value.', 'error': 'Invalid configuration.'} = <bound method args_handler.<locals>.args_wrapper of <test_return.TestReturn object at 0x7f8789d01b50>>({'location': ":/?#[]@!$&'()*+,;=", 'return': 301}, 'routes/0/action')
E        +    where <bound method args_handler.<locals>.args_wrapper of <test_return.TestReturn object at 0x7f8789d01b50>> = <test_return.TestReturn object at 0x7f8789d01b50>.conf

test/test_return.py:98: AssertionError
```

`$` is a special symbol there.  We might have collisions if we use variables there, don't we?

### DIFF
--- a/docs/changes.xml
+++ b/docs/changes.xml
@@ -33,6 +33,12 @@ NGINX Unit updated to 1.27.0.
 
 <change type="feature">
 <para>
+variables support in the "location" option of the "return" action.
+</para>
+</change>
+
+<change type="feature">
+<para>
 compatibility with GCC 12.
 </para>
 </change>
--- a/src/nxt_http_return.c
+++ b/src/nxt_http_return.c
@@ -8,13 +8,26 @@
 
 
 typedef struct {
-    nxt_http_status_t       status;
-    nxt_str_t               location;
+    nxt_http_status_t  status;
+    nxt_var_t          *location;
+    nxt_str_t          encoded;
+    uint8_t            loc_is_const;
 } nxt_http_return_conf_t;
+
+
+typedef struct {
+    nxt_http_action_t  *action;
+    nxt_str_t          location;
+    nxt_str_t          encoded;
+} nxt_http_return_ctx_t;
 
 
 static nxt_http_action_t *nxt_http_return(nxt_task_t *task,
     nxt_http_request_t *r, nxt_http_action_t *action);
+static nxt_int_t nxt_http_return_encode(nxt_mp_t *mp, nxt_str_t *encoded,
+    const nxt_str_t *location);
+static void nxt_http_return_send_ready(nxt_task_t *task, void *obj, void *data);
+static void nxt_http_return_var_error(nxt_task_t *task, void *obj, void *data);
 
 
 static const nxt_http_request_state_t  nxt_http_return_send_state;
@@ -24,8 +37,8 @@ nxt_int_t
 nxt_http_return_init(nxt_mp_t *mp, nxt_http_action_t *action,
     nxt_http_action_conf_t *acf)
 {
-    nxt_str_t               *loc;
-    nxt_uint_t              encode;
+    nxt_str_t               str;
+    nxt_var_t               *var;
     nxt_http_return_conf_t  *conf;
 
     conf = nxt_mp_zget(mp, sizeof(nxt_http_return_conf_t));
@@ -38,30 +51,22 @@ nxt_http_return_init(nxt_mp_t *mp, nxt_http_action_t *action,
 
     conf->status = nxt_conf_get_number(acf->ret);
 
-    if (acf->location.length > 0) {
-        if (nxt_is_complex_uri_encoded(acf->location.start,
-                                       acf->location.length))
-        {
-            loc = nxt_str_dup(mp, &conf->location, &acf->location);
-            if (nxt_slow_path(loc == NULL)) {
-                return NXT_ERROR;
-            }
+    if (acf->location.length == 0) {
+        conf->loc_is_const = 1;
+        return NXT_OK;
+    }
 
-        } else {
-            loc = &conf->location;
+    var = nxt_var_compile(&acf->location, mp, 0);
+    if (nxt_slow_path(var == NULL)) {
+        return NXT_ERROR;
+    }
 
-            encode = nxt_encode_complex_uri(NULL, acf->location.start,
-                                            acf->location.length);
-            loc->length = acf->location.length + encode * 2;
+    conf->location = var;
+    conf->loc_is_const = nxt_var_is_const(var);
 
-            loc->start = nxt_mp_nget(mp, loc->length);
-            if (nxt_slow_path(loc->start == NULL)) {
-                return NXT_ERROR;
-            }
-
-            nxt_encode_complex_uri(loc->start, acf->location.start,
-                                   acf->location.length);
-        }
+    if (conf->loc_is_const) {
+        nxt_var_raw(conf->location, &str);
+        return nxt_http_return_encode(mp, &conf->encoded, &str);
     }
 
     return NXT_OK;
@@ -72,13 +77,21 @@ nxt_http_action_t *
 nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
     nxt_http_action_t *action)
 {
-    nxt_http_field_t        *field;
+    nxt_int_t               ret;
+    nxt_str_t               loc;
+    nxt_http_return_ctx_t   *ctx;
     nxt_http_return_conf_t  *conf;
 
     conf = action->u.conf;
 
-    nxt_debug(task, "http return: %d (loc: \"%V\")",
-              conf->status, &conf->location);
+    if (conf->location == NULL) {
+        nxt_str_null(&loc);
+
+    } else {
+        nxt_var_raw(conf->location, &loc);
+    }
+
+    nxt_debug(task, "http return: %d (loc: \"%V\")", conf->status, &loc);
 
     if (conf->status >= NXT_HTTP_BAD_REQUEST
         && conf->status <= NXT_HTTP_SERVER_ERROR_MAX)
@@ -87,27 +100,123 @@ nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
         return NULL;
     }
 
+    ctx = nxt_mp_zget(r->mem_pool, sizeof(nxt_http_return_ctx_t));
+    if (nxt_slow_path(ctx == NULL)) {
+        goto fail;
+    }
+
+    ctx->action = action;
     r->status = conf->status;
     r->resp.content_length_n = 0;
 
-    if (conf->location.length > 0) {
+    if (conf->loc_is_const) {
+        ctx->encoded = conf->encoded;
+
+        nxt_http_return_send_ready(task, r, ctx);
+
+    } else {
+        ret = nxt_var_query_init(&r->var_query, r, r->mem_pool);
+        if (nxt_slow_path(ret != NXT_OK)) {
+            goto fail;
+        }
+
+        nxt_var_query(task, r->var_query, conf->location, &ctx->location);
+
+        nxt_var_query_resolve(task, r->var_query, ctx,
+                              nxt_http_return_send_ready,
+                              nxt_http_return_var_error);
+    }
+
+    return NULL;
+
+fail:
+
+    nxt_http_request_error(task, r, NXT_HTTP_INTERNAL_SERVER_ERROR);
+    return NULL;
+}
+
+
+static nxt_int_t
+nxt_http_return_encode(nxt_mp_t *mp, nxt_str_t *encoded,
+    const nxt_str_t *location)
+{
+    nxt_uint_t  encode;
+
+    if (nxt_is_complex_uri_encoded(location->start, location->length)) {
+        *encoded = *location;
+
+        return NXT_OK;
+    }
+
+    encode = nxt_encode_complex_uri(NULL, location->start, location->length);
+    encoded->length = location->length + encode * 2;
+
+    encoded->start = nxt_mp_nget(mp, encoded->length);
+    if (nxt_slow_path(encoded->start == NULL)) {
+        return NXT_ERROR;
+    }
+
+    nxt_encode_complex_uri(encoded->start, location->start, location->length);
+
+    return NXT_OK;
+}
+
+
+static void
+nxt_http_return_send_ready(nxt_task_t *task, void *obj, void *data)
+{
+    nxt_int_t               ret;
+    nxt_http_field_t        *field;
+    nxt_http_action_t       *action;
+    nxt_http_request_t      *r;
+    nxt_http_return_ctx_t   *ctx;
+    nxt_http_return_conf_t  *conf;
+
+    r = obj;
+    ctx = data;
+    action = ctx->action;
+    conf = action->u.conf;
+
+    if (!conf->loc_is_const) {
+        ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
+                                     &ctx->location);
+        if (nxt_slow_path(ret == NXT_ERROR)) {
+            goto fail;
+        }
+    }
+
+    if (ctx->encoded.length > 0) {
         field = nxt_list_zero_add(r->resp.fields);
         if (nxt_slow_path(field == NULL)) {
-            nxt_http_request_error(task, r, NXT_HTTP_INTERNAL_SERVER_ERROR);
-            return NULL;
+            goto fail;
         }
 
         nxt_http_field_name_set(field, "Location");
 
-        field->value = conf->location.start;
-        field->value_length = conf->location.length;
+        field->value = ctx->encoded.start;
+        field->value_length = ctx->encoded.length;
     }
 
     r->state = &nxt_http_return_send_state;
 
     nxt_http_request_header_send(task, r, NULL, NULL);
 
-    return NULL;
+    return;
+
+fail:
+
+    nxt_http_request_error(task, r, NXT_HTTP_INTERNAL_SERVER_ERROR);
+}
+
+
+static void
+nxt_http_return_var_error(nxt_task_t *task, void *obj, void *data)
+{
+    nxt_http_request_t  *r;
+
+    r = obj;
+
+    nxt_http_request_error(task, r, NXT_HTTP_INTERNAL_SERVER_ERROR);
 }
 
 
--- a/test/test_return.py
+++ b/test/test_return.py
@@ -83,7 +83,7 @@ Connection: close
         assert resp['body'] == ''
 
     def test_return_location(self):
-        reserved = ":/?#[]@!$&'()*+,;="
+        reserved = ":/?#[]@!&'()*+,;="
         unreserved = (
             "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
             "0123456789-._~"
@@ -107,15 +107,15 @@ Connection: close
         check_location(reserved)
 
         # After first "?" all other "?" encoded.
-        check_location("/?" + reserved, "/?:/%3F#[]@!$&'()*+,;=")
+        check_location("/?" + reserved, "/?:/%3F#[]@!&'()*+,;=")
         check_location("???", "?%3F%3F")
 
         # After first "#" all other "?" or "#" encoded.
-        check_location("/#" + reserved, "/#:/%3F%23[]@!$&'()*+,;=")
+        check_location("/#" + reserved, "/#:/%3F%23[]@!&'()*+,;=")
         check_location("##?#?", "#%23%3F%23%3F")
 
         # After first "?" next "#" not encoded.
-        check_location("/?#" + reserved, "/?#:/%3F%23[]@!$&'()*+,;=")
+        check_location("/?#" + reserved, "/?#:/%3F%23[]@!&'()*+,;=")
         check_location("??##", "?%3F#%23")
         check_location("/?##?", "/?#%23%3F")
 
--- a/test/test_return.py
+++ b/test/test_return.py
@@ -161,6 +161,23 @@ Connection: close
         ), 'location method not allowed'
         assert self.get()['headers']['Location'] == 'blah'
 
+        assert 'success' in self.conf(
+            '"https://${host}${uri}"', 'routes/0/action/location'
+        ), 'location with variables'
+        assert self.get()['headers']['Location'] == 'https://localhost/'
+
+        assert 'success' in self.conf(
+            '"/#$host"', 'routes/0/action/location'
+        ), 'location with encoding and a variable'
+        assert self.get()['headers']['Location'] == '/#localhost'
+
+        assert (
+            self.get(headers={"Host": "#foo?bar", "Connection": "close"})[
+                'headers'
+            ]['Location']
+            == "/#%23foo%3Fbar"
+        ), 'location with a variable with encoding'
+
     def test_return_invalid(self):
         def check_error(conf):
             assert 'error' in self.conf(conf, 'routes/0/action')
@@ -171,6 +188,8 @@ Connection: close
         check_error({"return": 1000})
         check_error({"return": -1})
         check_error({"return": 200, "share": "/blah"})
+        check_error({"return": 200, "location": "$hos"})
+        check_error({"return": 200, "location": "$hostblah"})
 
         assert 'error' in self.conf(
             '001', 'routes/0/action/return'
--- a/src/nxt_conf_validation.c
+++ b/src/nxt_conf_validation.c
@@ -635,6 +635,7 @@ static nxt_conf_vldt_object_t  nxt_conf_vldt_return_action_members[] = {
     }, {
         .name       = nxt_string("location"),
         .type       = NXT_CONF_VLDT_STRING,
+        .flags      = NXT_CONF_VLDT_VAR,
     },
 
     NXT_CONF_VLDT_END


alejandro-colomar:
Related: https://github.com/nginx/unit/issues/578

alejandro-colomar:
I've been reading a bit about `$` in URIs, and it seems to be a reserved character for use in URIs as a subdelimiter.  Although no RFC seems to be actually using the character (they only reserve it, so that it cannot appear anywhere in a URI, but just as a subdelimiter), it seems that an application could make use of it, and we shouldn't dissallow it, so my impression is that we cannot use native Unit variables in an URI.

I've found that the man page for uri(7) is a bit outdated, and filed a few bug to update it:

- <https://bugzilla.kernel.org/show_bug.cgi?id=215735>
- <https://bugzilla.kernel.org/show_bug.cgi?id=215736>
- <https://bugzilla.kernel.org/show_bug.cgi?id=215737>
- <https://bugzilla.kernel.org/show_bug.cgi?id=215738>

That'll make me read a bunch of RFCs and know a bit more about the subject.

alejandro-colomar:
That's what I want to know.  I'll try to find out reading all those RFCs. :)

Cc: @tippexs

src/nxt_http_return.c
```diff
@@ -38,30 +36,9 @@ nxt_http_return_init(nxt_mp_t *mp, nxt_http_action_t *action,
...
-            loc->length = acf->location.length + encode * 2;
-
-            loc->start = nxt_mp_nget(mp, loc->length);
-            if (nxt_slow_path(loc->start == NULL)) {
-                return NXT_ERROR;
-            }
-
-            nxt_encode_complex_uri(loc->start, acf->location.start,
-                                   acf->location.length);
-        }
```
VBart, line 64: Previously this encoding was done once during reconfiguration phase. Now it happens on every request.

Common approach here is to move this code into a separate function and then do encoding (if needed) on the reconfiguration phase if option doesn't contain variables, or do it during in runtime after interpolation.

VBart:
> That's what I want to know. I'll try to find out reading all those RFCs. :)
> 
> But knowing that the return directive is already using it, I guess either it's a small bug that no one noticed because no one uses `$` for a special meaning, or it's not a bug.

You're right, `$` is almost never used as a sub-delimiter.

> 
> `$` percent encoded is mandatory for dollars in non-delimiter parts of the URI. However, if it's used as a subdelimiter, it should not be percent encoded, as far as I could read.

Even if there will be some rare case when somebody will need literal non-encoded `$` sign, then we can introduce a variable that returns it, like `$sign`.

src/nxt_http_return.c
alejandro-colomar: Thanks.  I was concerned that it was very bad for performance the way I did it, but I wasn't sure how to do it better.  Which is the most similar code to look at?  share?  Or is there some simpler case?

hongzhidao: "chroot" option is a similar example.

alejandro-colomar: Thanks for the tips.  Now I truly understand how it works and am very happy with the code!  I used a mix of "pass", "share", and "chroot" tounderstand it :).

alejandro-colomar:
@VBart , @hongzhidao 

Now I have a small question (noted in the commit message too), but the implementation is ready, and can be reviewed regardless of that (since the related lines of code can just be removed without consequences).

> Should we send an empty location header?  Or should we completely
> skip an empty location header?
> 
> I did the second, do not send an empty location header, because
> that's what was implemented in the old code, but I have a few
> concerns:  Is it legal to send a redirect code (30x) without a
> location URI?  Also, it adds a few lines of code, which
> considerably optimize the case of empty location, which is
> unlikely, and I can't think of a real case when it's useful, at
> the cost of a small (couple of operations) in the common case.

src/nxt_http_return.c
tippexs: Based on the [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-6.4.2) the server SHOULD return a location header and the Client MAY use it to redirect the user. I follow your idea of NOT sending an empty HTTP Location header but do not send any.

docs/changes.xml
```diff
@@ -31,6 +31,12 @@ NGINX Unit updated to 1.27.0.
          date="" time=""
          packager="Andrei Belov &lt;defan@nginx.com&gt;">
 
+<change type="feature">
+<para>
+supporting variables in the "location" URI option in the "return" action.
```
tippexs, line 36: supporting variables in the "location" URI option AND in the "return" action.

test/test_return.py
```diff
@@ -161,6 +161,14 @@ def test_return_location_edit(self):
         ), 'location method not allowed'
         assert self.get()['headers']['Location'] == 'blah'
 
+        assert 'success' in self.conf(
+            {"return": 302, "location": "foo${host}baz"}, 'routes/0/action'
```
tippexs, line 165: For me an Integration Test always helps to understand how the software engineers behind the test wants the software to be used. In this case we are using variable values like `foo` `baz` and `blahh` which technically works totally fine. BUT just think about the test with this code
```python
 {"return": 302, "location": "http://${host}"}, 'routes/0/action'
```
and the assert
```python
 assert (
            self.get(headers={"Host": "foobaz.local", "Connection": "close"}
            )['headers']['Location'] == 'http://foobaz.local'
        ), 'location with a variable'
```
This would showcase and test a real world implementation using an protocol handle in the URI and shows the users how to use the location correctly.

From my point of view the integration test needs to be as close on a real world use case as possible.

tippexs (approved):

docs/changes.xml
lcrilly: But "location" belongs to "return", so the AND is confusing. I suggest:
```supporting variables in the "location" option of the "return" action.```

File: src/nxt_http_return.c [hunk 4]
```diff
@@ -38,30 +50,16 @@ nxt_http_return_init(nxt_mp_t *mp, nxt_http_action_t *action,
...
+    conf->location = nxt_var_compile(&acf->location, mp, 0);
```
hongzhidao, line 53: It's better to keep previous main logic.

```
if (acf->location.length > 0) {
     conf->location = nxt_var_compile(&acf->location, mp, 0);
}
```
Then, the condition `conf->location != NULL` can be used on runtime phase.

src/nxt_http_return.c
```diff
@@ -72,13 +70,16 @@ nxt_http_action_t *
...
+    nxt_str_t               loc;
+    nxt_str_t               *enc;
+    nxt_http_return_ctx_t   *ctx;
     nxt_http_return_conf_t  *conf;
 
     conf = action->u.conf;
 
-    nxt_debug(task, "http return: %d (loc: \"%V\")",
-              conf->status, &conf->location);
+    nxt_var_raw(conf->location, &loc);
```
hongzhidao, line 81: You may need to test in debug mode.

src/nxt_http_return.c
```diff
@@ -87,27 +88,139 @@ nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
...
+    nxt_http_action_t       *action;
+    nxt_http_request_t      *r;
+    nxt_http_return_ctx_t   *ctx;
+    nxt_http_return_conf_t  *conf;
+
+    r = obj;
+    ctx = data;
+    action = ctx->action;
+    conf = action->u.conf;
+
```
hongzhidao, line 173: It's just a suggestion.
```
if (conf->encoded.length > 0) {
     // add location header.
}

if (conf->location != NULL && ctx->location.length > 0) {
         nxt_http_return_encode(r->mem_pool, &encoded, &ctx->location);

         // add location header.
}
```

It seems `conf->loc_is_const` and `ctx->encoded` can be elimiated.

src/nxt_http_return.c
```diff
@@ -87,27 +88,139 @@ nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
         return NULL;
     }
 
+    ctx = nxt_mp_zget(r->mem_pool, sizeof(*ctx));
```
hongzhidao, line 91: The main logic will be:

```
if (conf->location != NULL) {
     ctx = ....
}
```

src/nxt_http_return.c
```diff
@@ -87,27 +88,139 @@ nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
...
 
     nxt_http_request_header_send(task, r, NULL, NULL);
 
-    return NULL;
+    return;
+}
+
+
+static void
+nxt_http_return_var_error(nxt_task_t *task, void *obj, void *data)
```
hongzhidao, line 217: I suggest renaming `nxt_http_return_send_error` instead of `nxt_http_return_var_error`.
I also need to rename similar function in other files in my related patches.

src/nxt_http_return.c
alejandro-colomar: Thanks for finding the RFC.  So it allows us to not send an HTTP Location header.  That's good.

But I still wonder the following:

Is there any real-world scenario where a server might have an empty Location string?
We're optimizing for it, and even if it just adds a couple of operations in the commit case (to check if the string is empty a couple of times), the code would be cleaner without that optimization.

However, if you think that it might be useful to have an empty location string for some reason, yes, let's optimise.

docs/changes.xml
alejandro-colomar: I like @lcrilly 's line.

test/test_return.py
alejandro-colomar: Thanks, will try to make it more meaningful.

src/nxt_http_return.c
alejandro-colomar: Yeah, maybe with the other logic that you put above I can remove it.  Will try.

src/nxt_http_return.c
alejandro-colomar: I prefer to keep it as `_var_error()` for now for consistency, as is used in other parts of the code (`"chroot"` and `"share"`), and then we can add a separate commit to rename them all at once.  What do you think?

hongzhidao: Agree, it makes sense.

src/nxt_http_return.c
lcrilly: Location must be a URI and a URI may not be empty. In the case of a 201 response, an empty Location header has the same meaning as no Location header.

test/test_return.py
```diff
@@ -161,6 +161,14 @@ def test_return_location_edit(self):
...
+        ), 'configure location with a variable'
+        assert (
+            self.get(headers={"Host": "bar", "Connection": "close"}
+            )['headers']['Location'] == 'foobarbaz'
+        ), 'location with a variable'
+
     def test_return_invalid(self):
```
andrey-zelenkov, line 172: I would suggest to add tests with 1) variable + encoding 2) something like `#...$var...` where `$var` contains `#?` symbols (to check that `#?` will be encoded).

src/nxt_http_return.c
```diff
@@ -87,27 +88,139 @@ nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
...
+    field->value = ctx->encoded.start;
+    field->value_length = ctx->encoded.length;
+
+send_state:
     r->state = &nxt_http_return_send_state;
 
     nxt_http_request_header_send(task, r, NULL, NULL);
 
-    return NULL;
+    return;
```
andrey-zelenkov, line 212: I was just thinking of removing this label (`send_state`) and moving all logic into `if` block since it's used only once. And/or creating label (`send_error`?) to remove repeating lines: 
```        
nxt_http_request_error(task, r, NXT_HTTP_INTERNAL_SERVER_ERROR);
return;
```
What do you think?

hongzhidao: >I was just thinking of removing this label (send_state) and moving all logic into if block since it's used only once.

Right.

src/nxt_http_return.c
alejandro-colomar: > Location must be a URI and a URI may not be empty.

That makes a lot of sense.

src/nxt_http_return.c
alejandro-colomar: > And/or creating label (`send_error`?) to remove repeating lines:
> 
> ```
> nxt_http_request_error(task, r, NXT_HTTP_INTERNAL_SERVER_ERROR);
> return;
> ```
> 
> What do you think?

I like the idea of centralizing the errors.

src/nxt_http_return.c
alejandro-colomar: Should that path be `nxt_slow_path()`?

hongzhidao: Nope, just the same as before.

src/nxt_http_return.c
alejandro-colomar: Actually I just realized that I had done a redundant copy here.

src/nxt_http_return.c
alejandro-colomar: The obsolete RFC 2616 defined the Location to be an absolute URI <https://www.rfc-editor.org/rfc/rfc2616#section-14.30>, which as you say, cannot be an empty string.

However, the current RFC allows the Location to be a relative URI <https://www.rfc-editor.org/rfc/rfc7231#section-7.1.2>, and a relative URI might be the empty string.

src/nxt_http_return.c
alejandro-colomar: I changed it considerably.  It still uses that variable, however.  Otherwise I needed to add more code.

We might consider using `nxt_var_is_const()` each time if you want to remove the variable from the structure, but I think letting `conf->location` be `NULL` adds more code than it removes.

src/nxt_http_return.c
alejandro-colomar: I'll mark this as resolved, since it won't be changed in this patch set.  I'll start a separate discussion for it.

alejandro-colomar: https://github.com/nginx/unit/issues/674

src/nxt_http_return.c
hongzhidao: If users don't set "location", it shouldn't call anything in "location", like `nxt_http_return_encode()`. But it seems it's called when "location" is not set.

src/nxt_http_return.c
```diff
@@ -87,27 +88,130 @@ nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
...
+        goto fail;
+    }
+
+    ctx->action = action;
     r->status = conf->status;
     r->resp.content_length_n = 0;
 
-    if (conf->location.length > 0) {
+    if (conf->loc_is_const) {
+        enc = nxt_str_dup(r->mem_pool, &ctx->encoded, &conf->encoded);
```
hongzhidao, line 101: I suggest that.
```
ctx->encoded = conf->encoded;
```

src/nxt_http_return.c
alejandro-colomar: Oops, that's right.

alejandro-colomar: I'm a bit confused about the previous code:

<https://github.com/nginx/unit/blob/aeed86c6829c62359e79f239b849766efb8857a7/src/nxt_http_return.c#L41>

Isn't `acf->location` supposed to be `NULL` if the user doesn't set `"location"`?

src/nxt_http_return.c
alejandro-colomar: Yup.  I didn't know if I could do that.  If you see any other copies that can be replaced by a simple assignment, please tell me.  When do I need to `nxt_str_dup()`?

hongzhidao: For me, I usually check it by their life cycle and if it will be changed. For example, `conf->encoded` has a longer life cycle than `ctx->encoded`, and `ctx->encoded` doesn't change after that.

alejandro-colomar: Thanks.  I'll replace the other one in line 144 too.

src/nxt_http_return.c
```diff
@@ -87,27 +97,123 @@ nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
         return NULL;
     }
 
+    ctx = nxt_mp_zget(r->mem_pool, sizeof(*ctx));
```
hongzhidao, line 100: We usually use structure type.
```
ctx = nxt_mp_zget(r->mem_pool, sizeof(nxt_http_return_ctx_t));
```

src/nxt_http_return.c
```diff
@@ -87,27 +97,123 @@ nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
...
+    nxt_http_request_t      *r;
+    nxt_http_return_ctx_t   *ctx;
+    nxt_http_return_conf_t  *conf;
+
+    r = obj;
+    ctx = data;
+    action = ctx->action;
+    conf = action->u.conf;
+
+    if (!conf->loc_is_const) {
```
hongzhidao, line 177: What happens if the resolved variable value is empty?

alejandro-colomar: You mean this?:

```json
"action": {
    "return": 301,
    "location": "$host"
}
```
```
$ curl --dump-header - -H "Host: " localhost:82
HTTP/1.1 301 Moved Permanently
Server: Unit/1.27.0
Date: Thu, 07 Apr 2022 23:30:29 GMT
Content-Length: 0
```

The header will be skipped, as shown above.

Rationale:

http://www.alejandro-colomar.es/src/alx/nginx/unit.git/tree/src/nxt_http_return.c?h=location-var#n177

`$ find src -type f | xargs sed -n '/^nxt_http_return_send_ready/,/^}/p'`
```c
nxt_http_return_send_ready(nxt_task_t *task, void *obj, void *data)
{
    nxt_int_t               ret;
    nxt_http_field_t        *field;
    nxt_http_action_t       *action;
    nxt_http_request_t      *r;
    nxt_http_return_ctx_t   *ctx;
    nxt_http_return_conf_t  *conf;

    r = obj;
    ctx = data;
    action = ctx->action;
    conf = action->u.conf;

    // If there's a variable, it will enter this.
    if (!conf->loc_is_const) {
        // See below for details, but this just copies around in the case of an empty string.  No error.
        ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
                                     &ctx->location);
            goto fail;
        }
    }

    // Since the empty string has length 0, don't add a header.  Skip this.
    if (ctx->encoded.length > 0) {
        field = nxt_list_zero_add(r->resp.fields);
        if (nxt_slow_path(field == NULL)) {
            goto fail;
        }

        nxt_http_field_name_set(field, "Location");

        field->value = ctx->encoded.start;
        field->value_length = ctx->encoded.length;
    }


    // Basically it did nothing until here.  Only a few unused copies.
    r->state = &nxt_http_return_send_state;

    nxt_http_request_header_send(task, r, NULL, NULL);

    return;

fail:

    nxt_http_request_error(task, r, NXT_HTTP_INTERNAL_SERVER_ERROR);
}
```

`$ find src -type f | xargs sed -n '/^nxt_http_return_encode/,/^}/p'`
```
nxt_http_return_encode(nxt_mp_t *mp, nxt_str_t *encoded,
    const nxt_str_t *location)
{
    nxt_uint_t  encode;

    // The following check will return true, so it will just copy the string structure and exit successfully.  See below for details.
    if (nxt_is_complex_uri_encoded(location->start, location->length)) {
        *encoded = *location;

        return NXT_OK;
    }

    encode = nxt_encode_complex_uri(NULL, location->start, location->length);
    encoded->length = location->length + encode * 2;

    encoded->start = nxt_mp_nget(mp, encoded->length);
    if (nxt_slow_path(encoded->start == NULL)) {
        return NXT_ERROR;
    }

    nxt_encode_complex_uri(encoded->start, location->start, location->length);

    return NXT_OK;
}
```

`$ find src -type f | xargs sed -n '/^nxt_is_complex_uri_encoded/,/^}/p'`
```
nxt_is_complex_uri_encoded(u_char *src, size_t length)
{
    u_char   *reserved, *end, ch;
    uint8_t  d0, d1;

    reserved = (u_char *) "?#\0";

    // If the string is empty, length == 0, and end == src, so skip this loop and return true.
    for (end = src + length; src < end; src++) {
        ch = *src;

        if (nxt_uri_escape[ch >> 5] & (1U << (ch & 0x1f))) {
            if (ch == '%') {
                if (end - src < 2) {
                    return 0;
                }

                d0 = nxt_hex2int[*++src];
                d1 = nxt_hex2int[*++src];

                if ((d0 | d1) >= 16) {
                    return 0;
                }

                continue;
            }

            if (ch == reserved[0]) {
                reserved++;
                continue;
            }

            if (ch == reserved[1]) {
                reserved += 2;
                continue;
            }

            return 0;
        }
    }

    return 1;
}
```

hongzhidao: It's possible that `ctx->location.length` is NULL.
For example: `"location": "$arg_foo"`. In the future, we will support var like `$arg_*`.
```
// If there's a variable, it will enter this.
    if (!conf->loc_is_const) {
        // See below for details, but this just copies around in the case of an empty string.  No error.
        ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
                                     &ctx->location);
            goto fail;
        }
    }
```

src/nxt_http_return.c
alejandro-colomar: Hmm yes.  That's something I don't really like.

I think it's much safer to use the pointer itself to get the size.  That way you can never make a mistake, since it's straightforward (to the point that I have a macro that does it for me: `err = alx_mallocarrays(&p, n);` which basically resolves to `p = malloc(sizeof(*p) * n); err = (p == NULL);` plus some overflow checking).  However, if you use the type, and you ever change the type of the pointer, you need to update _every_ allocation line.

However, for consistency, I'll use the type here.  I'll propose in a different thread to revise the allocations.

<https://software.codidact.com/posts/285898/285899#answer-285899>
<https://stackoverflow.com/a/605858/6872717>

src/nxt_http_return.c
alejandro-colomar: > It's possible that `ctx->location.length` is 0. For example: `"location": "$arg_foo"`. In the future, we will support var like `$arg_*`.

If you mean that the content of the variable is empty, that's no problem.  It's supported by the code, as shown above.  It simply sends no header (keeping the old behavior of considering the following equivalent, which I think is incorrect; but I think we should fix that in a different patch set, not here):

```
"action": {
    "return": 301
}
```
```
"action": {
    "return": 301,
    "location": ""
}
```

Check this branch, where I have a prototype (in fact, it's already tested and works): <https://github.com/alejandro-colomar/unit/tree/location-empty>

hongzhidao:
```
// If there's a variable, it will enter this.
    if (!conf->loc_is_const) {
        // See below for details, but this just copies around in the case of an empty string.  No error.
        ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
                                     &ctx->location);
            goto fail;
        }
    }
```
Just to note if `ctx->location.length` is 0, `nxt_http_return_encode()` doesn't have to be called.

hongzhidao: > Check this branch, where I have a prototype (in fact, it's already tested and works):

I'm a bit confused now, does this patch depend on `empty location` patch for now?

alejandro-colomar: I know.  I just didn't want to optimize that case for the following reasons:

- It's unlikely that a server wants to send an empty Location header (because where would it redirect?  to itself?).
- It would add a check in the code that doesn't improve readability, nor makes the common case faster (in fact, it makes the common case ~1 op slower).

alejandro-colomar: > > Check this branch, where I have a prototype (in fact, it's already tested and works):
> 
> I'm a bit confused now, does this patch depend on `empty location` patch for now?

No.  It's the other way around.  ``location-empty` depends on this, and that's the reason I haven't published it yet (I'll do when we merge this one).

alejandro-colomar: The current patch set (aka `location`, aka `location-var`) behaves the same for an unset Location as for an empty string (`""`) as for an empty variable (`"$arg_foo"`).  It doesn't send a Location header.

Because previously, the code behaved the same for an empty string (`""`) as for an unset Location, and I wanted to keep that (slightly incorrect) behavior.  It didn't send a Location header.

_After_ we merge this, I'll propose a new patch set called `location-empty` which will behave correctly (but will not optimize) for empty strings (before or after variable resolution), by sending an empty Location header.  I will add a changelog line of type bugfix.

docs/changes.xml
```diff
@@ -31,6 +31,12 @@ NGINX Unit updated to 1.27.0.
          date="" time=""
          packager="Andrei Belov &lt;defan@nginx.com&gt;">
 
+<change type="feature">
+<para>
+supporting variables in the "location" option of the "return" action.
```
artemkonev, line 36: Previous changelog entries had it like this: `variables support in the "XXXYYY" option...`

artemkonev (changes requested):

src/nxt_http_return.c
VBart: @alejandro-colomar If you decide to change the type of pointer, then you anyway need to check every line for possible consequences. It's a must have habit no matter of how you allocate memory.

Note also, that missing `*` in such type of allocation is [a very common mistake](https://cwe.mitre.org/data/definitions/467.html). And the argument that if you use pointer, then you will never allocate wrong size won't work at the moment when you decide to change to an array type, and then having no habit to check allocation may play trick to you.

The main reason to use the type name in allocations is that number of times when someone has to find a place where a particular object is allocated by several orders of magnitude bigger than changes of its type. Currently it's easy to find where particular objects are created by `alloc.*nxt_some_type_t` lookup request. Programmers read the code much more often than change it, and having a type in alocations makes searching and reading a bit easier.

docs/changes.xml
alejandro-colomar: That's what I wrote at first for the "index" patch set, IIRC, but was suggested to write it in this other form.

Which way do you prefer?

Cc: @hongzhidao
Cc: @tippexs

src/nxt_http_return.c
alejandro-colomar: > @alejandro-colomar If you decide to change the type of pointer, then you anyway need to check every line for possible consequences. It's a must have habit no matter of how you allocate memory.

Yes, but it's easier to make mistakes in the case of a type.

> 
> Note also, that missing `*` in such type of allocation is [a very common mistake](https://cwe.mitre.org/data/definitions/467.html). And the argument that if you use pointer, then you will never allocate wrong size won't work at the moment when you decide to change to an array type, and then having no habit to check allocation may play trick to you.

I have the rule that the argument to `sizeof()` must the be exact same string to the left of the `=` with a prefix of exactly one `*`.  Since it's a text replacement rule, I created a macro that does it for me.  I (almost) always allocate arrays, of size 1 when I want a single element:

`err = alx_mallocarrays(&ctx, 1);`

A trivial definition would be:

```
#define alx_mallocarray(nmemb, size)  reallocarray(NULL, nmemb, size)

#define alx_mallocarrays(ptr, nmemb)                 \
({                                                   \
    int          err_;                               \
    __auto_type  ptr_ = (ptr);                       \
                                                     \
    *ptr_ = alx_mallocarray(nmemb, sizeof(**ptr_));  \
    err_ = (*ptr == NULL);                           \
                                                     \
    err_;                                            \
})
```

Only when I want to specify a size that is not a multiple of the type I do call malloc directly, for example when I need to use the struct hack.

This has always worked for me.

So regarding safety, I believe this macro is much safer than any kind of `sizeof` usage.  And lacking this macro, a mental rule that always works (i.e., doing mentally the same work that this macro does) is probably the safest, but not perfect.

> 
> The main reason to use the type name in allocations is that number of times when someone has to find a place where a particular object is allocated by several orders of magnitude bigger than changes of its type. Currently it's easy to find where particular objects are created by `alloc.*nxt_some_type_t` lookup request. Programmers read the code much more often than change it, and having a type in alocations makes searching and reading a bit easier.

However, that is a strong argument.  I like being able to grep with simple patterns such as that one.  And it's true that we grep much more than we edit.

hongzhidao (approved):

src/nxt_http_return.c
alejandro-colomar: I just tested in debug mode, and we have a problem when `"location"` contains an empty string:
```json
{
  "listeners": {
    ...,
    "*:81": {
      "pass": "routes/empty"
    },
    ...
  },
  "routes": {
    ...,
    "empty": [
      {
        "action": {
          "return": 301,
          "location": ""
        }
      }
    ],
    ...
  }
}
```

docs/changes.xml
alejandro-colomar: Changes applied as suggested.

src/nxt_http_return.c
```diff
@@ -72,13 +77,21 @@ nxt_http_action_t *
...
 
     conf = action->u.conf;
 
-    nxt_debug(task, "http return: %d (loc: \"%V\")",
-              conf->status, &conf->location);
+    if (conf->location != NULL) {
+        nxt_var_raw(conf->location, &loc);
+
+    } else {
+        loc = (nxt_str_t) {0};
```
alejandro-colomar, line 91: @hongzhidao, this fixes the uninitialized problem mentioned in the other thread.  Do you like the initialization this way?

hongzhidao: Take this as example, I prefer to use `nxt_str_null()` instead.
```
     if (conf->types == NULL) {
            nxt_str_null(&exten);

        } else {
            nxt_http_static_extract_extension(shr, &exten);
            mtype = nxt_htt
```

alejandro-colomar:
> @alejandro-colomar
> 
> Just received the report by Coverity Scan:
> 
> ```
> Hi,
> 
> Please find the latest report on new defect(s) introduced to nginx unit found with Coverity Scan.
> 
> 1 new defect(s) introduced to nginx unit found with Coverity Scan.
> 
> 
> New defect(s) Reported-by: Coverity Scan
> Showing 1 of 1 defect(s)
> 
> 
> ** CID 378226:  Null pointer dereferences  (FORWARD_NULL)
> 
> 
> ________________________________________________________________________________________________________
> *** CID 378226:  Null pointer dereferences  (FORWARD_NULL)
> >>>     CID 378226:  Null pointer dereferences  (FORWARD_NULL)
> >>>     Passing null pointer "conf->location" to "nxt_var_query", which dereferences it.


Unless I'm missing something, the report is invalid.  See:

```c
nxt_int_t
nxt_http_return_init(nxt_mp_t *mp, nxt_http_action_t *action,
    nxt_http_action_conf_t *acf)
{
    nxt_str_t               str;
    nxt_var_t               *var;
    nxt_http_return_conf_t  *conf;

    ...

    if (acf->location.length == 0) {
        conf->loc_is_const = 1;
        return NXT_OK;
        // This is the only valid path where conf->location is NULL
        // But notice that we set conf->loc_is_const = 1
    }

    var = nxt_var_compile(&acf->location, mp, 0);
    if (nxt_slow_path(var == NULL)) {
        return NXT_ERROR;
    }

    conf->location = var;
    ...
}


nxt_http_action_t *
nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
    nxt_http_action_t *action)
{
    nxt_int_t               ret;
    nxt_str_t               loc;
    nxt_http_return_ctx_t   *ctx;
    nxt_http_return_conf_t  *conf;

    conf = action->u.conf;

    ...

    if (conf->loc_is_const) {
        // Here we go if conf->location == NULL, since conf->loc_is_const == 1
        ctx->encoded = conf->encoded;

        nxt_http_return_send_ready(task, r, ctx);

    } else {
            goto fail;
        }

        nxt_var_query(task, r->var_query, conf->location, &ctx->location);

                              nxt_http_return_send_ready,
                              nxt_http_return_var_error);
    }

    ...
}
```

alejandro-colomar:
This patch set introduced Undefined Behavior of the form `memcpy(dest, NULL, 0)`.
See the report: <https://github.com/nginx/unit/issues/697>

Issue #642, opened 2022-02-12
X-Forwarded-For (client_ip->source) with unix:


echolimazulu:
___
### - All users of the unix socket as a listener
___

Hello Nginx Team and Contributors,

**Commit with function:** https://github.com/nginx/unit/commit/ca373aaccd276fb412e59557a3971a8d06ada0f8

It is necessary to extend the function of replacing a remote client address by introducing support for specifying the appropriate source when using a unix socket.

In case you are using a `unix:` socket in your configuration, address replacing is not possible based on the CIDR `source` address format.

The format mask: `0.0.0.0/0` is not applicable (not operational) using the unix listener.

echolimazulu:
@ocanty Thanks for the work you've done. Is it possible to extend support for this feature to unix socket users as a listener?

alejandro-colomar:
Hi @echolimazulu,

I have some initial draft of a patch set to add this feature.  Do you have some configuration with which I can test it?

Thanks,
Alex

echolimazulu:
Good morning @alejandro-colomar,

Thank you for your reaction and work on this issue!

I apologize for the long answer.

Within 1 hour, I will provide you with the Unit configuration you requested.

echolimazulu:
@alejandro-colomar, 
I am glad to help.

**/var/lib/unit/conf.json**
```
{
    "listeners": {
        "unix:///var/run/flask.sock": {
            "pass": "applications/flask"
        }
    },
    "applications": {
        "flask": {
            "type": "python 3.10",
            "path": "/opt/flask/",
            "module": "asgi_app",
            "callable": "app"
        }
    }
}
```

**/opt/falcon/things_asgi.py**
```
import falcon
import falcon.asgi


class ThingsResource:
    async def on_get(self, req, resp):
        resp.status = falcon.HTTP_200  # This is the default status
        resp.content_type = falcon.MEDIA_TEXT  # Default is JSON, so override
        resp.text = ('\nTwo things awe me most, the starry sky '
                     'above me and the moral law within me.\n'
                     '\n'
                     '    ~ Immanuel Kant\n\n')


app = falcon.asgi.App()

things = ThingsResource()

app.add_route('/things', things)

```
**/bin/bash:**
```
# curl --header "X-Forwarded-For: 10.10.10.10" --no-buffer -XGET --unix-socket /var/run/falcon.sock http://localhost/things_asgi
```

alejandro-colomar:
Thanks!

[...]

Issue #635, opened 2022-01-25
strange error when serving app from starlette


sandeep-gh:
I get 503 error when running nginx+unit serving the example starlette app (https://unit.nginx.org/howto/starlette/). 
The complete log for unit is:
```
2022/01/26 00:37:51 [info] 33216#139127 discovery started
2022/01/26 00:37:51 [notice] 33216#139127 module: python 3.10.2 "/home/shared/Builds/unit-1.26.1/modules/python3.10.unit.so"
2022/01/26 00:37:51 [info] 36261#518939 controller started
2022/01/26 00:37:51 [notice] 36261#518939 process 33216 exited with code 0
2022/01/26 00:37:51 [info] 58715#394694 router started
2022/01/26 00:37:51 [info] 58715#394694 OpenSSL 1.1.1l  24 Aug 2021, 101010cf
2022/01/26 00:37:51 [info] 93430#495270 "myapp" prototype started
2022/01/26 00:37:51 [info] 32427#513572 "myapp" application started
2022/01/26 00:38:30 [alert] 32427#513572 [unit] #8: Python failed to create 'client' pair
2022/01/26 00:38:30 [alert] 32427#513572 [unit] Python failed to call 'loop.call_soon'
Traceback (most recent call last):
  File "/home/shared/Builds/Python-3.10.2/lib/python3.10/asyncio/base_events.py", line 745, in call_soon
    self._check_closed()
ValueError: invalid literal for int() with base 10: ''
```

I am using Python3.10.2. I have nginx pointing to unit via unix socket. The error is very strange I am not able to figure what could be wrong. 
Happy to provide any more information as needed.

mar0x:
Hello,

Most likely it caused by listening on unix socket. The code in ASGI module tries to parse remote address and extract host and port from the address - this need to be fixed.

For now, please try to reconfigure Unit to listen on TCP port.

VBart:
I guess you can workaround it by configuring `client_ip`: https://unit.nginx.org/configuration/#configuration-listeners-xff

sandeep-gh:
@VBart I am not sure I follow. It would be great if there is workaround for this. Could we add support of unix sockets?

sandeep-gh:
> 
@mar0x I have switched to sockets. I don't get the error anymore. However, I am still not able to serve the pages. I get "Not Found" message Seems to me that the app is crashing. Where can I see app's output. 

Below is the log from unit:
```
[...]

PR #664, opened 2022-03-25
Array: Multiple fixes, functionality enhancements.

echolimazulu:
**fixes:**
- removed incorrect array initialization (`nxt_array_init`) with a pointer to a non-existent set of elements;
- fixed the size of the allocated memory for storing the array structure (`nxt_array_t`) - waste of memory;
- fixed initialization of a set of elements (`elts`), pointer, memory size;
- fixed the formula for calculating the required amount of memory (`nalloc`) allocated for the created array;
- eliminated the possibility of early deletion of an existing pointer to a set of elements (`elts`) at the time of memory reallocation;

**improvements:**
- added descriptions of functions;
- added a function to get a pointer by the index of an element in an array.

### DIFF
--- a/src/nxt_array.c
+++ b/src/nxt_array.c
@@ -1,5 +1,6 @@
 
 /*
+ * Copyright (C) Evgenii Sokolov
  * Copyright (C) Igor Sysoev
  * Copyright (C) NGINX, Inc.
  */
@@ -7,144 +8,246 @@
 #include <nxt_main.h>
 
 
-nxt_array_t *
-nxt_array_create(nxt_mp_t *mp, nxt_uint_t n, size_t size)
-{
-    nxt_array_t  *array;
+static void nxt_array_elts_copy(void **dst_elts, nxt_array_t *src_array);
+static nxt_uint_t nxt_array_calc_alloc(nxt_uint_t nelts);
+static void *nxt_array_elts_realloc(nxt_array_t *array,
+                                    nxt_uint_t nalloc);
 
-    array = nxt_mp_alloc(mp, sizeof(nxt_array_t) + n * size);
+
+/*
+ * nxt_array_create - Create a new array and return a pointer to it.
+ */
+nxt_array_t *
+nxt_array_create(nxt_mp_t *mp, nxt_uint_t arr_size, size_t elt_size)
+{
+    nxt_uint_t     nalloc;
+    nxt_array_t    *array;
+    void           *elts;
+
+    array = nxt_mp_alloc(mp, sizeof(nxt_array_t));
 
     if (nxt_slow_path(array == NULL)) {
         return NULL;
     }
 
-    array->elts = nxt_pointer_to(array, sizeof(nxt_array_t));
+    nalloc = nxt_array_calc_alloc(arr_size);
+    elts = nxt_mp_alloc(mp, elt_size * nalloc);
+
+    if (nxt_slow_path(elts == NULL)) {
+        return NULL;
+    }
+
+    array->elts = elts;
     array->nelts = 0;
-    array->size = size;
-    array->nalloc = n;
+    array->size = elt_size;
+    array->nalloc = nalloc;
     array->mem_pool = mp;
 
     return array;
 }
 
 
+/*
+ * nxt_array_destroy - Destroy the existing array.
+ */
 void
 nxt_array_destroy(nxt_array_t *array)
 {
-    if (array->elts != nxt_pointer_to(array, sizeof(nxt_array_t))) {
-        nxt_mp_free(array->mem_pool, array->elts);
-    }
+    if (nxt_fast_path(array != NULL)) {
 
-    nxt_mp_free(array->mem_pool, array);
+        array->nelts = 0;
+
+        (void) nxt_mp_free(array->mem_pool, array->elts);
+        (void) nxt_mp_free(array->mem_pool, array);
+
+    }
 }
 
 
+/*
+ * nxt_array_add - Add an element to the existing array.
+ */
 void *
 nxt_array_add(nxt_array_t *array)
 {
-    void      *p;
-    uint32_t  nalloc, new_alloc;
+    if (nxt_fast_path(array != NULL)) {
 
-    nalloc = array->nalloc;
+        nxt_uint_t    nalloc;
+        void          *elt, *elts;
 
-    if (array->nelts == nalloc) {
+        if (array->nelts == array->nalloc) {
 
-        if (nalloc < 16) {
-            /* Allocate new array twice larger than current. */
-            new_alloc = (nalloc == 0) ? 4 : nalloc * 2;
+            nalloc = nxt_array_calc_alloc(array->nelts);
+            elts = nxt_array_elts_realloc(array, nalloc);
 
-        } else {
-            /* Allocate new array 1.5 times larger than current. */
-            new_alloc = nalloc + nalloc / 2;
-        }
-
-        p = nxt_mp_alloc(array->mem_pool, array->size * new_alloc);
-
-        if (nxt_slow_path(p == NULL)) {
-            return NULL;
-        }
-
-        nxt_memcpy(p, array->elts, array->size * nalloc);
-
-        if (array->elts != nxt_pointer_to(array, sizeof(nxt_array_t))) {
-            nxt_mp_free(array->mem_pool, array->elts);
-        }
-
-        array->elts = p;
-        array->nalloc = new_alloc;
-    }
-
-    p = nxt_pointer_to(array->elts, array->size * array->nelts);
-    array->nelts++;
-
-    return p;
-}
-
-
-void *
-nxt_array_zero_add(nxt_array_t *array)
-{
-    void  *p;
-
-    p = nxt_array_add(array);
-
-    if (nxt_fast_path(p != NULL)) {
-        nxt_memzero(p, array->size);
-    }
-
-    return p;
-}
-
-
-void
-nxt_array_remove(nxt_array_t *array, void *elt)
-{
-    void  *last;
-
-    last = nxt_array_last(array);
-
-    if (elt != last) {
-        nxt_memcpy(elt, last, array->size);
-    }
-
-    array->nelts--;
-}
-
-
-nxt_array_t *
-nxt_array_copy(nxt_mp_t *mp, nxt_array_t *dst, nxt_array_t *src)
-{
-    void      *data;
-    uint32_t  i, size;
-
-    size = src->size;
-
-    if (dst == NULL) {
-        dst = nxt_array_create(mp, src->nelts, size);
-        if (nxt_slow_path(dst == NULL)) {
-            return NULL;
-        }
-    }
-
-    nxt_assert(size == dst->size);
-
-    if (dst->nalloc >= src->nelts) {
-        nxt_memcpy(dst->elts, src->elts, src->nelts * size);
-
-    } else {
-        nxt_memcpy(dst->elts, src->elts, dst->nelts * size);
-
-        for (i = dst->nelts; i < src->nelts; i++) {
-            data = nxt_array_add(dst);
-            if (nxt_slow_path(data == NULL)) {
+            if (nxt_slow_path(elts == NULL)) {
                 return NULL;
             }
 
-            nxt_memcpy(data, src->elts + (i * size), size);
         }
+
+        elt = nxt_pointer_to(array->elts, array->size * array->nelts);
+
+        array->nelts++;
+
+        return elt;
+
+    } else {
+        return NULL;
+    }
+}
+
+
+/*
+ * nxt_array_zero_add - Add an element to the existing array,
+ * fill with zeros.
+ */
+void *
+nxt_array_zero_add(nxt_array_t *array)
+{
+    if (nxt_fast_path(array != NULL)) {
+
+        void    *elt;
+
+        elt = nxt_array_add(array);
+
+        if (nxt_fast_path(elt != NULL)) {
+
+            (void) nxt_memzero(elt, array->size);
+
+        }
+
+        return elt;
+
+    } else {
+        return NULL;
+    }
+}
+
+
+/*
+ * nxt_array_del - Delete the specified element in the array.
+ */
+void
+nxt_array_del(nxt_array_t *array, void *elt)
+{
+    if (nxt_fast_path(array != NULL)) {
+
+        void    *last;
+
+        last = nxt_array_pointer_to_last(array);
+
+        if (elt != last) {
+            (void) nxt_memcpy(elt, last, array->size);
+        }
+
+        (void) nxt_array_del_last(array);
+
+    }
+}
+
+
+/*
+ * nxt_array_del_last - Delete the last element from
+ * the existing array.
+ */
+void
+nxt_array_del_last(nxt_array_t *array)
+{
+    if (nxt_fast_path(array != NULL)) {
+
+        array->nelts--;
+
+    }
+}
+
+
+/*
+ * nxt_array_copy - Copies the existing array to a new one
+ * and returns a pointer to it.
+ */
+nxt_array_t *
+nxt_array_copy(nxt_mp_t *mp, nxt_array_t *array)
+{
+    if (nxt_fast_path(array != NULL)) {
+
+        nxt_array_t    *new_array;
+
+        new_array = nxt_array_create(mp, array->nelts, array->size);
+
+        if (nxt_slow_path(new_array == NULL)) {
+            return NULL;
+        }
+
+        (void) nxt_array_elts_copy(new_array->elts, array);
+
+        new_array->nelts = array->nelts;
+
+        return new_array;
+
+    } else {
+        return NULL;
+    }
+}
+
+
+/*
+ * nxt_array_elts_copy - Copy existing elements from source
+ * to destination array.
+ */
+static void
+nxt_array_elts_copy(void **dst_elts, nxt_array_t *src_array)
+{
+    (void) nxt_memcpy(dst_elts, src_array->elts,
+                      src_array->size * src_array->nelts);
+}
+
+
+/*
+ * nxt_array_calc_alloc - Calculate the required size
+ * of the memory area for the specified number of elements.
+ */
+static nxt_uint_t
+nxt_array_calc_alloc(nxt_uint_t nelts)
+{
+    nxt_uint_t    nalloc;
+
+    if (nelts <= 16) {
+
+        nalloc = (nelts == 0) ? 2 : nelts * 2;
+
+    } else {
+
+        nalloc = nelts + nelts / 2;
+
     }
 
-    dst->nelts = src->nelts;
+    return nalloc;
+}
 
-    return dst;
+
+/*
+ * nxt_array_elts_realloc - Reallocate the allocated memory
+ * for the array elements and return a pointer to the new block.
+ */
+static void *
+nxt_array_elts_realloc(nxt_array_t *array, nxt_uint_t nalloc)
+{
+    void    **old_elts, *new_elts;
+
+    old_elts = array->elts;
+    new_elts = nxt_mp_alloc(array->mem_pool, array->size * nalloc);
+
+    if (nxt_slow_path(new_elts == NULL)) {
+        return NULL;
+    }
+
+    (void) nxt_array_elts_copy(new_elts, array);
+
+    array->elts = new_elts;
+    array->nalloc = nalloc;
+
+    (void) nxt_mp_free(array->mem_pool, old_elts);
+
+    return new_elts;
 }
--- a/src/nxt_array.h
+++ b/src/nxt_array.h
@@ -1,5 +1,6 @@
 
 /*
+ * Copyright (C) Evgenii Sokolov
  * Copyright (C) Igor Sysoev
  * Copyright (C) NGINX, Inc.
  */
@@ -9,34 +10,32 @@
 
 
 typedef struct {
-    void      *elts;
-    /* nelts has uint32_t type because it is used most often. */
-    uint32_t  nelts;
-    uint16_t  size;
-    uint16_t  nalloc;
-    nxt_mp_t  *mem_pool;
+    void                *elts;
+    nxt_uint_t          nelts;
+    nxt_uint_t          size;
+    nxt_uint_t          nalloc;
+    nxt_mp_t            *mem_pool;
 } nxt_array_t;
 
 
-nxt_inline void
-nxt_array_init(nxt_array_t *array, nxt_mp_t *mp, size_t size)
-{
-    array->elts = nxt_pointer_to(array, sizeof(nxt_array_t));
-    array->size = size;
-    array->mem_pool = mp;
-}
-
 NXT_EXPORT nxt_array_t *nxt_array_create(nxt_mp_t *mp, nxt_uint_t n,
-    size_t size);
+                                         size_t size);
 NXT_EXPORT void nxt_array_destroy(nxt_array_t *array);
 NXT_EXPORT void *nxt_array_add(nxt_array_t *array);
 NXT_EXPORT void *nxt_array_zero_add(nxt_array_t *array);
-NXT_EXPORT void nxt_array_remove(nxt_array_t *array, void *elt);
-NXT_EXPORT nxt_array_t *nxt_array_copy(nxt_mp_t *mp, nxt_array_t *dst,
-    nxt_array_t *src);
+NXT_EXPORT void nxt_array_del(nxt_array_t *array, void *elt);
+NXT_EXPORT void nxt_array_del_last(nxt_array_t *array);
+NXT_EXPORT nxt_array_t *nxt_array_copy(nxt_mp_t *mp,
+                                       nxt_array_t *array);
+
 
 #define                                                                       \
-nxt_array_last(array)                                                         \
+nxt_array_pointer_to_index(array, index)                                      \
+    nxt_pointer_to((array)->elts, (array)->size * (index))
+
+
+#define                                                                       \
+nxt_array_pointer_to_last(array)                                              \
     nxt_pointer_to((array)->elts, (array)->size * ((array)->nelts - 1))
 
 
@@ -48,14 +47,6 @@ nxt_array_reset(array)                                                        \
 #define                                                                       \
 nxt_array_is_empty(array)                                                     \
     ((array)->nelts == 0)
-
-
-nxt_inline void *
-nxt_array_remove_last(nxt_array_t *array)
-{
-    array->nelts--;
-    return nxt_pointer_to(array->elts, array->size * array->nelts);
-}
 
 
 #endif /* _NXT_ARRAY_H_INCLUDED_ */
--- a/src/nxt_var.c
+++ b/src/nxt_var.c
@@ -33,8 +33,8 @@ typedef struct {
 
 
 struct nxt_var_query_s {
-    nxt_array_t         values;   /* of nxt_var_value_t */
-    nxt_array_t         parts;    /* of nxt_str_t * */
+    nxt_array_t         *values;   /* of nxt_var_value_t */
+    nxt_array_t         *parts;    /* of nxt_str_t * */
 
     nxt_lvlhsh_t        cache;
 
@@ -433,11 +433,18 @@ nxt_var_query_init(nxt_var_query_t **query_p, void *ctx, nxt_mp_t *mp)
             return NXT_ERROR;
         }
 
-        nxt_array_init(&query->values, mp, sizeof(nxt_var_value_t));
-        nxt_array_init(&query->parts, mp, sizeof(nxt_str_t *));
+        query->values = nxt_array_create(mp, 1, sizeof(nxt_var_value_t));
+        if (nxt_slow_path(query->values == NULL)) {
+            return NXT_ERROR;
+        }
+
+        query->parts = nxt_array_create(mp, 1, sizeof(nxt_str_t *));
+        if (nxt_slow_path(query->parts == NULL)) {
+            return NXT_ERROR;
+        }
 
     } else {
-        nxt_array_reset(&query->values);
+        nxt_array_reset(query->values);
     }
 
     query->ctx = ctx;
@@ -469,7 +476,7 @@ nxt_var_query(nxt_task_t *task, nxt_var_query_t *query, nxt_var_t *var,
         return;
     }
 
-    mp = query->values.mem_pool;
+    mp = query->values->mem_pool;
     subs = nxt_var_subs(var);
     value = query->spare;
 
@@ -509,7 +516,7 @@ nxt_var_query(nxt_task_t *task, nxt_var_query_t *query, nxt_var_t *var,
 
     query->spare = value;
 
-    val = nxt_array_add(&query->values);
+    val = nxt_array_add(query->values);
     if (nxt_slow_path(val == NULL)) {
         goto fail;
     }
@@ -566,9 +573,9 @@ nxt_var_query_finish(nxt_task_t *task, nxt_var_query_t *query)
         goto done;
     }
 
-    val = query->values.elts;
+    val = query->values->elts;
 
-    for (i = 0; i < query->values.nelts; i++) {
+    for (i = 0; i < query->values->nelts; i++) {
         var = val[i].var;
 
         subs = nxt_var_subs(var);
@@ -579,7 +586,7 @@ nxt_var_query_finish(nxt_task_t *task, nxt_var_query_t *query)
 
             nxt_assert(str != NULL);
 
-            part = nxt_array_add(&query->parts);
+            part = nxt_array_add(query->parts);
 
             if (nxt_slow_path(part == NULL)) {
                 query->failed = 1;
@@ -591,7 +598,7 @@ nxt_var_query_finish(nxt_task_t *task, nxt_var_query_t *query)
             length += str->length - subs[j].length;
         }
 
-        p = nxt_mp_nget(query->values.mem_pool, length + var->strz);
+        p = nxt_mp_nget(query->values->mem_pool, length + var->strz);
         if (nxt_slow_path(p == NULL)) {
             query->failed = 1;
             goto done;
@@ -600,7 +607,7 @@ nxt_var_query_finish(nxt_task_t *task, nxt_var_query_t *query)
         val[i].value->length = length;
         val[i].value->start = p;
 
-        part = query->parts.elts;
+        part = query->parts->elts;
         src = nxt_var_raw_start(var);
 
         last = 0;
@@ -625,7 +632,7 @@ nxt_var_query_finish(nxt_task_t *task, nxt_var_query_t *query)
             *p = '\0';
         }
 
-        nxt_array_reset(&query->parts);
+        nxt_array_reset(query->parts);
 
         nxt_debug(task, "var: \"%*s\" -> \"%V\"", var->length, src,
                   val[i].value);
--- a/src/nxt_cert.c
+++ b/src/nxt_cert.c
@@ -894,7 +894,7 @@ nxt_cert_store_load(nxt_task_t *task, nxt_mp_t *mp)
 
 
         if (nxt_slow_path(ret != NXT_OK)) {
-            nxt_array_remove_last(certs);
+            nxt_array_del_last(certs);
             continue;
         }
 
--- a/src/nxt_isolation.c
+++ b/src/nxt_isolation.c
@@ -565,7 +565,7 @@ nxt_isolation_set_lang_mounts(nxt_task_t *task, nxt_process_t *process,
     mp = process->mem_pool;
 
     /* copy to init mem pool */
-    mounts = nxt_array_copy(mp, NULL, lang_mounts);
+    mounts = nxt_array_copy(mp, lang_mounts);
     if (mounts == NULL) {
         return NXT_ERROR;
     }
--- a/src/nxt_runtime.c
+++ b/src/nxt_runtime.c
@@ -734,7 +734,7 @@ nxt_runtime_thread_pool_exit(nxt_task_t *task, void *obj, void *data)
     for (i = 0; i < n; i++) {
 
         if (tp == thread_pools[i]) {
-            nxt_array_remove(rt->thread_pools, &thread_pools[i]);
+            nxt_array_del(rt->thread_pools, &thread_pools[i]);
 
             nxt_free(tp);
 


src/nxt_array.h
```diff
@@ -9,34 +10,32 @@
...
-    nxt_mp_t  *mem_pool;
+    void                *elts;
+    nxt_uint_t          nelts;
+    nxt_uint_t          size;
+    nxt_uint_t          nalloc;
+    nxt_mp_t            *mem_pool;
 } nxt_array_t;
 
 
-nxt_inline void
```
echolimazulu, line 21: removed incorrect array initialization (nxt_array_init) with a pointer to a non-existent set of elements;

src/nxt_array.h
```diff
@@ -9,34 +10,32 @@
...
-nxt_array_init(nxt_array_t *array, nxt_mp_t *mp, size_t size)
-{
-    array->elts = nxt_pointer_to(array, sizeof(nxt_array_t));
```
echolimazulu, line 24: The pointer to "elts" was changing at the time of the memory reallocation. This offset would point to a non-existent pointer.

src/nxt_array.c
```diff
@@ -1,150 +1,253 @@
...
 nxt_array_t *
-nxt_array_create(nxt_mp_t *mp, nxt_uint_t n, size_t size)
+nxt_array_create(nxt_mp_t *mp, nxt_uint_t arr_size, size_t elt_size)
 {
-    nxt_array_t  *array;
+    nxt_uint_t     nalloc;
+    nxt_array_t    *array;
+    void           *elts;
 
-    array = nxt_mp_alloc(mp, sizeof(nxt_array_t) + n * size);
```
echolimazulu, line 15: The allocated memory is actually used until the first reallocation of memory. During the subsequent work of the program, the specified memory area was not used, remaining abandoned.

src/nxt_array.c
```diff
@@ -1,150 +1,253 @@
...
+    array = nxt_mp_alloc(mp, sizeof(nxt_array_t));
 
     if (nxt_slow_path(array == NULL)) {
         return NULL;
     }
 
-    array->elts = nxt_pointer_to(array, sizeof(nxt_array_t));
+    nalloc = nxt_array_calc_alloc(arr_size);
+    elts = nxt_mp_alloc(mp, elt_size * nalloc);
```
echolimazulu, line 34: - fixed initialization of a set of elements (elts), pointer, memory size;

src/nxt_array.c
```diff
@@ -1,150 +1,253 @@
...
 
-        if (nxt_slow_path(p == NULL)) {
-            return NULL;
         }
 
-        nxt_memcpy(p, array->elts, array->size * nalloc);
+        elt = nxt_pointer_to(array->elts, array->size * array->nelts);
 
-        if (array->elts != nxt_pointer_to(array, sizeof(nxt_array_t))) {
-            nxt_mp_free(array->mem_pool, array->elts);
```
echolimazulu, line 70: - eliminated the possibility of early deletion of an existing pointer to a set of elements (elts) at the time of memory reallocation;

`array->elts` is freed before the new pointer is set.

src/nxt_array.c
```diff
@@ -1,150 +1,253 @@
...
-    nalloc = array->nalloc;
+    if (nxt_fast_path(array != NULL)) {
 
-    if (array->nelts == nalloc) {
+        nxt_uint_t    nalloc;
+        void          *elt, *elts;
 
-        if (nalloc < 16) {
-            /* Allocate new array twice larger than current. */
-            new_alloc = (nalloc == 0) ? 4 : nalloc * 2;
```
echolimazulu, line 54: - fixed the formula for calculating the required amount of memory (nalloc) allocated for the created array;

4 is too big for an empty and possibly never used array in the future.

src/nxt_array.c
```diff
@@ -1,150 +1,253 @@
...
-    uint32_t  nalloc, new_alloc;
-
-    nalloc = array->nalloc;
+    if (nxt_fast_path(array != NULL)) {
 
-    if (array->nelts == nalloc) {
+        nxt_uint_t    nalloc;
+        void          *elt, *elts;
 
-        if (nalloc < 16) {
```
echolimazulu, line 52: - fixed the formula for calculating the required amount of memory (nalloc) allocated for the created array;

it is better to use an even value (2, 4, 8, 16, 32, etc) and also a multiple of 8.

src/nxt_array.h
```diff
@@ -9,34 +10,32 @@
...
-    nxt_array_t *src);
+NXT_EXPORT void nxt_array_del(nxt_array_t *array, void *elt);
+NXT_EXPORT void nxt_array_del_last(nxt_array_t *array);
+NXT_EXPORT nxt_array_t *nxt_array_copy(nxt_mp_t *mp,
+                                       nxt_array_t *array);
+
 
 #define                                                                       \
-nxt_array_last(array)                                                         \
+nxt_array_pointer_to_index(array, index)                                      \
```
echolimazulu, line 33: - added a function to get a pointer by the index of an element in an array.

tippexs:
Thanks for submitting the PR. Please let us have a look on it. We will come back with feedback asap.

echolimazulu:
Hello @tippexs! 

Thank you for your attention to my suggested fixes and improvements. 
Of course, I will be glad to receive feedback from you or your colleagues.

src/nxt_array.c
```diff
@@ -1,150 +1,253 @@
 
 /*
+ * Copyright (C) Evgenii Sokolov
  * Copyright (C) Igor Sysoev
  * Copyright (C) NGINX, Inc.
  */
 
 #include <nxt_main.h>
```
hongzhidao, line 9: It seems you broke the design, I don't understand why you thought there is a problem with the current way.

Take a look at the change.

```
[](https://github.com/nginx/unit/pull/664/files#diff-b340fa7a4c4fbec4cfadd48c7469e88ccb7a183b31423c74a0d5a23047d11e85)[](https://github.com/nginx/unit/pull/664/files#diff-b340fa7a4c4fbec4cfadd48c7469e88ccb7a183b31423c74a0d5a23047d11e85)struct nxt_var_query_s {
    nxt_array_t         values;   /* of nxt_var_value_t */
    nxt_array_t         parts;    /* of nxt_str_t * */
    nxt_array_t         *values;   /* of nxt_var_value_t */
    nxt_array_t         *parts;    /* of nxt_str_t * */

    nxt_lvlhsh_t        cache;

@@ -433,11 +433,18 @@ nxt_var_query_init(nxt_var_query_t **query_p, void *ctx, nxt_mp_t *mp)
            return NXT_ERROR;
        }

        nxt_array_init(&query->values, mp, sizeof(nxt_var_value_t));
        nxt_array_init(&query->parts, mp, sizeof(nxt_str_t *));
        query->values = nxt_array_create(mp, 1, sizeof(nxt_var_value_t));
        if (nxt_slow_path(query->values == NULL)) {
            return NXT_ERROR;
        }

        query->parts = nxt_array_create(mp, 1, sizeof(nxt_str_t *));
        if (nxt_slow_path(query->parts == NULL)) {
            return NXT_ERROR;
        }
```

That's a few more allocations than before, and that's why `nxt_array_init()` is required.

echolimazulu: Hello @hongzhidao, 
Thank you for your attention to my suggested fixes and improvements!

I want to draw your attention to the fact that `nxt_array_init` has been completely removed and replaced in favor of `nxt_array_create`.

I thank you for bringing my additional attention to the concepts used in **nxt_var.c**. In this regard, I paid more attention to the proposed changes and related objects.

I came to the conclusion that in the next commit to the PR I will propose changes taking into account the need to use `nxt_array_init` and not changing **nxt_var.c** due to the peculiarities of working with the `nxt_var_query_s` structure, since some of its properties are closely related to caching for which I would not want influence in any way.

I'll explain why I'm proposing these changes in my next answer. 
Thank you for your attention!

src/nxt_array.c
VBart: You're trying to optimize a case that is rare in critical parts, while pessimizing the most common case.

src/nxt_array.c
VBart: There's nothing wrong in freeing `array->elts` pointer before setting another one. No parallel access is expected to arrays anyway.

src/nxt_array.h
```diff
@@ -9,34 +10,32 @@
 
 
 typedef struct {
-    void      *elts;
-    /* nelts has uint32_t type because it is used most often. */
-    uint32_t  nelts;
-    uint16_t  size;
-    uint16_t  nalloc;
-    nxt_mp_t  *mem_pool;
```
VBart, line 17: With this change, you made the array structure bigger without any reason.

src/nxt_array.h
VBart: It's non existing in the particular `nxt_var_query_init()` case, but it may exist in other cases that this function in designed for (but they aren't yet represented in the code).

There's nothing incorrect in setting pointer to the end of array structure as soon as the correct number of initialized elements is provided (zero in that case).

The only issues with this function is that it's missing the "nalloc" parameter to be used in non-zero cases.

src/nxt_array.h
VBart: It wouldn't, as by the time of array initialization there cannot be any reallocations. For already existing arrays the "nxt_array_reset()" function is supposed to be used.

src/nxt_var.c
```diff
@@ -433,11 +433,18 @@ nxt_var_query_init(nxt_var_query_t **query_p, void *ctx, nxt_mp_t *mp)
...
-        nxt_array_init(&query->parts, mp, sizeof(nxt_str_t *));
+        query->values = nxt_array_create(mp, 1, sizeof(nxt_var_value_t));
+        if (nxt_slow_path(query->values == NULL)) {
+            return NXT_ERROR;
+        }
+
+        query->parts = nxt_array_create(mp, 1, sizeof(nxt_str_t *));
+        if (nxt_slow_path(query->parts == NULL)) {
+            return NXT_ERROR;
+        }
```
VBart, line 444: This results in 4 additional allocations in the most common case.

src/nxt_array.h
```diff
@@ -9,34 +10,32 @@
...
-nxt_array_init(nxt_array_t *array, nxt_mp_t *mp, size_t size)
-{
-    array->elts = nxt_pointer_to(array, sizeof(nxt_array_t));
-    array->size = size;
-    array->mem_pool = mp;
-}
-
 NXT_EXPORT nxt_array_t *nxt_array_create(nxt_mp_t *mp, nxt_uint_t n,
-    size_t size);
+                                         size_t size);
```
VBart, line 22: Here you add a style issue. The indentation for additional parameter lines in prototype is 4 spaces.

VBart (commented):
It looks like a refactoring in the sake of refactoring without solving any real problem. As a result, some cases will be a bit better, but some common cases will work worse.

src/nxt_array.c
```diff
@@ -1,150 +1,253 @@
 
 /*
+ * Copyright (C) Evgenii Sokolov
```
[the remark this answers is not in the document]
echolimazulu: Clear. I didn't know this. Since many files say otherwise. Please search: "Copyright (C) Valentin V. Bartenev"

src/nxt_array.c
echolimazulu: What is this? One more?

**Please provide 2 common cases:**
one for `x < 16 = x * 2`
one for `x > 16 = x + x / 2`

Otherwise, it is not clear to me what specific cases you are talking about and how I can get acquainted with them, check and make sure that the information you provided is objective, and not fictitious. 

How is the result `16 * 2 = 32` worse than `15 * 2 = 30`?

How then does your overgrowth logic fit into the initial initialization of an empty array in size `(nalloc == 0 ) ? 4`?

src/nxt_array.c
```diff
@@ -1,150 +1,253 @@
...
 
-    array = nxt_mp_alloc(mp, sizeof(nxt_array_t) + n * size);
+    array = nxt_mp_alloc(mp, sizeof(nxt_array_t));
 
     if (nxt_slow_path(array == NULL)) {
         return NULL;
     }
 
-    array->elts = nxt_pointer_to(array, sizeof(nxt_array_t));
+    nalloc = nxt_array_calc_alloc(arr_size);
```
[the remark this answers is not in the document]
echolimazulu: I agree, it's logical.

src/nxt_array.c
```diff
@@ -1,150 +1,253 @@
...
+static void *nxt_array_elts_realloc(nxt_array_t *array,
+                                    nxt_uint_t nalloc);
+
+
+/*
+ * nxt_array_create - Create a new array and return a pointer to it.
+ */
 nxt_array_t *
-nxt_array_create(nxt_mp_t *mp, nxt_uint_t n, size_t size)
+nxt_array_create(nxt_mp_t *mp, nxt_uint_t arr_size, size_t elt_size)
```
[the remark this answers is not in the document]
echolimazulu: Noted in my notes. But somehow this does not really correlate with the logic you described regarding trivial comments and the need to read the contents of the code of the functions used. However, I accept and respect your rules for naming functions and their parameters.

src/nxt_array.h
echolimazulu: I marked it in my notes, I will be careful with the indents, since in my case they are set automatically.

src/nxt_array.c
echolimazulu: Logically, I will think about how to improve this in the context of the changes I propose.

But I want to note that in the changes I propose, the allocated and unused memory at the time of creating the array is released in the event of its subsequent expansion, and in the context of existing code, this memory area is not released at all and is not used after it is allocated from the moment the array is created, regardless of the need to change the value.

src/nxt_array.c
echolimazulu: The fact that parallel access to arrays is not supposed to be said somewhere? Maybe in other not trivial and not silly comments?

**Question to you from the opposite:**
What's wrong with assigning a new pointer before deleting the old pointer?

src/nxt_array.c
VBart: It makes only sense to release big amounts of memory, at least more than a page size. All small allocations will be released when the memory pool be destroyed. Those small allocations don't make any visible difference to the memory consumption during request processing, but any additional calls can introduce additional delays.

src/nxt_array.c
VBart: Have you seen any locking in the array functions? For example, nxt_array_add() is clearly not atomic with or without your patch. Actually, it's ok to assume that there's no parallel access supposed to any of the Unit's changeable data structures unless they have some locking around. The memory pools and every code, that uses memory pools isn't atomic too. We try to avoid parallel access as much as possible, keeping such structures and places at minimum among the project.

src/nxt_array.h
echolimazulu: > It's non existing in the particular `nxt_var_query_init()` case, but it may exist in other cases that this function in designed for (but they aren't yet represented in the code).

I already made the decision to revert `nxt_array_init` and discard changes to **nxt_var.c** back in @hongzhidao answer. I will also take into account additional information about possible upcoming changes that you provided. Thanks.

> 
> There's nothing incorrect in setting pointer to the end of array structure as soon as the correct number of initialized elements is provided (zero in that case).

So you don't see any problem with using `nxt_array_init` with `array->elts = nxt_pointer_to(array, sizeof(nxt_array_t));` in the context of `nxt_var_query_s`:
- `values` points to `parts`
- `parts` points to `cache`
at initialization time? 

Instead of being `NULL` for example until `array->elts` is fully initialized with `nxt_arrray_add` for example?

> 
> The only issues with this function is that it's missing the "nalloc" parameter to be used in non-zero cases.

I took it into account, even at the moment when @hongzhidao answered. But thanks for pointing it out to me.

src/nxt_array.h
echolimazulu: I meant memory reallocation in case of array expansion.

src/nxt_array.c
VBart: In most cases the initial array size is set to fit all the values. But this value can be not very precise and sometimes the array can grow a bit. The idea is to find a balance between too much memory wasted and too many reallocations. Current logic avoids multiple reallocations for small values by using more steep growing formula for low number of elements and at the same time it avoids too much overallocation for adding a few more elements over already relatively high number of elements.

4 and 16 are constants which obtained empirically by Igor Sysoev based on his 20+ years of experience in nginx development.

src/nxt_array.c
VBart: > What's wrong with assigning a new pointer before deleting the old pointer?

Sure there's nothing wrong. Except it just adds a bit more code, a bit more logic that needs to be reviewed and will be read by everybody. Much worse if seeing such handling of previous pointer they will make some assumption about the reasons why it was made like that.

Any variant of code makes sense and would look fine in this particular case, but there should be a reason for changing already existing code. Current code is dumb and straightforward right as is as it's required in this particular place.

echolimazulu: I mean, is there anything to warn other developers against using array-based memory manipulation in parallel, other than internal rules known and available only to your team?

As you aptly noted, the operation of adding an element to the array was not atomic either before or after the changes I proposed, as was the parallel use, so using this argument as an excuse for not adjusting, albeit insignificant, but nevertheless not useless, looks extremely strange .

**You still haven't answered the question:**
>
> **Question to you from the opposite:**
> What's wrong with assigning a new pointer before deleting the old pointer?

src/nxt_array.h
VBart: > Instead of being NULL for example until array->elts is fully initialized with nxt_array_add for example?

The "array->elts" serves for two different purposes:
1. It indicates whenever elements were allocated from the same memory as the array structure. In this case this memory can't be freed.
2. It points to the allocated elements, if any.

By setting it to NULL in the current code, you will break the first purpose, so the check in nxt_array_add():
```
        if (array->elts != nxt_pointer_to(array, sizeof(nxt_array_t))) {
            nxt_mp_free(array->mem_pool, array->elts);
        }
```
won't work right.

echolimazulu: Perhaps I expressed myself incorrectly. I didn't suggest using NULL with the old code. I think it's obvious to everyone where this will lead.

Sorry, but it seems to me that you somehow selectively answer the questions I asked. Pulling out of the context of my question only what is necessary for you in order to make it meaningless or incorrect.

In the future, I will try to explain the essence of my ideas (for example) in more detail, so that you do not have the opportunity to double-read about the questions I asked and the opportunity, as it seems to me, to present them in the way you need.

**I remind you that you still haven't answered the question:**
>
> So you don't see any problem with using `nxt_array_init` with `array->elts = nxt_pointer_to(array, sizeof(nxt_array_t));` in the context of `nxt_var_query_s`:

> `values` points to `parts`
> `parts` points to `cache`
> at initialization time?

src/nxt_array.c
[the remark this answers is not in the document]
VBart: > Please search: "Copyright (C) Valentin V. Bartenev"

Most of these copyrights were added for the newly created files even before the first public beta of Unit 0.1.
But for example, Tiago's huge contributions were already without additional "Copyright" lines, except copy'n'pasting existing: https://github.com/nginx/unit/commit/c554941b4f826d83d92d5ca8d7713bea4167896e

The issue here is that we should either add everyone in every file which is touched even with some typo fix and can end up with endless list of copyrights in each source, or have some complex rules when and why to add a copyright, which may disappoint someone.

I've already several times got prepared a patch to remove all the individual copyrights from files, but each time was unlucky as such commit could not look good by the community due to some events (like this one: https://www.nginx.com/blog/do-svidaniya-igor-thank-you-for-nginx/ ). So it was postponed in order to not cause some negative correlations.

src/nxt_array.c
VBart: > I mean, is there anything to warn other developers against using array-based memory manipulation in parallel, other than internal rules known and available only to your team?

I'd rather warn any developers of using any function without reading and understanding the function's code. And it's just a common sense especially when dealing with some internal APIs of the project.

Even public APIs of some very popular projects are very far from being well-understood without reading the internals. I can't count how many times I had to read the Linux kernel's code behind some syscall to understand nuances of its functioning that not covered anywhere. And for sure, anyone will fail, trying to use PHP's interpreter API or OpenSSL only by manuals and without reading internals.

So, the answer to your question is experience. It takes some time to get familiar with the project's architecture and the right way how to code it. And a change to how the particular pointer is manipulated won't make this code thread safe.

> What's wrong with assigning a new pointer before deleting the old pointer?

I don't know if while writing it you've seen my [second message](https://github.com/nginx/unit/pull/664/commits/8d1df4121295f929c7b2b5ea974f223ca25e50d5#r836395146) to the thread or not, but anyway, in short:

1. Making non thread safe code to look in some parts like it can be thread safe - only misleads.
2. Nothing wrong, except it wasn't done this way initially for the sake of simplicity and it's not required.

src/nxt_array.h
VBart: > So you don't see any problem with using nxt_array_init with array->elts = nxt_pointer_to(array, sizeof(nxt_array_t)); in the context of nxt_var_query_s:

As long as this pointer works only like a flag (and it works like a flag, which I've tried to explain by naming two purposes of the pointer) I see no problem here.

> values points to parts
> parts points to cache
> at initialization time?

"values" doesn't point to "parts".  It points to the end of the array structure in "nxt_var_query_s", where zero number of elements should start.  Sure, due to zero number of elements this value is equal to position of the "parts" field.  Which is just a coincidence in this particular case.  But, for example, if there would be not zero, but N elements before the "parts" field, how is this different?

If it overflows, it will corrupt the "parts" and possibly other fields. If it will be placed somewhere else, then it may corrupt some other structures - no difference here too.

In your version of code you've allocated a space from memory pool for 1 element.

So I don't see how overflowing 0 elements is different from overflowing 1 element (2, or 3, or N... doesn't matter). And I don't see how the consequences of overwriting the "nxt_var_query_s" structure is more dangerous to any consequences of overflowing any other structure that can be placed in the memory pool right after the 1 allocated element in your code.

src/nxt_array.c
echolimazulu: > In most cases the initial array size is set to fit all the values. But this value can be not very precise and sometimes the array can grow a bit. The idea is to find a balance between too much memory wasted and too many reallocations. Current logic avoids multiple reallocations for small values by using more steep growing formula for low number of elements and at the same time it avoids too much overallocation for adding a few more elements over already relatively high number of elements.
> 
> 4 and 16 are constants which obtained empirically by Igor Sysoev based on his 20+ years of experience in nginx development.

I accept the empirical method of **Igor Sysoev**, since I have great respect for the work he has done, so let's leave it as it is. 

The actual number of uses of the value 4 (nalloc) with the initial 0 is unknown to me at the moment. Perhaps in the future you should evaluate the real optimal value for these parameters, since most scenarios assume a known value in advance, which in most cases is greater than 0. 

In this regard, I think this will not be any problem for you if this issue really affects performance, then there must be a real basis and based on performance testing of this logic so that the values ​​bare justified and you do not have to refer first to existing use cases and then to an empirical method.

**I'm starting from the logic:**
Don't allocate too much memory unless you really need to. When is the next call to reallocate memory for the array being expanded? - unknown, possibly never. In that case, which is more important? Allocated and never used memory, or one extra call to memory when you need to expand the array?

src/nxt_array.h
VBart: "point to a non-existent pointer" makes no sense to me. Pointer is just an arbitrary integer value that is big enough to keep any possible memory address and can be used with pointer arithmetic. Depending on the purpose, it can hold a memory address that is valid for accessing (dereferencing), or not valid for accessing. Pointers to non-valid memory addresses are often used for pointer arithmetic or like some indicators.

E.g. there are can be pointers indicating the buffer end. In your terminology the "end" field of "nxt_buf_mem_t" structure always points "to a non-existent pointer" as it keeps the address of high border of the allocated buffer memory.

Here it points to the memory where elements supposed to start. If there's zero elements, then dereferencing the first element [0] of zero elements array is a mistake. And this mistake is not different from dereferencing of the second element [1] of 1 elements array, or dereferencing of the third element [2] of 2 elements array, and etc...

src/nxt_array.c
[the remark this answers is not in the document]
echolimazulu: All logically, I fully support your choice.

src/nxt_array.c
VBart: 4 was just a good start for the *2 rule... assuming that most elements are just 2-4 pointers long and nxt_mp_alloc() won't work good for very small allocations anyway.

> which is more important? Allocated and never used memory, or one extra call to memory when you need to expand the array?

Sure it depends on how big is the extra memory that is allocated and never used. Any extra call that can possibly lead to a syscall is very bad. Too much allocation isn't good either.

VBart: Note also, that nxt_mp_alloc() doesn't always allocate the size that you've asked for. Sometimes it over-allocates too.

src/nxt_array.h
echolimazulu: I'm sorry, I made a mistake with the names:
`values.elts` (offset size - `nxt_array_t`) - the actual pointer goes to `parts`;
`parts.elts` (offset size - `nxt_array_t`) - actual pointer goes to `cache`;
It's about the old `nxt_array_init` code.

Regarding the changes affecting **nxt_var.c**, I refused as soon as @hongzhidao drew my attention to the design, as I understand that they gave an insufficient assessment of the correctness of the work of all components.

The difference is that the initialization of the `array->elts` object that I propose and the allocation of memory for it are initially independent of the main array structure and the memory allocated for its storage. Which is allocated and released as the array grows without leaving traces in the form of abandoned memory areas in favor of 1 additional call when creating the array. The array structure in itself contains only a pointer to the set of elements stored in it, which is replaced without any problems by another pointer, while guaranteed to free the previously used memory area.

The question here is not only about possible overflow and overwriting, since it can be achieved in other simpler ways if desired. And that the pointer is initially set to a place in memory that has nothing to do with the current set of elements. In my understanding, if it does not exist, then it should be NULL, and not point to another object. I remember you said that the worst thing is if the developers do not understand why the code was written and for what end goal it was done. Here is the same case already contained in the current version of the code, which is sincerely incomprehensible to me and actually misleads me. This piece of code looks like it was copied from nxt_array_create and the programmer simply forgot about determining the dimension of the array, considering that this pointer refers to an existing list of elements.

src/nxt_array.h
echolimazulu: I understand your logic of following the cursor outside the array. But it actually works until the first reallocation of memory (in the old code), since at that moment the memory is re-allocated in a separate block and a new pointer to a new memory area not related to the current array is assigned to array->elts and actually stops reflecting the real and actual boundary the end of the array.

src/nxt_array.h
echolimazulu: I'm starting from two comments:
```
/*
 * nxt_int_t corresponds to the most efficient integer type,
 * an architecture word. It is usually the long type
 * but on Win64 the long is int32_t, so pointer size suits better.
 * nxt_int_t must be no less than int32_t.
 */
```
```
/* nelts has uint32_t type because it is used most often. */
```
From this I conclude that `int32_t`, `uint32_t` are the most productive in the opinion of **Igor Sysoev** and my personal opinion. If we talk about the reality of using `uint16_t`, then I would like to see a place in the Unit program in the form of an array expanding to 65535 elements (`nalloc`). Therefore, I do not see a significant difference, except for the architectural one, which works faster with 32-bit values. Moreover, this type of structure is allocated once for each individual array. I think that the issue of frequency of calls is of course relevant in this case. I will review this structure to revert to the previous uint size for `nalloc` and `size`.

src/nxt_array.h
VBart: "stops reflecting the real and actual boundary the end of the array" and this indicates that memory isn't allocated with the array structure, but with additional nxt_mp_alloc() call and therefore can be freed in case of additional reallocation needed.

The nxt_array_init() function call reflects the same state as nxt_array_create() call with n set to 0.

src/nxt_array.h
VBart: > The difference is that the initialization of the array->elts object that I propose and the allocation of memory for it are initially independent of the main array structure and the memory allocated for its storage.

The initial design was made this way because in many cases requests are processed without any variables interpolation using only const string values. So, the `nxt_var_query_init()` call allocates the query structure, then `nxt_var_query()` calls actually do nothing as they return on checking `nxt_var_is_const(var)`, so we have just one allocation in this common case. If there will be variable strings, then we will have 3 allocations in total (query structure and two array elements).

After your change, there will be always 3 allocations.

> This piece of code looks like it was copied from nxt_array_create and the programmer simply forgot about determining the dimension of the array, thinking that this pointer refers to an existing list of elements.

You can improve it this way to make it look more consistent:
```c
nxt_inline void
nxt_array_init(nxt_array_t *array, nxt_mp_t *mp, nxt_uint_t n, size_t size)
{
    array->elts = nxt_pointer_to(array, sizeof(nxt_array_t));
    array->nelts = 0;
    array->size = size;
    array->nalloc = n;
    array->mem_pool = mp;
}


nxt_array_t *
nxt_array_create(nxt_mp_t *mp, nxt_uint_t n, size_t size)
{
    nxt_array_t  *array;

    array = nxt_mp_alloc(mp, sizeof(nxt_array_t) + n * size);

    if (nxt_slow_path(array == NULL)) {
        return NULL;
    }

    nxt_array_init(array, mp, n, size);

    return array;
}
```
with the relevant changes to `nxt_var.c`:
```diff
diff --git a/src/nxt_var.c b/src/nxt_var.c
--- a/src/nxt_var.c
+++ b/src/nxt_var.c
@@ -433,8 +433,8 @@ nxt_var_query_init(nxt_var_query_t **que
             return NXT_ERROR;
         }
 
-        nxt_array_init(&query->values, mp, sizeof(nxt_var_value_t));
-        nxt_array_init(&query->parts, mp, sizeof(nxt_str_t *));
+        nxt_array_init(&query->values, mp, 0, sizeof(nxt_var_value_t));
+        nxt_array_init(&query->parts, mp, 0, sizeof(nxt_str_t *));
 
     } else {
         nxt_array_reset(&query->values);
```

VBart: It refers to the "existing" zero number of elements. It's also possible to initialize with more than zero number of elements, but `nxt_var_query_s` will be bigger:
```c
struct nxt_var_query_s {
    nxt_array_t         values;   /* of nxt_var_value_t */
    nxt_var_value_t     values_elts[4];
    nxt_array_t         parts;    /* of nxt_str_t * */
    nxt_str_t           parts_elts[4];

    nxt_lvlhsh_t        cache;

    nxt_str_t           *spare;
    nxt_uint_t          waiting;
    nxt_uint_t          failed;   /* 1 bit */

    void                *ctx;
    void                *data;

    nxt_work_handler_t  ready;
    nxt_work_handler_t  error;
};
```

src/nxt_array.c
```diff
@@ -1,150 +1,253 @@
...
 
+static void nxt_array_elts_copy(void **dst_elts, nxt_array_t *src_array);
+static nxt_uint_t nxt_array_calc_alloc(nxt_uint_t nelts);
+static void *nxt_array_elts_realloc(nxt_array_t *array,
+                                    nxt_uint_t nalloc);
+
+
+/*
+ * nxt_array_create - Create a new array and return a pointer to it.
+ */
```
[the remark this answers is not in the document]
echolimazulu: > We avoid such silly comments. If something isn't clear from the function's prototype, then a bad function name has chosen. We prefer to leave comments only in non-trivial parts of the code. Developers should read code, not comments. The code is executed, the comments aren't. People make mistakes, add bugs, so no one can be 100% sure that the description of a function clearly represents how the function works. It's better if someone read the function code again before using it, rather than reading the comment.

Thank you!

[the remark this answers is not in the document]
VBart: > Thank you!

Hope that I didn't accidentally offended you by the word "silly" regarding that comment.  I've actually meant an "obvious" one. Because to me it looks like:
 - This is a function to create arrays
 - Yeap, I see it, it's called array create...
 - And it returns a pointer to it!
 - No surprise here too! Its return type is `nxt_array_t *`. I can read C.

But in fact, sometimes it can also return `NULL`.

src/nxt_array.h
echolimazulu: I understand why `nxt_array_init` appeared as soon as @hongzhidao pointed out to me the design used in **nxt_var.c**, so I immediately abandoned almost any changes associated with this file, except for the explicit initialization of `nalloc` and `nelts` in `nxt_array_init` and a pointer at a given offset actually setting a pointer to an adjacent object that you define as necessary before the first expansion of the array, whose function is to enforce the bounds of the array. I understand why `nxt_array_init` should be lightweight in the context of a query and why it was originally marked up as inline, so I have no questions about its necessity.

src/nxt_array.h
echolimazulu: As you can see, you, like the programmer (about whom I wrote), also fell into this trap earlier (I speak with care). Especially in the context of how you advocate the importance of holding array bounds until the first reallocation of memory.

**Explanation:**
You say that `nxt_array_create` and `nxt_array_init` are essentially the same when `n = 0`, and they are. But for some reason you forget about what is expected after the execution of these functions, namely: `nxt_array_add`, which immediately resets the `array->elts` array end pointer to a new one. What then is the initial meaning of pointing to a neighboring object, if immediately after it loses this boundary? Taking into account what you wrote earlier about the need to specify the array boundary in case there are elements in the array, if boundary control may be necessary, then only in the case of `nxt_array_create` with a value of `n > 1` (not `n == 0`), since in any other case this the boundary is useless and ineffective and is essentially an over-indicator of anything that you are overly focused on and give unjustified importance to, like the programmer before you, which in my humble opinion is wrong and it confirms a number of irrefutable deductive facts.

echolimazulu:
> It looks like a refactoring in the sake of refactoring without solving any real problem. As a result, some cases will be a bit better, but some common cases will work worse.

You are free to name this PR whatever you like.

**I can give you a few more options:**
- change for the sake of change;
- code review for the sake of code review;
- improvements for the sake of improvements;
- fixes for the sake of fixes.

There is no such thing as a lot of well-thought-out and optimized code. Improvements and corrections are exactly the same continuous processes as the development of a product by adding new functionality to it.

The fact that the improvements and fixes I've suggested don't align with your main product release plan - yes, I can afford it, as I'm proposing those improvements and fixes that I personally need first, but often affect most other users of your product as well. As it was before with issues: 
**listen sockets (unix)** https://github.com/nginx/unit/issues/643 - issue not resolved;
**source mask (client_ip - X-Forward-For for unix)** https://github.com/nginx/unit/issues/642 - solution found, but still not included in the main branch;
**Python module with ASGI "error 503"** https://github.com/nginx/unit/issues/635 - issue not resolved.

There are too many unresolved issues and problems that I found in the Unit product immediately after the first launch of the application, don't you think? Maybe the product really lacks refactoring for the sake of refactoring or code review for the sake of code review?

Yes, the specifically proposed improvement and fixes are almost not directly related to solving these problems (with the exception of listening sockets), but I raise this discussion not for the sake of discussing and wasting my and your precious time, but in order to initially determine whether this is a violation of the underlying logic significant and whether it is necessary to eliminate it in the future, if so then what is the best way to do this.

src/nxt_array.c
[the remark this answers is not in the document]
echolimazulu: > > Thank you!
> 
> Hope that I haven't accidentally offended you by the word "silly" regarding that comment. I've actually meant an "obvious" one. Because to me it looks like:
> 
> * This is a function to create arrays
> * Yeap, I see it, it's called array create...
> * And it returns a pointer to it!
> * No surprise here too! Its return type is `nxt_array_t *`. I can read C.
> 
> But in fact, sometimes it can also return `NULL`.

It's not my problem.

**For several reasons:**
- this is your own vision, perhaps a generally accepted principle within your team;
- I do not give such definitions to anyone's work;
- I am not an official representative and face of a public company;
- my real experience of programming in C is about 1 month and this is my first project in C in principle.

I am strictly aware of my current knowledge, skills and experience in C programming and I am aware that I can be wrong in many ways, in this regard, I am trying to get feedback from "heavyweights" in programming with a lot of experience behind them like you and yours colleagues. My principle is this: "the recognition of one's own mistakes and wrongs is the basis of personal growth."

The only thing that worries me a little is the fate of this project, at a time when such a respected and experienced specialist (without sarcasm) in C programming, accepts the current codebase as a dogma and cannot explain to me (for noob) the ultimate meaning and benefits of the chosen algorithms of memory allocation and pointer assignment by his colleagues, as well as other features of the application, based on the principles of the generally accepted technical justification laid down in all the foundations of C programming.

src/nxt_array.h
VBart: > I want to remind you that in the current version of the code (the old version of the code), you simply leave the initial allocated memory area abandoned, which is not normally freed for later reallocation of memory.

It will be freed when the request processing be finished. Keeping this structure even without using their fields simplifies the code, allowing to have no difference between cases when it was used and when it wasn't.

> What then is the initial meaning of pointing to a neighboring object, if immediately after it loses this boundary?

The initial meaning is to indicate that the `array->elts` pointer is a part of the array structure. It doesn't controls boundary anyhow. The allowed addresses are controlled by the `array->nelts` counter, where 0 means that there's no allowed elements for reading in by the `array->elts` pointer. When `array->nelts` set to 0, the `array->elts` pointer shouldn't be dereferenced and only indicates that there's no separate memory allocated for the elements.

> any other case this the boundary is useless and ineffective and is essentially an over-indicator of anything that you are overly focused on and give unjustified importance to, like the programmer before you, which in my humble opinion is wrong and it confirms a number of irrefutable deductive facts

If you're trying to say that the code here isn't easy, friendly and very intuitive, then you're right. Is it bad? It depends on the projects goals. One of the goals is to squeeze every bit of performance.

src/nxt_array.c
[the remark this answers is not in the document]
VBart: The problem here is that the project you've chosen is really too complex for beginning. It has no purposes to teach someone as well.

You've probably read some books or watched some courses that showed some classical and simplified picture, that was easier to explain. And this is what you probably meant by "the principles of the generally accepted technical justification laid down in all the foundations of C programming". But they are not. These are just principles to safe beginers from mistakes in their first steps and to not blow their minds with too complexity and flexability. The reality isn't black and white: comments can be bad, changes from good to better can be bad, pointers can be used in some tricky way and point to random addresses, and "goto" operator is fine when used the right way, and etc... etc.... etc.. This all can't be explained without years of experience. The same code looks very different for a beginner and for someone with 20+ years of programming in C. Personally, I've started programming in C when I was 14yo, now I'm 35.

This is one aspect. The other aspect is that every code have some costs. I've spent time writing this message and reviewing this change, you've spent time on discussions and preparing this pull request, studying the code. Then others devs will spend time to get familiar with newer version of the code and a QA will spend time to test it. Time costs money, even if you're doing it as a hobby, then you've just pay for it by your own time and energy, that in other case could be spent on something profitable. Since everybody have to eat, have to heat their homes and pay for electricity, there are no free reviews, free pull requests, free changes, or free education. Even if it looks like, then it's just paid by someone else.

The existing code has been written by someone from the team, then has been reviewed by others, and tested by QA. And all that were paid by NGINX, Inc, part of F5.

One more aspect is that most of the code will be rewritten multiple times during project evolution, so it doesn't make sense to focus too much on polishing some existing code without clear requirements. It will be thrown away as soon as the requirements and use cases will be changed.

Another one is that there's no perfect code. The perfect code is no code at all. Any existing code is a huge compromise between performance, readability, flexibility, maintainability, and time spent on it.

And one more is that the code details don't really matter for the end user, as long as it's safe, well tested, works good and fast enough, and does what the user wants.

Taking into account all of this, would I accept and spend much time on a patch, that doesn't fix any real bug and doesn't introduce any real feature, but just makes the code look more canonical to the articles and books on C programming for beginners? No, just because the benefits aren't clear, but cost of it is far from zero, it's not well tested, and who knows, maybe it introduces some bugs, that we've failed to spot.

Look at the list in your pull request descriptions:
> removed incorrect array initialization (nxt_array_init) with a pointer to a non-existent set of elements;

Does this array initialization introduce some bug? No.
Is such pointer prohibited by C-standard, or has undefined behaviour? No, as long as it isn't dereferenced. And in this case it's not dereferenced.

So, why is it incorrect? Just because someone thinks that pointers in C should always point to only existent set of elements? They shouldn't. You can do a lot of fun stuff with pointers, but in order to do it safe, you should have some good level in C programming. Devs of this project has very good level of C programming. We don't hire with lower level.

> fixed the size of the allocated memory for storing the array structure (nxt_array_t) - waste of memory;

Does it introduce a memory leak? No.
Is it just a simple fix of dumb typo in allocation size? No.

Unused memory doesn't mean wasted memory. Allocating more memory than needed usually is used for simplifying runtime code (like in this particular case) or better aligning of structures.

> fixed initialization of a set of elements (elts), pointer, memory size;

This one looks similar to the first one. So, not clear what specifically it's about.

> fixed the formula for calculating the required amount of memory (nalloc) allocated for the created array;

Does the previous formal have bugs, e.g. giving less amount of memory than required? No. Not clear, what exactly was fixes... what is the real issue with it?

> eliminated the possibility of early deletion of an existing pointer to a set of elements (elts) at the time of memory reallocation;

Could you clarify how it affects the existing project code? Early for what? As we've discussed already since there's no concurrent access to the pointer and cannot be any concurrent access at all, since the whole API isn't thread-safe, then nohow. Current code has no such possibility and absolutely safe here.

> added descriptions of functions;

Already discussed. The best description of any function is the code of this function. Sure there are different points of view on this aspect, it depends on the team, project, programming language, and etc...

Have you seen descriptions near other similar functions? No, then no reasons to think that those descriptions will be welcome and considered as an improvement.

> added a function to get a pointer by the index of an element in an array.

Does it improve some functionality of the project? No, it doesn't. It doesn't introduce any new feature. It's not required in the existing code.

Will it be required in the future? Maybe, but should it be introduced before such feature will be implemented? No, it just some additional change to the code, that, as I've explained before, has costs.

src/nxt_array.h
VBart: `uint16_t` for `size` and `nalloc` has been chosen also by Igor, because it reduced the whole structure size by 8 bytes (on the 64-bits architectures). This is free economy as there are no such big arrays expected anywhere and the difference in performance for `uint16_t` in this case isn't measurable at all.

VBart:
> There are too many unresolved issues and problems that I found in the Unit product immediately after the first launch of the application, don't you think?

I think the product that has no issues is the one that nobody use. Then more use cases the product are used, then more issues will be found. That shows good interest for the product for different use cases, even those not handled well yet.

Thanks.

This is really a very young product with lots of basic stuff not implemented yet, so what would you expect? If you'd like to look at a really matured product, then look at nginx, which has development history since 2002. The Unit shares no code with nginx and it's a young teenager in terms of development time and feature-richness.

> Maybe the product really lacks refactoring for the sake of refactoring or code review for the sake of code review?

How is refactoring for the sake of refactoring can solve those real issues users have?
And everyone who knows how meticulous our reviews are will laugh on this "lacks code reviews".

src/nxt_array.h
VBart: > You have singled out only one use case out of several dozen different ones, using, among other things, the nxt_array_create you mentioned earlier, which, as you said, has the same state as nxt_array_init when n == 0.

Sure, it makes sense to talk only use cases existing in the current code. Sorry, I can't talk about any imaginable use-cases or future use-cases. It's not a public API, and we don't develop arrays for general purposes. This API focuses on the current code and current use-cases.

> For nxt_array_init, the memory for storing array->elts is generally initially is not allocated, so I don't understand what kind of freed memory you are talking about when mentioning requests.

The memory where the array structure itself is stored.

> Moreover nxt_array.c and nxt_array.h are not only used for queries and in other cases may have a longer abandoned and unreleased lifetime for array->elts using nxt_array_create. I remind you that we originally talked about wasted memory using nxt_array_create when n > 1.

Could you clarify what memory is wasted when nxt_array_create() is called with n > 1?

Please also take into account:
1. Saving on syscalls is a WAY more important from performance point of view than doing free as soon as possible.
2. All allocated memory from memory pools will be freed anyway with destruction of that pool.
3. In nginx, for example, you barely find any free() call in runtime. It only allocates memory and never frees it during request processing. So, in your terms it wastes a lot of memory, but in fact it just frees it with some delay (when the request processing ends). This delay isn't significant comparing to performance benefits of not freeing memory during request processing. Unit follows the same ideology, but due to more complex design of memory pools, it sometimes  calls nxt_mp_free().

> I don't understand where you saw that I suggest to dereference the pointer after it has been initialized? Please re-read my answer to you earlier: https://github.com/nginx/unit/pull/664#discussion_r836379187

I don't understand what is wrong with setting:
```
array->elts = nxt_pointer_to(array, sizeof(nxt_array_t));
```
in context of `nxt_var_query_s`.

I don't care where it points in this case at all. It's the same as setting it to NULL, or 0... just some value. Then you can check for NULL, or can check for `nxt_pointer_to(array, sizeof(nxt_array_t));`. To avoid adding more conditions the value of `nxt_pointer_to(array, sizeof(nxt_array_t));` is used.

> Instead of being NULL for example until array->elts is fully initialized with nxt_array_add for example?

Then you will need additional conditions as instead of 2 states of this pointer:
1. point to the end of `nxt_array_t`
2. point to somewhere else

You will have 3 states of this flag:
1. NULL
2. point to the end of `nxt_array_t`
3. point to somewhere else

Then you will need to check for 3 states instead of 2 states.

And, if you will always allocate `array->elts` memory separate from the `nxt_array_t` structure, then you will always have 2 allocation instead of 1 allocation when creating a new array.

> As we have already been able to find out. Performance - in your case, just a cover for not considering a different point of view, forcing you to accept your own technically unreasonable.

Please provide a point that actually improves something and clear explanation of that improvement.

Check for 3 states instead of 2 states isn't improvement, but opposite - pessimization.
2 allocations instead of 1 allocation isn't improvement, but opposite - pessimization.

VBart:
> The fact that the improvements and fixes I've suggested don't align with your main product release plan - yes, I can afford it, as I'm proposing those improvements and fixes that I personally need first, but often affect most other users of your product as well. As it was before with issues:
listen sockets (unix) https://github.com/nginx/unit/issues/643 - issue not resolved;
source mask (client_ip - X-Forward-For for unix) https://github.com/nginx/unit/issues/642 - solution found, but still not included in the main branch;
Python module with ASGI "error 503" https://github.com/nginx/unit/issues/635 - issue not resolved.

The real answer, why Unix sockets are handled so bad is just nobody tried to implement them yet. Why does Unit support Unix sockets then? Because the Unit was forked from some other project and we've just inherited that support as it was.

src/nxt_array.c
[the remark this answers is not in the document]
VBart: > You are already at least the fifth time trying to catch on to the changes I proposed, which I immediately refused at the first answer of @hongzhidao:
https://github.com/nginx/unit/pull/664#discussion_r836244623

He asked me to look at this patch. I did. If you've already refused those changes, then what we discuss?

Sorry, but I'm completely lost in what your proposal is. If we don't discuss the current change, then what change you suggest? Could you please update the pull request then.

> What is your goal by doing this?

I've tried many times to explain why that pointer is initialized to that value instead of NULL. It's initialized that way because it's a correct way of initializing nxt_array_t structure for zero number of elements, that's it.

> In many ways, you are trying to explain things that are absolutely banal and understandable to everyone (costs, time, etc.), which I told you about even earlier (https://github.com/nginx/unit/pull/664#discussion_r837444780), but for some reason you they assumed that I could not know about it.

Because this particular pull request has many "fixed" words in description and doesn't fix any bug. It just changes the code in a way that you think is better.

> Also note the fact that it is not I who can give you the technical justification for the logic I propose and explain its benefits, and you cannot do the same for your own logic. Based on your own logic, this project is difficult for you, not for me.

It's you, who proposed this pull request, not me. So, it's your obligation to give the technical justification for every single line of this change.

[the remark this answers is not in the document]
echolimazulu: > He asked me to look at this patch. I did. If you've already refused those changes, then what do we discuss?
>
> Sorry, but I'm completely lost in what your proposal is. If we don't discuss the current change, then what change do you suggest? Could you please update the pull request then.

If you understood this only from the fifth time, then you need to ask this question to yourself.

> correct way of initializing nxt_array_t

The only attempt that I consider as a real attempt by you to do this: https://github.com/nginx/unit/pull/664#discussion_r837876182
Of course there were others, but you abandoned almost all of them after my first leading question.

> Because this particular pull request has many "fixed" words in description and doesn't fix any bug. It just changes the code in a way that you think is better.
>
> It's you, who proposed this pull request, not me. So, it's your obligation to give the technical justification for every single line of this change.

I don't see anything wrong with offering a new, I think improved vision. If you do not consider it as such, then it is your responsibility as a professional with 20+ years of C programming experience to briefly and clearly explain why you think so, as @hongzhidao did it with regards to the design of the application, using only **2 sentences for this**, and **not using dozens of what seemed to me saturated with aggression comments**, to the first real attempt to at least somehow expand his vision using the technical justification for your position, without twisting previously said by me and you, without jumping from one question to another and without using other tricks in order to confuse your interlocutor.

I have stated my position regarding the pointer (`array->elts`), the abandoned allocated memory (`nxt_array_create, n> 1`), the early dereference of variables (`array->elts`) and other points in great detail and repeatedly. I think that even your colleagues will agree with this statement.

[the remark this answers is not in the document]
echolimazulu: I believe that in order to complete our debate with you, it is necessary to involve a third party. Let it be the managers of your product and company. And you and I will unconditionally accept their decision on the ideological issues raised in this discussion and will continue to rely on their vision of what is correct from the opinions voiced by both parties and whether your current position reflects the internal values ​​​​of F5 Networks.

@tippexs, @lcrilly Good afternoon! 
Sorry for the worry. Please give us some of your valuable time and give us an assessment of what is happening from your point of view. Thank you!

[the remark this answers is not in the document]
hongzhidao: > 

> This is a review thing and what @VBart did on it. This is not just for you, but also for our internal review.
> 
> > Maybe the product really lacks refactoring for the sake of refactoring or code review for the sake of code review?
> 
> First, the below picture is about the software source code quality.
> 
> 
> 
> You can see that Unit's code is pretty qualified.
> 
> Then, welcome to take a look at the insights of contributors. Our code is about 1 to 3 in terms of deletions and additions. It shows that we attach great importance to refactoring. https://github.com/nginx/unit/graphs/contributors
> 
> Sure, you are welcome and free to submit patches here. We didn't find why the patch makes sense, though we found some reasonable in your patch. This is normal for someone unfamiliar with Unit's source code.
> 
> So it's better to go back to the code itself. Unit is young, we have to choose which ones to merge and which ones to put behind. Thank you again for the issues found recently.

This is a review thing and what @VBart[](https://user-images.githubusercontent.com/8190423/160812238-00dbbaff-4533-4ded-8ff7-78ea25a28201.jpg) did on it. This is not just for you, but also for our internal review.

Maybe the product really lacks refactoring for the sake of refactoring or code review for the sake of code review?

First, the below picture is about the software source code quality.

1905745632

You can see that Unit's code is pretty qualified.

Then, welcome to take a look at the insights of contributors. Our code is about 1 to 3 in terms of deletions and additions. It shows that we attach great importance to refactoring.
https://github.com/nginx/unit/graphs/contributors

Sure, you are welcome and free to submit patches here. We didn't find why the patch makes sense, though we found some reasonable in your patch. This is normal for someone unfamiliar with Unit's source code.

So it's better to go back to the code itself.
Unit is young, we have to choose which ones to merge and which ones to put behind.
Thank you again for the issues found recently.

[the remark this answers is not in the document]
hongzhidao: This is a review thing and what @VBart[](https://user-images.githubusercontent.com/8190423/160812238-00dbbaff-4533-4ded-8ff7-78ea25a28201.jpg) did on it. This is not just for you, but also for our internal review.

Maybe the product really lacks refactoring for the sake of refactoring or code review for the sake of code review?

First, the below picture is about the software source code quality.



You can see that Unit's code is pretty qualified.

Then, welcome to take a look at the insights of contributors. Our code is about 1 to 3 in terms of deletions and additions. It shows that we attach great importance to refactoring. https://github.com/nginx/unit/graphs/contributors

Sure, you are welcome and free to submit patches here. We didn't find why the patch makes sense, though we found some reasonable in your patch. This is normal for someone unfamiliar with Unit's source code.

So it's better to go back to the code itself. Unit is young, we have to choose which ones to merge and which ones to put behind. Thank you again for the issues found recently.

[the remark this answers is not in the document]
echolimazulu: I would like to clarify. I do not insist on the merger of this PR. The changes I propose are not vital to me.

By coincidence, the discussion that has arisen on the basis of the changes I propose can be regarded as checking the state of compliance with the publicly declared values ​​​​of F5 Networks in relation to new participants (perhaps even employees) interested in developing a separate project and the company as a whole, as well as identifying in the current team persons potentially hindering their development.

I also ask you not to correct my comments (https://github.com/nginx/unit/pull/664#discussion_r838277973) unless they violate the GitHub Community Guidelines. Each such change in my comments, I will officially record.

I asked your colleague to technically substantiate for me the logic of the current version of the code he defends, which he did not initially do, and in some places entered into an unconstructive debate.

I am potentially interested in all possible help to me (as an open source contributor) in the development of the product and I will be glad for the subsequent prosperity, otherwise I would step back from the discussion and close this PR.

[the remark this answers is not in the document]
hongzhidao: >I also ask you not to correct my comments (https://github.com/nginx/unit/pull/664#discussion_r838277973) unless they violate the GitHub Community Guidelines. Each such change in my comments, I will officially record.

It was my mistake. I commented on your message, but I have no experience in commenting Github pull request board. I knew it after I submit a comment but it covered your comment. So I tried to recover it.

VBart:
> If you use the word "performance" so often, then why is the work of unix sockets in Unit still not properly implemented, and the importance of their implementation is equated by you with a historical derivative from some other project?

Simply because unix sockets are useful only for localhost connections, while the primary goal of Unit has always been to directly deal with the network without any local proxy in front of it.

VBart:
> No matter how well you concentrate on the number and frequency of system calls, the lack of full support or incorrect work of unix sockets negates all your efforts to optimize the program's operation at the same time as using the word "performance".

No matter what you will use, tcp/ip or uds, the overhead of any additional proxy layer will be much bigger.

echolimazulu:
> > > > If you use the word "performance" so often, then why is the work of unix sockets in Unit still not properly implemented, and the importance of their implementation is equated by you with a historical derivative from some other project?
> > > 
> > > 
> > > Simply because unix sockets are useful only for localhost connections, while the primary goal of Unit has always been to directly deal with the network without any local proxy in front of it.
> > 
> > 
> > **You probably don't know the following real use cases:**
> > ```
> > * unit using connection to it via unix socket in the cluster;
> > 
> > * nginx as a proxying frontend server, for example, to use compression or caching of the unit application.
> > ```
> 
> Compression and caching are already in the roadmap for Unit. If you think that some feature has more priority than other features, then you have two choices, either implement it yourself, or just show your interest in it (what you already did, thanks).
> 
> But, please, be respectful that others can have their own hit-parade of "must have" features.

To be honest, I did not expect to hear from you anything other than the veiled: "Get in the general queue, you are the last one."

I respect the needs of your customers and your master plan. The problem is not my attitude towards your plan, but your attitude towards the needs of real users, such as myself, and taking their needs into your release plan. 

In case you haven't noticed, it's you who stubbornly resists unix socket support by trying to express it in your opinion from a developer's point of view, arguing that there is no real use case for unix sockets by users. But here's the problem, I'm the same user who would like to use your program with unix sockets. **Where can I see the updated implementation plan and deadline?**

I am completely satisfied with the caching and compression implemented in nginx, as well as many other features in it that I have been working with for many years.

It makes me extremely frustrated when I see a project like Unit with huge growth potential and the best development team, the quality of work of which I am very familiar with and has almost no direct competitors and analogues in some areas, but does not implement the basic functionality of most that is often available by default in much smaller projects, with less popularity and less professional development teams, most of which do not position themselves as high performance, as they are written in C.

VBart:
> In case you haven't noticed, it's you who stubbornly resists unix socket support by trying to express it in your opinion from a developer's point of view, arguing that there is no real use case for unix sockets by users.

So, I've asked Alejandro some weeks ago to work on #642 and suggested him a configuration, and now you blame me, that I'm somehow resisting the implementation of the Unix socket.

I haven't said that there are no real use-cases. I've pointed out, that this feature wasn't the most demanded one.

tippexs:
Good evening / night / morning to all! First of all - Thank you very much to all of the great comments and discussions here @VBart  and @echolimazulu  but I have a feeling we are finally reached the top of the mountain here.

As so many things were already discussed in great detail I would just like to pick up the open questions or parts you wanted a comment on @echolimazulu .

1. The overall architectural concept

We as a team have full trust in @VBart experience and opinion as a dev leader for this product. I think over the time of this PR we lost a little bit track of what we were discussing or asking at the beginning.

The changes you suggested in the PR are great ideas and first steps into the UNIT source code. Of course there are not aligned with the overall internal structure of the project. How could they even after such a short period of time but this is not the case at all. The fact you have worked with the code and the internal API is a good step in a direction towards other implementations and features.

As the array functions as well as many other underplaying „Unit kernel“ features are the bred and butter of the product it is not easy to change them. This does mean we are never doing it. If features are implemented changes to this inner kernel are important and we are doing them. Having your opinion on different implementations we made are valuable because of very good reason. CHALLENGING. With each and every question you are asking we are checking the source again and ask ourselves if the changes are worth to implement. And btw this is the same we are doing internally. Just imagine how often I have submitted ideas and changes and ended up in a similar discussion . But there is nothing bad about it. We learn a lot from those discussions. Stop learning and challenging yourself and the product will be the end for everything on this planet. So please - keep up asking question and challenge the implementation. As you already noticed we are more then happy to answer as many questions you have. Feel free to use the NGINX Unit mailing list or check out the NGINX community Slack channel to work with the team on new features and resolve problems. That’s why we are here on Github, Slack and friends.

I believe the questions regrading memory management are discussed well enough for the moment and we can move this discussion to some other platform if you want and focus on the Problem we are trying to solve here.

2. UNIX Socket Support as a listener in UNIT

I was working with @alejandro-colomar on the socket issue with aegi as described in #635 and we made great progress. We are working on a way to make the roadmap as well as the features / issues for upcoming releases more transparent for the community. This would make it easier to track what’s scheduled and what will be the next important feature.

That said as pointed out by @VBart and @alejandro-colomar your feedback about the UNIX socket implementation was very much appreciated and we are working on a fix for the asgi implementation as well as common test to cover the overall support as unix sockets as Unit listeners.

Atm, we are working with a fork of your repo. To be efficient I would like to invite you to collaborate with @alejandro-colomar and myself on this fork.

Last words from me.

I was always reading each and every comment of this PR at least twice to fully understand what’s the point of the discussion and made the decision to write my comment now as I saw we are technically reaching the top of the mountain here. I would make the suggestion we are spending our energy and enthusiasm on the Issue #635 and create a new feature for Unit and start thinking about the next one and the next one and the next one. I will close this comment with a quote I always like to remember myself about

“A ship is safe in harbor, but that’s not what ships are built for.”
- John G. Shed.

In this sense. See you in issue #635 and work on a great new feature. I would like to thank you all for a great discussion and all the time you have spent on this. I really looking forward to talk in this round again about ideas, features, problems and visions.

src/nxt_array.c
alejandro-colomar: Apart from what @VBart said, allocating more than what you need is actually a good thing, not only because of the direct impact of the reduced number of `malloc(3)` and `free(3)` calls, but because of how it works underneath:

The system tries to find a contiguous block of memory that is big enough for what you asked.  If you ask big chunks of memory (especially good if you allocate 4K blocks), it's especially easy for the system.  Then, you deallocate big chunks at once, and the free memory tables receive the full blocks of memory at once, and can reuse them easily.  If you allocate small amounts of memory, and you deallocate them as soon as you can, then the free memory blocks become small and disperse, which becomes a problem because the allocator has to check each of those small chunks to see if they meet the requirements.  That's the same thing that happen(ed) to hard drives: fragmentation.

Yes, we do explicitly waste memory, but systems have more and more memory these days.  I guess we can waste some.  The benefits are:  less calls to `malloc(3)` and `free(3)` *and* each of those few calls is faster because of having "defragmented" memory.

src/nxt_array.c
alejandro-colomar: Ignoring if this change is good or not, which should need much more investigation from my side...

Please, when submitting changes that are logically different from each other, submit them as small separate commits.  Even if it all makes sense together as a cleanup (so everything makes sense as a pull request), each small change makes sense separately as a single commit.  It is much easier to review them that way, and check correctness of each small bit.  That allows that some changes may be rejected while others may be cherry-picked.

src/nxt_array.c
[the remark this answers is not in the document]
alejandro-colomar: Hi @VBart and @echolimazulu,
> > removed incorrect array initialization (nxt_array_init) with a pointer to a non-existent set of elements;
> 
> Does this array initialization introduce some bug? No. Is such pointer prohibited by C-standard, or has undefined behaviour? No, as long as it isn't dereferenced. And in this case it's not dereferenced.
> 

I couldn't review the code, because the changes are huge.  Small 5 line commits with 50 lines of commit message detailing motivation and references to the ISO C Standard and other sources, and possibly testing would greatly help.

Said that, I'm concerned that in some case we might be actually invoking Undefined Behavior.   Since I haven't worked yet with that part of the code, I'm not yet familiar with it, but if I see a small commit proposing a fix to a possible UB, I'll happily review it.

I'm especially concerned because compilers are becoming more aggresive with each release, and the latest versions of the Standard disallow not only dereferencing invalid pointers, but define some cases where even storing or reading a pointer pointing to some invalid location will already invoke UB.  For example, one may not read a pointer after one past an array (which in the case of a dynamic array is not very clear if it's the requested size, or the actual allocated size), or one may not read a pointer to a deallocated (or reallocated) array after the `free(3)` or `realloc(3)` call.

src/nxt_array.h
alejandro-colomar: I highly recommend you to turn off any automatic indentation that your editor has.  I have yet to see any autoindent feature that I consider actually good.  Maybe the one in `vim(1)`... but not even that.

Check https://nginx.org/en/docs/dev/development_guide.html#code_style for our style guide.

echolimazulu:
> > In case you haven't noticed, it's you who stubbornly resists unix socket support by trying to express it in your opinion from a developer's point of view, arguing that there is no real use case for unix sockets by users.
>
> So, I've asked Alejandro some weeks ago to work on #642 and suggested him a configuration, and now you blame me, that I'm somehow resisting the implementation of the Unix socket.

I'm sorry, but I cannot allow you to try to resort to another trick to sway public opinion in your favor by taking the **position of "victim"**. In this regard, I will give the last time clarifications on the message you wrote.

**To resist does not mean to do nothing at all.**

You are certainly well done for responding in a timely manner and suggesting @alejandro-colomar to solve this problem (perhaps guided by the job description) and you have no idea how much I was inspired by this step on your part. After you promptly decided to deal with this problem, I placed great hopes on this project and chose it as the basis for my own projects, as I saw the work of a professional in action, which I expected to see by choosing this project. But let's call a spade a spade. Both you and I are well aware that the problem and its solution that you are talking about (https://github.com/nginx/unit/issues/642) is rather related to the fact that in the initial implementation of client_ip (X-Forwarded-For) the use case with unix sockets was missed and this function is inherently not related to the work of unix sockets themselves, but is related to the function of parsing request headers containing the client address when the Unit is behind the front proxy server. As for the work of the unix sockets themselves, you forgot to mention the problem: https://github.com/nginx/unit/issues/643, which has not received any response in the request body for almost a month and a half since the publication. I think that other comments on this issue would be redundant.

> I haven't said that there are no real use-cases. I've pointed out that this feature wasn't the most demanded one.

Sorry, maybe I really misunderstood you. Or my mind is already biased and falsely perceives your answers associating it with your previous answers affecting real use cases. Yes, similar if not literally written by you, nevertheless, I can extract it by reading between the lines of your answers like:

> Simply because unix sockets are useful only for localhost connections, while the primary goal of Unit has always been to directly deal with the network without any local proxy in front of it.

All I'm doing here, outside of PR itself, is asking you and your colleagues about the current state of the product from a C programming novice's point of view, and stating facts trying to reason with my current experience, sometimes trying to draw conclusions from your answers. The way you perceive and react to my appeals, unlike your colleagues, is not a problem for me and the reason why I should refuse this kind of activity. In my understanding of the situation, you yourself, without my help, are doing everything in order to decredit yourself in every possible way, when I do not pursue such a goal in relation to you. I am not your competitor or rival, and we are working with you for a positive result for all of us, in this regard, we must strengthen each other when working with a common challenge, and yes, sometimes this can happen through consideration or acceptance of an alternative point of view - this is normal .

In the future, I will try to follow the opinion of @tippexs (https://github.com/nginx/unit/pull/664#issuecomment-1083655332), as it seems to me the most constructive for product development.

src/nxt_array.h
echolimazulu: > I highly recommend you to turn off any automatic indentation that your editor has. I have yet to see any autoindent feature that I consider actually good. Maybe the one in `vim(1)`... but not even that.
> 
> Check https://nginx.org/en/docs/dev/development_guide.html#code_style for our style guide.

Wow! Really very useful information, thank you for this, since earlier I had to almost intuitively determine what style of code you use.

lcrilly:
@echolimazulu, please contact me: liam@nginx.com 
I'd like to share my perspective on the project goals and better understand what motivated you to submit this PR.

VBart:
> I'm sorry, but I cannot allow you to try to resort to another trick to sway public opinion in your favor by taking the **position of "victim"**. In this regard, I will give the last time clarifications on the message you wrote.

I'm not playing those games, nor need to, neither want to. You asked questions or made wrong assumptions, I've tried to answer and explain some of the reasons behind. You can listen and learn from that, or you, for sure, have a right not to do so. I actually could not answer you at all. There's no obligation to, it was just my good will to explain.
 
> 
> **To resist does not mean to do nothing at all.**
> 
> You are certainly well done for responding in a timely manner and suggesting @alejandro-colomar to solve this problem (perhaps guided by the job description) and you have no idea how much I was inspired by this step on your part. After you promptly decided to deal with this problem, I placed great hopes on this project and chose it as the basis for my own projects, as I saw the work of a professional in action, which I expected to see by choosing this project.

That's nice to hear to me as a founder of this project.

> But let's call a spade a spade. Both you and I are well aware that the problem and its solution that you are talking about (#642) is rather related to the fact that in the initial implementation of client_ip (X-Forwarded-For) the use case with unix sockets was missed and this function is inherently not related to the work of unix sockets themselves, but is related to the function of parsing request headers containing the client address when the Unit is behind the front proxy server.

The "client_ip" was implemented by utilizing already existed function used for routes matching (`nxt_http_route_addr_rule_create()` and all the machinery behind it). Adding support for unix sockets to the IP-address matching function clearly is a separate task (and actually a separate feature of "unix" sockets matching to routing). Nobody before you ever asked to support unix sockets in "source" or "destination" matching.

> As for the work of the unix sockets themselves, you forgot to mention the problem: #643, which has not received any response in the request body for almost a month and a half since the publication. I think that other comments on this issue would be redundant.

If I had unlimited resources, I'd assign someone to fix every issue and implement every feature, that anyone ever asked. Everybody would be happy, and there would be no one waiting for fix or implementation for a week, a months, or for years.

> 
> > I haven't said that there are no real use-cases. I've pointed out that this feature wasn't the most demanded one.
> 
> Sorry, maybe I really misunderstood you. Or my mind is already biased and falsely perceives your answers associating it with your previous answers affecting real use cases. Yes, similar if not literally written by you, nevertheless, I can extract it by reading between the lines of your answers like:
> 
> > Simply because unix sockets are useful only for localhost connections, while the primary goal of Unit has always been to directly deal with the network without any local proxy in front of it.
> 

I don't put any additional sense between words or between lines. I calmly explained why support for unix sockets wasn't considered from performance critical perspective (since you appealed to performance in that message), and since it wasn't popular feature among requests from our users, and since I don't have unlimited resources - nobody worked on it yet.

Pointing that some other product has some feature isn't useful for prioritization. There are a lot of features that some other products (even more simple) have, but we don't have yet, unfortunately. And depending on the group of users you will ask, they will name different products and different features. E.g. I know a big amount of users, who have been asking about Dot Net support for 5 years already, and I'm glad that finally we have started R&D in on such module.

> All I'm doing here, outside of PR itself, is asking you and your colleagues about the current state of the product from a C programming novice's point of view, and stating facts trying to reason with my current experience, sometimes trying to draw conclusions from your answers. The way you perceive and react to my appeals, unlike your colleagues, is not a problem for me and the reason why I should refuse this kind of activity. In my understanding of the situation, you yourself, without my help, are doing everything in order to decredit yourself in every possible way, when I do not pursue such a goal in relation to you.

You've sent a pull request, which means a request for review, and my colleagues asked me to look at it specifically, which I did, like we all do every time for every change. Our usual approach is that every line of every change need to have a good reason. You've asked additional questions, I've answered them. You can learn, you can agree, you can disagree, that's your right. I even explained much more than should in this case. I don't see  why you've turned into blaming someone in something and started to be so emotional. You've already blamed me that I take position of a victim, asked for more managers, and now you show me your discreditation, that even not mentioning your sarcastic answers.

> I am not your competitor or rival, and we are working with you for a positive result for all of us, in this regard, we must strengthen each other when working with a common challenge, and yes, sometimes this can happen through consideration or acceptance of an alternative point of view - this is normal .

Then, maybe, it makes sense to return back to the programming.

VBart:
btw, you may find these links useful:
https://tirania.org/blog/archive/2010/Dec-31.html
https://teejeetech.medium.com/etiquette-in-open-source-projects-f302b6e19c38
https://dev.to/azure/contributing-to-open-source-projects-contributors-etiquette-1bdm

src/nxt_array.c
echolimazulu: Now I understand. Thank you for the detailed explanation. I thought about it, but I did not understand the principles of defragmentation in Unit so well and the high cost of each individual call. Now it's all clear. Thanks again!

src/nxt_array.c
echolimazulu: I'm sorry, when I answered in the previous message, the page with your answer has not yet been updated. Heard you.

src/nxt_var.c
echolimazulu: I agree, I underestimated the design of the project in this place, and also did not know about the high cost of calls

src/nxt_array.c
```diff
@@ -1,150 +1,253 @@
...
+/*
+ * nxt_array_destroy - Destroy the existing array.
+ */
 void
 nxt_array_destroy(nxt_array_t *array)
 {
-    if (array->elts != nxt_pointer_to(array, sizeof(nxt_array_t))) {
-        nxt_mp_free(array->mem_pool, array->elts);
-    }
+    if (nxt_fast_path(array != NULL)) {
```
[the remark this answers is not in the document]
echolimazulu: Understood, fixed it in my notes.

src/nxt_array.c
```diff
@@ -1,150 +1,253 @@
...
+
+        array->nelts = 0;
 
-    nxt_mp_free(array->mem_pool, array);
+        (void) nxt_mp_free(array->mem_pool, array->elts);
+        (void) nxt_mp_free(array->mem_pool, array);
```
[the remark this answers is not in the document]
echolimazulu: yes, there is another mistake on my part besides this one. thank you.

echolimazulu:
> @echolimazulu, please contact me: [liam@nginx.com](mailto:liam@nginx.com) I'd like to share my perspective on the project goals and better understand what motivated you to submit this PR.

Hello. Thank you for your attention. I emailed you as requested.

Issue #676, opened 2022-04-13
nginx unit fails when web application load time is longer than 60 seconds


Phylu:
I have a python web application, that has a very long loading time due to some heavy database calculations that need to be run one on start time of the application. When running the application in a docker container with nginx unit, the setup of the application fails. The curl request in the docker entrypoint that configures the web app (https://github.com/nginx/unit/blob/master/pkg/docker/docker-entrypoint.sh#L43) fails with the following error when the application takes longer than 60 seconds to start:

```
curl: (52) Empty reply from server
```

Here is my sample application code:

Directory Structure:
```
- app.py
- config.json
- Dockerfile
- requirements.txt
```

Dockerfile:
```
FROM nginx/unit:1.26.1-python3.9

COPY requirements.txt /config/requirements.txt
RUN pip install -r /config/requirements.txt 

COPY config.json /docker-entrypoint.d/config.json
COPY app.py /www/app.py
```

app.py:
```
from flask import Flask
import time

app = Flask(__name__)
application = app

# Artificially ensure a high load time
time.sleep(70)

@app.route('/')
def hello_world():
    return 'Hello, World!'
```

config.json:
```
{
    "settings": {
        "http":{
            "header_read_timeout": 300,
            "body_read_timeout":300,
            "idle_timeout":300,
            "send_timeout":300        
       }
    },
    "listeners": {
        "*:8080": {
            "pass": "applications/python_app"
        }
    },
    "applications": {
        "python_app": {
            "type": "python",
            "path": "/www/",
            "module": "app",
            "callable": "app",
            "limits":{
                "timeout": 300,
                "requests": 500
            }
        }
    }
}

```

requirements.txt:
```
Flask==2.0.2
```

I tried to configure all timeouts in a way that the should not produce problems, but this did not help. Is there anything that I am doing wrong?

Thanks for any support in advance already!

tippexs:
Hi @Phylu thanks for reaching out. You are right - that sounds like an timeout. Have you tried that without docker? Does it behave the same way? To be able to reproduce are you able to share your main python file? I will try to reproduce this with a simple timeout in the main startup phase of the application.

Phylu:
Hi @tippexs 
Thanks for the response! I am not sure how to test this with unit outside of Docker easily due to this special configuration logic. You can find my main python file (`app.py`) in the issue description.

tippexs:
Perfect. That is the exact same approach I am following right now. Let me quickly test this with and without docker and share the result here.

[...]

PR #681, opened 2022-04-25
Increase readtimeout for configuration endpoint

tippexs:
As reported in issue #676. In case an application or better to say the set off all applications will take longer then `60` seconds to start, Unit will close the connection to the client sent the new configuration (curl for example).

As the timeout value in this case will not effect Unit during runtime, it is a small change without much impact. One word about the downsides of it. If Unit is hosted in a container, and an application failed to start or is waiting on a database or any other 3rd party connection to be established, the container will not be killed till the new timeout of `300` seconds exceeded. But I think this is a valid point and nothing we have to worry about.

### DIFF
--- a/docs/changes.xml
+++ b/docs/changes.xml
@@ -31,6 +31,13 @@ NGINX Unit updated to 1.27.0.
          date="" time=""
          packager="Andrei Belov &lt;defan@nginx.com&gt;">
 
+
+<change type="change">
+<para>
+increased the applications startup timeout.
+</para>
+</change>
+	
 <change type="feature">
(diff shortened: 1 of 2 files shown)


hongzhidao (approved):

alejandro-colomar (approved):

alejandro-colomar:
Could you please add a line in docs/changes.xml?

tippexs:
@alejandro-colomar done. @artemkonev pls review the changelog.

alejandro-colomar:
Merged.

PR #682, opened 2022-04-26
Added NXT_MAYBE_UNUSED for __attribute__((__unused__)).

alejandro-colomar:
When testing some configurations of compilers and OSes, I noticed
that clang(1) 13 on Debian caused a function to be compiled but
unused, and the compiler triggered a compile error.

To avoid that error, use __attribute__((__unused__)).  Let's call
our wrapper NXT_MAYBE_UNUSED, since it describes itself more
precisely than the GCC attribute name.  It's also the name that
C2x (likely C23) has given to the standard attribute, which is
[[maybe_unused]], so it's also likely to be more readable because
of that name being in ISO C.

### DIFF
--- a/auto/clang
+++ b/auto/clang
@@ -176,3 +176,21 @@ nxt_feature_test="struct s {
                       return 1;
                   }"
 . auto/feature
+
+
+nxt_feature="GCC __attribute__ unused"
+nxt_feature_name=NXT_HAVE_GCC_ATTRIBUTE_UNUSED
+nxt_feature_run=
+nxt_feature_incs=
+nxt_feature_libs=
+nxt_feature_test="static void f(void) __attribute__ ((__unused__));
+
+                  static void f(void)
+                  {
+                      return;
+                  }
+
+                  int main(void) {
+                      return 0;
+                  }"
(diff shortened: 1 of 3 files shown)


ac000:
This is due to the fact that clang warns about unused static inline functions (in .c files) and gcc does not, as is discussed [here](https://lore.kernel.org/lkml/alpine.DEB.2.10.1705241400510.49680@chino.kir.corp.google.com/) and [here](https://lore.kernel.org/lkml/20170530181306.GV141096@google.com/). There is *debate* to which is *correct*.

But marking it as *maybe_unused* is certainly a valid response.

alejandro-colomar:
BTW, I just checked again, and we have a mix of uppercase and lowercase identifiers for attributes.  Which do we prefer? `NXT_MAYBE_UNUSED` or `nxt_maybe_unused`?

hongzhidao:
It seems if it's around function declaration, we use uppercase. But `nxt_inline` is lowercase.
Here I prefer `NXT_MAYBE_UNUSED`.

hongzhidao (approved):

PR #683, opened 2022-04-26
Added .gitignore.

alejandro-colomar:


ac000:
It would be customary to ignore editor backup files, the following covers vi/vim and numerous other editors.

```
*.swp
*~
```

alejandro-colomar:
> It would be customary to ignore editor backup files, the following covers vi/vim and numerous other editors.
> 
> ```
> *.swp
> *~
> ```

Thanks for the suggestion.  That was my initial proposal too.  But after some internal discussion, we agreed that the project gitignore should only ignore project-related files.  Files created by the local editor should go to a personal git exclude file; see <https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/> and <https://stackoverflow.com/questions/7335420/global-git-ignore>, so that everyone can globally configure their own editor preferences without having to touch the project gitignore.

alejandro-colomar:
Merged.  But github didn't detect it.

PR #684, opened 2022-04-26
Supporting new "index" option.

alejandro-colomar:
This contains several patch sets in itself, but it's too complex for github pull requests.
The breakdown is:

<strike>
Arrays:
```
0f184fe (HEAD -> index-arr, gh/index-arr, alx/index-arr) Static: optimized "index" iteration.
e145bb3 Tests: added tests for "index" (array) option.
825982f Static: supporting multiple paths in the "index" option.
304a838 Simplified branching.
```
</strike>

Variables:
```
93371e9 (gh/index-var, alx/index-var, index-var) Tests: added tests for "index" (var) option.
a3532c9 Static: supporting variables in the "index" option.
5642e81 Generalized code.
```

Simple string:
```
2506eba (gh/index-str, alx/index-str, index-str) Tests: added tests for "index" (string) option.
53e4662 Static: supporting new "index" option.
```
Some cleanup and minor fixes before the patch set itself (already merged):
```
bc4990a (gh/ret404, alx/ret404, ret404) Static: returning 404 when "index" is a non-regular file.
1b5ce9c (gh/rename, alx/rename, rename) Renamed nxt_http_static_ctx_t field 'index' to 'share_idx'.

```

### DIFF
--- a/src/nxt_http_static.c
+++ b/src/nxt_http_static.c
@@ -19,7 +19,7 @@ typedef struct {
 typedef struct {
     nxt_uint_t                  nshares;
     nxt_http_static_share_t     *shares;
-    nxt_str_t                   index;
+    nxt_var_t                   *index;
 #if (NXT_HAVE_OPENAT2)
     nxt_var_t                   *chroot;
     nxt_uint_t                  resolve;
@@ -82,6 +82,8 @@ nxt_http_static_init(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
     nxt_conf_value_t        *cv;
     nxt_http_static_conf_t  *conf;
 
+    static const nxt_str_t  default_index = nxt_string("index.html");
+
     mp = tmcf->router_conf->mem_pool;
 
     conf = nxt_mp_zget(mp, sizeof(nxt_http_static_conf_t));
@@ -113,12 +115,18 @@ nxt_http_static_init(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
     }
 
     if (acf->index == NULL) {
-        nxt_str_set(&conf->index, "index.html");
+        str = default_index;
 
     } else {
         nxt_conf_get_string(acf->index, &str);
-        nxt_str_dup(mp, &conf->index, &str);
     }
+
+    var = nxt_var_compile(&str, mp, 1);
+    if (nxt_slow_path(var == NULL)) {
+        return NXT_ERROR;
+    }
+
+    conf->index = var;
 
 #if (NXT_HAVE_OPENAT2)
     if (acf->chroot.length > 0) {
@@ -223,9 +231,9 @@ nxt_http_static_iterate(nxt_task_t *task, nxt_http_request_t *r,
     nxt_http_static_ctx_t *ctx)
 {
     nxt_int_t                ret;
+    nxt_bool_t               shr_is_const, idx_is_const;
     nxt_http_static_conf_t   *conf;
     nxt_http_static_share_t  *share;
-    nxt_bool_t               shr_is_const;
 
     conf = ctx->action->u.conf;
 
@@ -236,7 +244,7 @@ nxt_http_static_iterate(nxt_task_t *task, nxt_http_request_t *r,
     nxt_str_t  idx;
 
     nxt_var_raw(share->var, &shr);
-    idx = conf->index;
+    nxt_var_raw(conf->index, &idx);
 
 #if (NXT_HAVE_OPENAT2)
     nxt_str_t  chr;
@@ -256,8 +264,9 @@ nxt_http_static_iterate(nxt_task_t *task, nxt_http_request_t *r,
 #endif /* NXT_DEBUG */
 
     shr_is_const = share->is_const;
+    idx_is_const = nxt_var_is_const(conf->index);
 
-    if (!shr_is_const) {
+    if (!shr_is_const || !idx_is_const) {
         ret = nxt_var_query_init(&r->var_query, r, r->mem_pool);
         if (nxt_slow_path(ret != NXT_OK)) {
             nxt_http_request_error(task, r, NXT_HTTP_INTERNAL_SERVER_ERROR);
@@ -265,7 +274,12 @@ nxt_http_static_iterate(nxt_task_t *task, nxt_http_request_t *r,
         }
     }
 
-    ctx->index = conf->index;
+    if (idx_is_const) {
+        nxt_var_raw(conf->index, &ctx->index);
+
+    } else {
+        nxt_var_query(task, r->var_query, conf->index, &ctx->index);
+    }
 
     if (shr_is_const) {
         nxt_var_raw(share->var, &ctx->share);
@@ -286,7 +300,7 @@ nxt_http_static_iterate(nxt_task_t *task, nxt_http_request_t *r,
 #endif
     }
 
-    if (shr_is_const) {
+    if (shr_is_const && idx_is_const) {
         nxt_http_static_send_ready(task, r, ctx);
 
     } else {
--- a/src/nxt_http_static.c
+++ b/src/nxt_http_static.c
@@ -19,6 +19,7 @@ typedef struct {
 typedef struct {
     nxt_uint_t                  nshares;
     nxt_http_static_share_t     *shares;
+    nxt_str_t                   index;
 #if (NXT_HAVE_OPENAT2)
     nxt_var_t                   *chroot;
     nxt_uint_t                  resolve;
@@ -30,6 +31,7 @@ typedef struct {
 typedef struct {
     nxt_http_action_t           *action;
     nxt_str_t                   share;
+    nxt_str_t                   index;
 #if (NXT_HAVE_OPENAT2)
     nxt_str_t                   chroot;
 #endif
@@ -108,6 +110,14 @@ nxt_http_static_init(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
 
         conf->shares[i].var = var;
         conf->shares[i].is_const = nxt_var_is_const(var);
+    }
+
+    if (acf->index == NULL) {
+        nxt_str_set(&conf->index, "index.html");
+
+    } else {
+        nxt_conf_get_string(acf->index, &str);
+        nxt_str_dup(mp, &conf->index, &str);
     }
 
 #if (NXT_HAVE_OPENAT2)
@@ -222,8 +232,10 @@ nxt_http_static_iterate(nxt_task_t *task, nxt_http_request_t *r,
 
 #if (NXT_DEBUG)
     nxt_str_t  shr;
+    nxt_str_t  idx;
 
     nxt_var_raw(share->var, &shr);
+    idx = conf->index;
 
 #if (NXT_HAVE_OPENAT2)
     nxt_str_t  chr;
@@ -235,11 +247,14 @@ nxt_http_static_iterate(nxt_task_t *task, nxt_http_request_t *r,
         nxt_str_set(&chr, "");
     }
 
-    nxt_debug(task, "http static: \"%V\" (chroot: \"%V\")", &shr, &chr);
+    nxt_debug(task, "http static: \"%V\", index: \"%V\" (chroot: \"%V\")",
+              &shr, &idx, &chr);
 #else
-    nxt_debug(task, "http static: \"%V\"", &shr);
+    nxt_debug(task, "http static: \"%V\", index: \"%V\"", &shr, &idx);
 #endif
 #endif /* NXT_DEBUG */
+
+    ctx->index = conf->index;
 
     if (share->is_const) {
         nxt_var_raw(share->var, &ctx->share);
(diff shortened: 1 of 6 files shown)


src/nxt_http_static.c
```diff
@@ -571,7 +629,9 @@ nxt_http_static_send_ready(nxt_task_t *task, void *obj, void *data)
         /* Not a file. */
         nxt_file_close(task, f);
 
-        if (nxt_slow_path(!nxt_is_dir(&fi))) {
+        if (nxt_slow_path(!nxt_is_dir(&fi)
+                          || shr->start[shr->length - 1] == '/'))
```
hongzhidao, line 633: If `shr` ends with `/`, it can return directly with 404 before opening file.

src/nxt_http_static.c
```diff
@@ -235,42 +279,58 @@ nxt_http_static_iterate(nxt_task_t *task, nxt_http_request_t *r,
...
-        if (nxt_slow_path(ret != NXT_OK)) {
-            nxt_http_request_error(task, r, NXT_HTTP_INTERNAL_SERVER_ERROR);
-            return;
-        }
-
         nxt_var_query(task, r->var_query, share->var, &ctx->share);
 
 #if (NXT_HAVE_OPENAT2)
-        if (conf->chroot != NULL && ctx->index == 0) {
+        if (conf->chroot != NULL && ctx->share_idx == 0) {
```
hongzhidao, line 320: There is an optimization here. For example:
```
{
     "chroot": "text$var",
     "share": [...]
}
```
The variable of chroot only resolves once no matter how many `share` has.

It seems that there should have similar optimization with an array-like index.
Anyway, consider this configuration:
```
{
     "chroot": "text$var",
     "share": [var1, var2, var3],
     "index": [var1, var2, var3, var4]
}
```

It's better not to resolve multiple times on the same variable.

src/nxt_http_static.c
alejandro-colomar: The rationale here is:

If `shr` ends with `/`, we use (append) `index`; otherwise we don't.  That's okay, and not a failure condition.

Then, independently of having used `index` or not, we open the file.

If it's a regular file, we proceed normally.
If it's not a regular file, but not a directory, we had this 404.
If it's a directory, we were redirecting (301) to `${r->path}/`.

Since the index was hardcoded to `"index.html"`, it was hard to get that to be a file (unless the server did something stupid such as `mdkir index.html` or maybe symlinked the index file to a dir), so that was enough.

Now that we have a configurable index, it's easier to have it be a directory: `"index": ""` or `"index": {}` will result in nothing appended to `shr`, and therefore, if `shr` ends with `/`, `r->path` will also end with `/`.

Since we are appending nothing, the problem is really deeper, because the redirection, if followed would just append `/` to the URI, which would result in a loop that ends up in something like `${r->path}//////`.

The condition to fail, independently of when we apply it, should be:

**if**
we have used `index` (the easiest thing to test this is "if `shr` ends with `/`)
**and**
 the file is a directory (this can be tested in two ways; see below).

The 2 ways we can test the second part of the condition are:
- by opening the file and checking that it's a directory.  This is 100% sure.
- by checking that either `index` is empty or it ends with `/`.  This is good enough, but it's not perfect, because it would not account for the case that `"index": "dir"`, which would then return 301 and a Location of `$shr/dir/` (which would be inconsistent with the rest).

The first way is the one I decided to implement because of:

It consistently returns 404 for directories after appending `index`.
Since having a directory _after_ appending index means that something is wrong in the server (because why would that happen in a sane server?), I don't want to optimize for that case at the cost of having some larger code that will make the common (and correct) case a bit slower.

alejandro-colomar: @tippexs @lcrilly @hongzhidao Ignoring performance, are there any security implications of choosing one of the above or the other?

hongzhidao: > It consistently returns 404 for directories after appending index.

I agree.

> If shr ends with /, it can directly enter nxt_http_static_next() before opening file.

Sorry that I didn't make it clear. Take a look again.

```
diff -r bb4e9ca76ef9 src/nxt_http_static.c
--- a/src/nxt_http_static.c     Fri May 13 19:46:08 2022 +0800
+++ b/src/nxt_http_static.c     Fri May 13 20:49:44 2022 +0800
@@ -347,6 +348,9 @@ nxt_http_static_send_ready(nxt_task_t *t
         fname = ctx->share.start;
     }

+    // if fname ends with "/\0",
+    // nxt_http_static_next(task, r, ctx, NXT_HTTP_NOT_FOUND);
+
     nxt_memzero(&file, sizeof(nxt_file_t));

     file.name = fname;
```

src/nxt_http_static.c
hongzhidao: Do you think if we must support array-like index? Is a single index good enough for users?

src/nxt_http_static.c
alejandro-colomar: Yes, I also considered that.  It's a mix of both ways.

The thing with this is that this optimizes for the cases that `"index": {}` or `"index": ""` or `"index": "dir/"`.
At the cost of adding 2 lines of code.

I consider all of those 3 cases to be broken.  Especially, the last two seem completely useless to me.  The first one may be useful to an automated configuration generator that may be adding or removing lines, and may end up with an empty array; but still I don't think we should optimize for that corner case, if it adds 2 LOC to the common case, do you?

src/nxt_http_static.c
alejandro-colomar: I don't know.  I use a single one myself.  We could maybe add support for string and variable, and hold array.  If someone ever requests it, we could merge the implementation we have, and if not, let's keep unit simpler.

I think it's rare to have a need for an array of indices.

alejandro-colomar: I'll remove the index-arr from this pull request, and keep it in a branch so that we have the code if we ever need it.

alejandro-colomar: Even a variable would be rarely needed for `index`, I would say.

alejandro-colomar: I'll post the optimization, so that you can check it.  Then I'll remove the array implementation from the PR.

hongzhidao:
> I removed the implementation of `index` as an array. As suggested by @hongzhidao, I'm not sure we need it. [#684 (comment)](https://github.com/nginx/unit/pull/684#discussion_r872368368)
> 
> I'll keep the code in a branch in my personal repository, in case we need it in the future.
> 
> If someone requests the feature in the future, we can always reconsider it; but let's keep it simple while we're unsure.

Makes sense.

src/nxt_http_static.c
```diff
@@ -235,42 +256,58 @@ nxt_http_static_iterate(nxt_task_t *task, nxt_http_request_t *r,
...
-    nxt_debug(task, "http static: \"%V\"", &shr);
+    nxt_debug(task, "http static: \"%V\", index: \"%V\"", &shr, &idx);
 #endif
 #endif /* NXT_DEBUG */
 
-    if (share->is_const) {
+    shr_is_const = share->is_const;
+    idx_is_const = nxt_var_is_const(conf->index);
+
+    if (!shr_is_const || !idx_is_const) {
```
hongzhidao, line 269: What if `chroot` is a variable string?
If you didn't run tests, I suggest running them first.

hongzhidao:
> I removed the implementation of `index` as an array. As suggested by @hongzhidao, I'm not sure we need it. [#684 (comment)](https://github.com/nginx/unit/pull/684#discussion_r872368368)
> 
> I'll keep the code in a branch in my personal repository, in case we need it in the future.
> 
> If someone requests the feature in the future, we can always reconsider it; but let's keep it simple while we're unsure.

Note that if there is only one single index, it should be resolved only once if it's a variable string.

src/nxt_http_static.c
alejandro-colomar: When I wrote that code, I didn't fully understand how `chroot` was being handled.  I had that same question to myself.

See the code without variables in index:

```c
$ grepc -tfd nxt_http_static_iterate
./src/nxt_http_static.c:220:
static void
nxt_http_static_iterate(nxt_task_t *task, nxt_http_request_t *r,
    nxt_http_static_ctx_t *ctx)
{
    ...

    if (share->is_const) {
        nxt_var_raw(share->var, &ctx->share);

#if (NXT_HAVE_OPENAT2)
        if (conf->chroot != NULL && ctx->share_idx == 0) {
            nxt_var_raw(conf->chroot, &ctx->chroot);
        }
#endif

        nxt_http_static_send_ready(task, r, ctx);

    } else {
            nxt_http_request_error(task, r, NXT_HTTP_INTERNAL_SERVER_ERROR);
            return;
        }

        nxt_var_query(task, r->var_query, share->var, &ctx->share);

#if (NXT_HAVE_OPENAT2)
        if (conf->chroot != NULL && ctx->share_idx == 0) {
            nxt_var_query(task, r->var_query, conf->chroot, &ctx->chroot);
        }
#endif

                              nxt_http_static_send_ready,
                              nxt_http_static_var_error);
     }
}
```

This is the code before adding variables to index.  It seemed to me that variables in `chroot` were only being handled correctly if `share` also contained variables.  I didn't want to touch code that I didn't understand, and so I left it as is (I think I did; and old pytests pass, so they confirm).

I now checked `test/test_static_chroot.py` and found no tests that set a variable in `chroot` while not using variables in `share`.  We either have tests that have variables in both, or only in `share`.

When I programmed this, I assumed that if `chroot` contained variables, `share` would too (that's what the code was telling me, AFAICR).

So, the question I have now is: is it possible to have variables in `chroot` while not using variables in `share`?  If so, we should probably open a new issue (as it's not completely related to this one; just a little bit), and also provide some tests for that case.

hongzhidao: >  I didn't fully understand how chroot was being handled. 

It's common for devs, I usually understand them by reading tests to know how to use them, then reading source code.

> It seemed to me that variables in chroot were only being handled correctly if share also contained variables. 

`chroot` option can be a const string or a variable string. Previous logic is we call `nxt_http_static_send_ready()` only both `share` and `chroot` are constant strings.

> I now checked test/test_static_chroot.py and found no tests that set a variable in chroot while not using variables in share. We either have tests that have variables in both, or only in share.

It's better to ask @andrey-zelenkov 

>  is it possible to have variables in chroot while not using variables in share

Both of them can be const string and variables.

alejandro-colomar: Okay; I opened a separate issue with that.  I think we should block this PR until we resolve that.

alejandro-colomar:
For the first patch set (`index-str`; constant index string), I tested the following config, which works both with and without  `--debug`:

```json
{
  "listeners": {
    "*:80": { "pass": "routes/idx-html"        },
    "*:81": { "pass": "routes/idx"             },
    "*:82": { "pass": "routes/empty"           },
    "*:83": { "pass": "routes/no-idx"          },
    "*:84": { "pass": "routes/dir"             },
    "*:85": { "pass": "routes/dir-slash"       },
    "*:86": { "pass": "routes/no-slash-share"  },
    "*:87": { "pass": "routes/share"           },
    "*:88": { "pass": "routes/uri"             }
  },

  "routes": {
    "idx-html": [{
      "action": {"share": "/home/alx/srv/www/", "index": "readme.html"}
    }],
    "idx": [{
      "action": {"share": "/home/alx/srv/www/", "index": "readme"}
    }],
    "empty": [{
      "action": {"share": "/home/alx/srv/www/", "index": ""}
    }],
    "no-idx": [{
      "action": {"share": "/home/alx/srv/www/"}
    }],
    "dir": [{
      "action": {"share": "/home/alx/srv/www/", "index": "dir"}
    }],
    "dir-slash": [{
      "action": {"share": "/home/alx/srv/www/", "index": "dir/"}
    }],
    "no-slash-share": [{
      "action": {"share": "/home/alx/srv/www", "index": "readme.html"}
    }],
    "share": [{
      "action": {"share": "/home/alx/srv/www/readme"}
    }],
    "uri": [{
      "action": {"share": "/home/alx/srv/www$uri", "index": "readme.html"}
    }]
  }
}
```

With the following results:

```
$ seq 80 88 | while read p; do url="localhost:$p"; echo "URL: $url"; curl localhost:$p; echo; done
URL: localhost:80
rmdh

URL: localhost:81
rdm

URL: localhost:82
<!DOCTYPE html><title>Error 404</title><p>Error 404.

URL: localhost:83
idx

URL: localhost:84
<!DOCTYPE html><title>Error 404</title><p>Error 404.

URL: localhost:85
<!DOCTYPE html><title>Error 404</title><p>Error 404.

URL: localhost:86

URL: localhost:87
rdm

URL: localhost:88
rmdh

```

alejandro-colomar:
I just **merged** the following:

```
(ret404) Static: returning 404 when "index" is a non-regular file.
(rename) Renamed nxt_http_static_ctx_t field 'index' to 'share_idx'.
```

alejandro-colomar:
Also merged:

```
(index-str) Tests: added tests for "index" (string) option.
            Static: supporting new "index" option.
```

Issue #674, opened 2022-04-09
Support empty "location" headers.


alejandro-colomar:
An empty string in Location is handled specially, by not sending a Location header.

This is a useless scenario, I think.  But it might happen, especially after variable resolution, so we need to consider it.

The obsolete RFC 2616 defined the Location header to consist of an absolute URI
<https://www.rfc-editor.org/rfc/rfc2616#section-14.30>, which cannot be an empty string.

However, the current RFC 7231 allows the Location to be a relative URI
<https://www.rfc-editor.org/rfc/rfc7231#section-7.1.2>,
and a relative URI might be the empty string
<https://stackoverflow.com/a/43338457>.

For that reason, I think we must allow an empty Location header, by changing the current behavior from sending no Location header to sending an empty one.  It will also help simplify the code, which will be more readable and (negligibly) faster, while being more conformant to the current RFC.  The only downside is that the code will not be optimized for that rare case, but that shouldn't be a problem.

This issue was first raised in <https://github.com/nginx/unit/pull/660>

Cc: @tippexs 
Cc: @hongzhidao

alejandro-colomar:
After fixing an issue with <https://github.com/nginx/unit/pull/660>, it's not clear that supporting empty Location headers will be better (nor worse).

We can add support, but it's not something that I would care too much for now.

tippexs:
As discussed today. Sounds like a good idea. Given the variable resolves to an empty string `""` the Location-Header will be `Location: ` which is fine in that case. Agree on that.

alejandro-colomar:
>  I would stage this a medium? Lmk if you would like to change that.

I'd say yes.  But I already have the code, so I'll assign it to myself :)

PR #686, opened 2022-04-28
Supporting empty Location URIs

alejandro-colomar:
https://github.com/nginx/unit/issues/674

Cc: @hongzhidao 
Cc: @tippexs

### DIFF
--- a/src/nxt_http_return.c
+++ b/src/nxt_http_return.c
@@ -11,12 +11,10 @@ typedef struct {
     nxt_http_status_t  status;
     nxt_var_t          *location;
     nxt_str_t          encoded;
-    uint8_t            loc_is_const;
 } nxt_http_return_conf_t;
 
 
 typedef struct {
-    nxt_http_action_t  *action;
     nxt_str_t          location;
     nxt_str_t          encoded;
 } nxt_http_return_ctx_t;
@@ -38,7 +36,6 @@ nxt_http_return_init(nxt_mp_t *mp, nxt_http_action_t *action,
     nxt_http_action_conf_t *acf)
 {
     nxt_str_t               str;
-    nxt_var_t               *var;
     nxt_http_return_conf_t  *conf;
 
     conf = nxt_mp_zget(mp, sizeof(nxt_http_return_conf_t));
@@ -51,20 +48,18 @@ nxt_http_return_init(nxt_mp_t *mp, nxt_http_action_t *action,
 
     conf->status = nxt_conf_get_number(acf->ret);
 
-    if (acf->location.length == 0) {
-        conf->loc_is_const = 1;
+    if (acf->location == NULL) {
         return NXT_OK;
     }
 
-    var = nxt_var_compile(&acf->location, mp, 0);
-    if (nxt_slow_path(var == NULL)) {
+    nxt_conf_get_string(acf->location, &str);
+
+    conf->location = nxt_var_compile(&str, mp, 0);
+    if (nxt_slow_path(conf->location == NULL)) {
         return NXT_ERROR;
     }
 
-    conf->location = var;
-    conf->loc_is_const = nxt_var_is_const(var);
-
-    if (conf->loc_is_const) {
+    if (nxt_var_is_const(conf->location)) {
         nxt_var_raw(conf->location, &str);
         return nxt_http_return_encode(mp, &conf->encoded, &str);
     }
@@ -100,17 +95,23 @@ nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
         return NULL;
     }
 
-    ctx = nxt_mp_zget(r->mem_pool, sizeof(nxt_http_return_ctx_t));
-    if (nxt_slow_path(ctx == NULL)) {
-        goto fail;
+    if (conf->location == NULL) {
+        ctx = NULL;
+
+    } else {
+        ctx = nxt_mp_zget(r->mem_pool, sizeof(nxt_http_return_ctx_t));
+        if (nxt_slow_path(ctx == NULL)) {
+            goto fail;
+        }
     }
 
-    ctx->action = action;
     r->status = conf->status;
     r->resp.content_length_n = 0;
 
-    if (conf->loc_is_const) {
-        ctx->encoded = conf->encoded;
+    if (ctx == NULL || nxt_var_is_const(conf->location)) {
+        if (ctx != NULL) {
+            ctx->encoded = conf->encoded;
+        }
 
         nxt_http_return_send_ready(task, r, ctx);
 
@@ -167,25 +168,21 @@ nxt_http_return_send_ready(nxt_task_t *task, void *obj, void *data)
 {
     nxt_int_t               ret;
     nxt_http_field_t        *field;
-    nxt_http_action_t       *action;
     nxt_http_request_t      *r;
     nxt_http_return_ctx_t   *ctx;
-    nxt_http_return_conf_t  *conf;
 
     r = obj;
     ctx = data;
-    action = ctx->action;
-    conf = action->u.conf;
 
-    if (!conf->loc_is_const) {
-        ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
-                                     &ctx->location);
-        if (nxt_slow_path(ret == NXT_ERROR)) {
-            goto fail;
+    if (ctx != NULL) {
+        if (ctx->location.length > 0) {
+            ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
+                                         &ctx->location);
+            if (nxt_slow_path(ret == NXT_ERROR)) {
+                goto fail;
+            }
         }
-    }
 
-    if (ctx->encoded.length > 0) {
         field = nxt_list_zero_add(r->resp.fields);
         if (nxt_slow_path(field == NULL)) {
             goto fail;
--- a/test/test_return.py
+++ b/test/test_return.py
@@ -178,6 +178,21 @@ Connection: close
             == "/#%23foo%3Fbar"
         ), 'location with a variable with encoding'
 
+        assert 'success' in self.conf(
+            '""', 'routes/0/action/location'
+        ), 'location empty'
+        assert self.get()['headers']['Location'] == ''
+
+        assert 'success' in self.conf(
+            '"${host}"', 'routes/0/action/location'
+        ), 'location empty with variable'
+        assert (
+            self.get(headers={"Host": "", "Connection": "close"})['headers'][
+                'Location'
+            ]
+            == ""
+        ), 'location with empty variable'
+
     def test_return_invalid(self):
         def check_error(conf):
             assert 'error' in self.conf(conf, 'routes/0/action')
--- a/docs/changes.xml
+++ b/docs/changes.xml
@@ -37,6 +37,12 @@ Ruby Rack environment parameter "SCRIPT_NAME" support.
 </para>
 </change>
 
+<change type="change">
+<para>
+supporting empty strings in the "location" option of the "return" action.
+</para>
+</change>
+
 <change type="feature">
 <para>
 variables support in the "location" option of the "return" action.
--- a/src/nxt_http.h
+++ b/src/nxt_http.h
@@ -206,7 +206,7 @@ typedef struct nxt_http_route_addr_rule_s  nxt_http_route_addr_rule_t;
 typedef struct {
     nxt_conf_value_t                *pass;
     nxt_conf_value_t                *ret;
-    nxt_str_t                       location;
+    nxt_conf_value_t                *location;
     nxt_conf_value_t                *proxy;
     nxt_conf_value_t                *share;
     nxt_str_t                       chroot;
--- a/src/nxt_http_route.c
+++ b/src/nxt_http_route.c
@@ -626,7 +626,7 @@ static nxt_conf_map_t  nxt_http_route_action_conf[] = {
     },
     {
         nxt_string("location"),
-        NXT_CONF_MAP_STR,
+        NXT_CONF_MAP_PTR,
         offsetof(nxt_http_action_conf_t, location)
     },
     {


test/test_return.py
```diff
@@ -178,6 +178,19 @@ def test_return_location_edit(self):
...
         ), 'location with a variable with encoding'
 
+        assert 'success' in self.conf(
+            '""', 'routes/0/action/location'
+        ), 'location empty'
+        assert self.get()['headers']['Location'] == ''
+
+        assert 'success' in self.conf(
+            '"${host}"', 'routes/0/action/location'
+        ), 'location empty with variable'
```
tippexs, line 188: shouldn't that be `location with empty variable` as well?

tippexs (changes requested):

test/test_return.py
alejandro-colomar: Both tests are testing the same thing, but I used slightly different names in case one of them fails it is easier to tell which one failed.

tippexs: All good :) Makes sense.

tippexs (approved):

src/nxt_http_return.c
```diff
@@ -114,6 +115,9 @@ nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
 
         nxt_http_return_send_ready(task, r, ctx);
 
+    } else if (conf->location == NULL) {
```
alejandro-colomar, line 118: I'm not convinced by this code.

Let's consider two of the ways this can be written:

A:
```c
    if (conf->loc_is_const) {
        ctx->encoded = conf->encoded;

        nxt_http_return_send_ready(task, r, ctx);

    } else if (conf->location == NULL) {
        nxt_http_return_send_ready(task, r, ctx);

    } else {
            goto fail;
        }

        nxt_var_query(task, r->var_query, conf->location, &ctx->location);

                              nxt_http_return_send_ready,
                              nxt_http_return_var_error);
    }
```
B:
```c
    if (conf->loc_is_const || conf->location == NULL) {
        if (conf->loc_is_const) {
            ctx->encoded = conf->encoded;
        }

        nxt_http_return_send_ready(task, r, ctx);

    } else {
            goto fail;
        }

        nxt_var_query(task, r->var_query, conf->location, &ctx->location);

                              nxt_http_return_send_ready,
                              nxt_http_return_var_error);
    }
```

A is flatter, with simpler conditions, but has a bit of repetition.
B has less branches, and doesn't have code repetition, but the conditionals are a bit more complex (and there's repetition in the conditionals).

Overall, I think I prefer A (as is written in this patch set), but none of them fully convinces me.

@hongzhidao, @VBart  which way would you write?

hongzhidao: @alejandro-colomar 
My suggestion is to try to get rid of `loc_is_const` first. I tried to simplify it based on your logic.
```
diff -r df887748fffb src/nxt_http_return.c
--- a/src/nxt_http_return.c     Wed Dec 15 12:18:59 2021 +0100
+++ b/src/nxt_http_return.c     Fri May 06 15:13:22 2022 +0800
@@ -11,12 +11,10 @@ typedef struct {
     nxt_http_status_t  status;
     nxt_var_t          *location;
     nxt_str_t          encoded;
-    uint8_t            loc_is_const;
 } nxt_http_return_conf_t;


 typedef struct {
-    nxt_http_action_t  *action;
     nxt_str_t          location;
     nxt_str_t          encoded;
 } nxt_http_return_ctx_t;
@@ -38,7 +36,6 @@ nxt_http_return_init(nxt_mp_t *mp, nxt_h
     nxt_http_action_conf_t *acf)
 {
     nxt_str_t               str;
-    nxt_var_t               *var;
     nxt_http_return_conf_t  *conf;

     conf = nxt_mp_zget(mp, sizeof(nxt_http_return_conf
_t));
@@ -51,22 +48,16 @@ nxt_http_return_init(nxt_mp_t *mp,
nxt_h

     conf->status = nxt_conf_get_number(acf->ret);

-    if (acf->location.length == 0) {
-        conf->loc_is_const = 1;
-        return NXT_OK;
-    }
+    if (acf->location.length > 0) {
+        conf->location = nxt_var_compile(&acf->locatio
n, mp, 0);
+        if (nxt_slow_path(conf->location == NULL)) {
+        }

-    var = nxt_var_compile(&acf->location, mp, 0);
-    if (nxt_slow_path(var == NULL)) {
-        return NXT_ERROR;
-    }
-
-    conf->location = var;
-    conf->loc_is_const = nxt_var_is_const(var);
-
-    if (conf->loc_is_const) {
-        nxt_var_raw(conf->location, &str);
-        return nxt_http_return_encode(mp, &conf->encod
ed, &str);
+        if (nxt_var_is_const(conf->location)) {
+            nxt_var_raw(conf->location, &str);
+            return nxt_http_return_encode(mp, &conf->e
ncoded, &str);
+        }
     }

     return NXT_OK;
@@ -105,16 +96,14 @@ nxt_http_return(nxt_task_t *task,
nxt_ht
         goto fail;
     }

-    ctx->action = action;
     r->status = conf->status;
     r->resp.content_length_n = 0;

-    if (conf->loc_is_const) {
+    if (conf->location == NULL || conf->encoded.length
 > 0) {
         ctx->encoded = conf->encoded;
-
         nxt_http_return_send_ready(task, r, ctx);

-    } else {
+    } else if (conf->location != NULL) {
         ret = nxt_var_query_init(&r->var_query, r, r->
mem_pool);
             goto fail;
@@ -165,19 +154,15 @@ nxt_http_return_encode(nxt_mp_t *
mp, nxt
 static void
 nxt_http_return_send_ready(nxt_task_t *task, void *obj
, void *data)
 {
-    nxt_int_t               ret;
-    nxt_http_field_t        *field;
-    nxt_http_action_t       *action;
-    nxt_http_request_t      *r;
-    nxt_http_return_ctx_t   *ctx;
-    nxt_http_return_conf_t  *conf;
+    nxt_int_t              ret;
+    nxt_http_field_t       *field;
+    nxt_http_request_t     *r;
+    nxt_http_return_ctx_t  *ctx;

     r = obj;
     ctx = data;
-    action = ctx->action;
-    conf = action->u.conf;

-    if (!conf->loc_is_const) {
+    if (ctx->location.length > 0) {
         ret = nxt_http_return_encode(r->mem_pool, &ctx
->encoded,
                                      &ctx->location);
```

If it works, then add the empty location.

alejandro-colomar: I'll also post here a version of your patch, applied on top of my changes, to easier compare them:

```diff
diff --git a/src/nxt_http_return.c b/src/nxt_http_return.c
index 37d78e8..f615d10 100644
--- a/src/nxt_http_return.c
+++ b/src/nxt_http_return.c
@@ -11,12 +11,10 @@ typedef struct {
     nxt_http_status_t  status;
     nxt_var_t          *location;
     nxt_str_t          encoded;
-    uint8_t            loc_is_const;
 } nxt_http_return_conf_t;
 
 
 typedef struct {
-    nxt_http_action_t  *action;
     nxt_str_t          location;
     nxt_str_t          encoded;
 } nxt_http_return_ctx_t;
@@ -38,7 +36,6 @@ nxt_http_return_init(nxt_mp_t *mp, nxt_http_action_t *action,
     nxt_http_action_conf_t *acf)
 {
     nxt_str_t               str;
-    nxt_var_t               *var;
     nxt_http_return_conf_t  *conf;
 
     conf = nxt_mp_zget(mp, sizeof(nxt_http_return_conf_t));
@@ -51,23 +48,16 @@ nxt_http_return_init(nxt_mp_t *mp, nxt_http_action_t *action,
 
     conf->status = nxt_conf_get_number(acf->ret);
 
-    if (acf->location == NULL) {
-        return NXT_OK;
-    }
-
-    nxt_conf_get_string(acf->location, &str);
-
-    var = nxt_var_compile(&str, mp, 0);
-    if (nxt_slow_path(var == NULL)) {
-        return NXT_ERROR;
-    }
-
-    conf->location = var;
-    conf->loc_is_const = nxt_var_is_const(var);
+    if (acf->location.length > 0) {
+        conf->location = nxt_var_compile(&acf->location, mp, 0);
+        if (nxt_slow_path(conf->location == NULL)) {
+            return NXT_ERROR;
+        }
 
-    if (conf->loc_is_const) {
-        nxt_var_raw(conf->location, &str);
-        return nxt_http_return_encode(mp, &conf->encoded, &str);
+        if (nxt_var_is_const(conf->location)) {
+            nxt_var_raw(conf->location, &str);
+            return nxt_http_return_encode(mp, &conf->encoded, &str);
+        }
     }
 
     return NXT_OK;
@@ -106,19 +96,14 @@ nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
         goto fail;
     }
 
-    ctx->action = action;
     r->status = conf->status;
     r->resp.content_length_n = 0;
 
-    if (conf->loc_is_const) {
+    if (conf->location == NULL || conf->encoded.length > 0) {
         ctx->encoded = conf->encoded;
-
         nxt_http_return_send_ready(task, r, ctx);
 
-    } else if (conf->location == NULL) {
-        nxt_http_return_send_ready(task, r, ctx);
-
-    } else {
+    } else if (conf->location != NULL) {
         ret = nxt_var_query_init(&r->var_query, r, r->mem_pool);
         if (nxt_slow_path(ret != NXT_OK)) {
             goto fail;
@@ -169,28 +154,23 @@ nxt_http_return_encode(nxt_mp_t *mp, nxt_str_t *encoded,
 static void
 nxt_http_return_send_ready(nxt_task_t *task, void *obj, void *data)
 {
-    nxt_int_t               ret;
-    nxt_http_field_t        *field;
-    nxt_http_action_t       *action;
-    nxt_http_request_t      *r;
-    nxt_http_return_ctx_t   *ctx;
-    nxt_http_return_conf_t  *conf;
+    nxt_int_t              ret;
+    nxt_http_field_t       *field;
+    nxt_http_request_t     *r;
+    nxt_http_return_ctx_t  *ctx;
 
     r = obj;
     ctx = data;
-    action = ctx->action;
-    conf = action->u.conf;
-
-    if (conf->location != NULL) {
 
-        if (!conf->loc_is_const) {
-            ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
-                                         &ctx->location);
-            if (nxt_slow_path(ret == NXT_ERROR)) {
-                goto fail;
-            }
+    if (ctx->location.length > 0) {
+        ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
+                                     &ctx->location);
+        if (nxt_slow_path(ret == NXT_ERROR)) {
+            goto fail;
         }
+    }
 
+    if (ctx->encoded.length > 0) {
         field = nxt_list_zero_add(r->resp.fields);
         if (nxt_slow_path(field == NULL)) {
             goto fail;
```

alejandro-colomar: > I'll also post here a version of your patch, applied on top of my changes, to easier compare them:
> 
> ```diff
> diff --git a/src/nxt_http_return.c b/src/nxt_http_return.c
> index 37d78e8..f615d10 100644
> --- a/src/nxt_http_return.c
> +++ b/src/nxt_http_return.c
> @@ -11,12 +11,10 @@ typedef struct {
>      nxt_http_status_t  status;
>      nxt_var_t          *location;
>      nxt_str_t          encoded;
> -    uint8_t            loc_is_const;
>  } nxt_http_return_conf_t;
>  
>  
>  typedef struct {
> -    nxt_http_action_t  *action;
>      nxt_str_t          location;
>      nxt_str_t          encoded;
>  } nxt_http_return_ctx_t;
> @@ -38,7 +36,6 @@ nxt_http_return_init(nxt_mp_t *mp, nxt_http_action_t *action,
>      nxt_http_action_conf_t *acf)
>  {
>      nxt_str_t               str;
> -    nxt_var_t               *var;
>      nxt_http_return_conf_t  *conf;
>  
>      conf = nxt_mp_zget(mp, sizeof(nxt_http_return_conf_t));
> @@ -51,23 +48,16 @@ nxt_http_return_init(nxt_mp_t *mp, nxt_http_action_t *action,
>  
>      conf->status = nxt_conf_get_number(acf->ret);
>  
> -    if (acf->location == NULL) {
> -        return NXT_OK;
> -    }
> -
> -    nxt_conf_get_string(acf->location, &str);
> -
> -    var = nxt_var_compile(&str, mp, 0);
> -    if (nxt_slow_path(var == NULL)) {
> -        return NXT_ERROR;
> -    }
> -
> -    conf->location = var;
> -    conf->loc_is_const = nxt_var_is_const(var);
> +    if (acf->location.length > 0) {
> +        conf->location = nxt_var_compile(&acf->location, mp, 0);
> +        if (nxt_slow_path(conf->location == NULL)) {
> +            return NXT_ERROR;
> +        }
>  
> -    if (conf->loc_is_const) {
> -        nxt_var_raw(conf->location, &str);
> -        return nxt_http_return_encode(mp, &conf->encoded, &str);
> +        if (nxt_var_is_const(conf->location)) {
> +            nxt_var_raw(conf->location, &str);
> +            return nxt_http_return_encode(mp, &conf->encoded, &str);
> +        }
>      }
>  
>      return NXT_OK;
> @@ -106,19 +96,14 @@ nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
>          goto fail;
>      }
>  
> -    ctx->action = action;
>      r->status = conf->status;
>      r->resp.content_length_n = 0;
>  
> -    if (conf->loc_is_const) {
> +    if (conf->location == NULL || conf->encoded.length > 0) {
>          ctx->encoded = conf->encoded;
> -
>          nxt_http_return_send_ready(task, r, ctx);
>  
> -    } else if (conf->location == NULL) {
> -        nxt_http_return_send_ready(task, r, ctx);
> -
> -    } else {
> +    } else if (conf->location != NULL) {
>          ret = nxt_var_query_init(&r->var_query, r, r->mem_pool);
>          if (nxt_slow_path(ret != NXT_OK)) {
>              goto fail;
> @@ -169,28 +154,23 @@ nxt_http_return_encode(nxt_mp_t *mp, nxt_str_t *encoded,
>  static void
>  nxt_http_return_send_ready(nxt_task_t *task, void *obj, void *data)
>  {
> -    nxt_int_t               ret;
> -    nxt_http_field_t        *field;
> -    nxt_http_action_t       *action;
> -    nxt_http_request_t      *r;
> -    nxt_http_return_ctx_t   *ctx;
> -    nxt_http_return_conf_t  *conf;
> +    nxt_int_t              ret;
> +    nxt_http_field_t       *field;
> +    nxt_http_request_t     *r;
> +    nxt_http_return_ctx_t  *ctx;
>  
>      r = obj;
>      ctx = data;
> -    action = ctx->action;
> -    conf = action->u.conf;
> -
> -    if (conf->location != NULL) {
>  
> -        if (!conf->loc_is_const) {
> -            ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
> -                                         &ctx->location);
> -            if (nxt_slow_path(ret == NXT_ERROR)) {
> -                goto fail;
> -            }
> +    if (ctx->location.length > 0) {
> +        ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
> +                                     &ctx->location);
> +        if (nxt_slow_path(ret == NXT_ERROR)) {
> +            goto fail;
>          }
> +    }
>  
> +    if (ctx->encoded.length > 0) {

No, this won't work, I think.  This doesn't allow me to differentiate between

```json
"action": {
    "return": 301
}
```

and

```json
"action": {
    "return": 301,
    "location": "$host"
}
```

where `$host` is empty.

>          field = nxt_list_zero_add(r->resp.fields);
>          if (nxt_slow_path(field == NULL)) {
>              goto fail;
> ```

alejandro-colomar: I'll prepare an alternative patch, where we don't support empty Location headers, and I'll apply some of your simplification, to compare.  We might prefer that.

hongzhidao: Btw, actually, the `ctx` can be processed on-demand. Take the below as an example:
```
{
     "return": 200
}
```
In this case, we don't need to allocate `ctx` for reqeusts.
```
    ctx = nxt_mp_zget(r->mem_pool, sizeof(nxt_http_return_ctx_t));
    if (nxt_slow_path(ctx == NULL)) {
        goto fail;
    }
```

alejandro-colomar: Hmmm, good catch.  It's not so simple, because of how `nxt_http_return_send_ready()` receives its parameters.  But we can define two similar functions, one that accepts a conf object for simple cases, and one that receives a ctx object in complex cases.  I'm not sure if that will be better, but I'll write it so that we can evaluate it.

alejandro-colomar: The optimization makes sense to me:

```diff
diff --git a/src/nxt_http_return.c b/src/nxt_http_return.c
index 37d78e8..d349125 100644
--- a/src/nxt_http_return.c
+++ b/src/nxt_http_return.c
@@ -16,7 +16,6 @@ typedef struct {
 
 
 typedef struct {
-    nxt_http_action_t  *action;
     nxt_str_t          location;
     nxt_str_t          encoded;
 } nxt_http_return_ctx_t;
@@ -101,24 +100,20 @@ nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
         return NULL;
     }
 
-    ctx = nxt_mp_zget(r->mem_pool, sizeof(nxt_http_return_ctx_t));
-    if (nxt_slow_path(ctx == NULL)) {
-        goto fail;
-    }
-
-    ctx->action = action;
     r->status = conf->status;
     r->resp.content_length_n = 0;
 
-    if (conf->loc_is_const) {
-        ctx->encoded = conf->encoded;
+    if (conf->loc_is_const || conf->location == NULL) {
+        nxt_http_return_send_ready_simple(task, r, conf);
 
-        nxt_http_return_send_ready(task, r, ctx);
+    } else {
+        ctx = nxt_mp_zget(r->mem_pool, sizeof(nxt_http_return_ctx_t));
+        if (nxt_slow_path(ctx == NULL)) {
+            goto fail;
+        }
 
-    } else if (conf->location == NULL) {
-        nxt_http_return_send_ready(task, r, ctx);
+        ctx->action = action;
 
-    } else {
         ret = nxt_var_query_init(&r->var_query, r, r->mem_pool);
         if (nxt_slow_path(ret != NXT_OK)) {
             goto fail;
@@ -167,29 +162,18 @@ nxt_http_return_encode(nxt_mp_t *mp, nxt_str_t *encoded,
 
 
 static void
-nxt_http_return_send_ready(nxt_task_t *task, void *obj, void *data)
+nxt_http_return_send_ready_simple(nxt_task_t *task, void *obj, void *data)
 {
     nxt_int_t               ret;
     nxt_http_field_t        *field;
-    nxt_http_action_t       *action;
     nxt_http_request_t      *r;
-    nxt_http_return_ctx_t   *ctx;
     nxt_http_return_conf_t  *conf;
 
     r = obj;
-    ctx = data;
-    action = ctx->action;
-    conf = action->u.conf;
 
-    if (conf->location != NULL) {
+    conf = data;
 
-        if (!conf->loc_is_const) {
-            ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
-                                         &ctx->location);
-            if (nxt_slow_path(ret == NXT_ERROR)) {
-                goto fail;
-            }
-        }
+    if (conf->location != NULL) {
 
         field = nxt_list_zero_add(r->resp.fields);
         if (nxt_slow_path(field == NULL)) {
@@ -198,10 +182,51 @@ nxt_http_return_send_ready(nxt_task_t *task, void *obj, void *data)
 
         nxt_http_field_name_set(field, "Location");
 
-        field->value = ctx->encoded.start;
-        field->value_length = ctx->encoded.length;
+        field->value = conf->encoded.start;
+        field->value_length = conf->encoded.length;
+    }
+
+    r->state = &nxt_http_return_send_state;
+
+    nxt_http_request_header_send(task, r, NULL, NULL);
+
+    return;
+
+fail:
+
+    nxt_http_request_error(task, r, NXT_HTTP_INTERNAL_SERVER_ERROR);
+}
+
+
+static void
+nxt_http_return_send_ready(nxt_task_t *task, void *obj, void *data)
+{
+    nxt_int_t               ret;
+    nxt_http_field_t        *field;
+    nxt_http_action_t       *action;
+    nxt_http_request_t      *r;
+    nxt_http_return_ctx_t   *ctx;
+
+    r = obj;
+
+    ctx = data;
+    action = ctx->action;
+
+    ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded, &ctx->location);
+    if (nxt_slow_path(ret == NXT_ERROR)) {
+        goto fail;
+    }
+
+    field = nxt_list_zero_add(r->resp.fields);
+    if (nxt_slow_path(field == NULL)) {
+       goto fail;
     }
 
+    nxt_http_field_name_set(field, "Location");
+
+    field->value = ctx->encoded.start;
+    field->value_length = ctx->encoded.length;
+
     r->state = &nxt_http_return_send_state;
 
     nxt_http_request_header_send(task, r, NULL, NULL);
```

alejandro-colomar: After applying the above patch optimizing ctx, I can also apply this patch, which removes loc_is_const:

```diff
diff --git a/src/nxt_http_return.c b/src/nxt_http_return.c
index 5155859..92ee7de 100644
--- a/src/nxt_http_return.c
+++ b/src/nxt_http_return.c
@@ -11,7 +11,6 @@ typedef struct {
     nxt_http_status_t  status;
     nxt_var_t          *location;
     nxt_str_t          encoded;
-    uint8_t            loc_is_const;
 } nxt_http_return_conf_t;
 
 
@@ -37,7 +36,6 @@ nxt_http_return_init(nxt_mp_t *mp, nxt_http_action_t *action,
     nxt_http_action_conf_t *acf)
 {
     nxt_str_t               str;
-    nxt_var_t               *var;
     nxt_http_return_conf_t  *conf;
 
     conf = nxt_mp_zget(mp, sizeof(nxt_http_return_conf_t));
@@ -56,15 +54,12 @@ nxt_http_return_init(nxt_mp_t *mp, nxt_http_action_t *action,
 
     nxt_conf_get_string(acf->location, &str);
 
-    var = nxt_var_compile(&str, mp, 0);
-    if (nxt_slow_path(var == NULL)) {
+    conf->location = nxt_var_compile(&str, mp, 0);
+    if (nxt_slow_path(conf->location == NULL)) {
         return NXT_ERROR;
     }
 
-    conf->location = var;
-    conf->loc_is_const = nxt_var_is_const(var);
-
-    if (conf->loc_is_const) {
+    if (nxt_var_is_const(conf->location)) {
         nxt_var_raw(conf->location, &str);
         return nxt_http_return_encode(mp, &conf->encoded, &str);
     }
@@ -103,7 +98,7 @@ nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
     r->status = conf->status;
     r->resp.content_length_n = 0;
 
-    if (conf->loc_is_const || conf->location == NULL) {
+    if (conf->location == NULL || nxt_var_is_const(conf->location)) {
         nxt_http_return_send_ready_simple(task, r, conf);
 
     } else {
```

alejandro-colomar: After testing, I'll update the PR with those patches applied.

hongzhidao: > But we can define two similar functions, one that accepts a conf object for simple cases, and one that receives a ctx object in complex cases. 

You can reuse `nxt_http_return_send_ready(..., NULL)` for the simple case.
```
nxt_http_return_send_ready(..., ctx)
{
     if (ctx == NULL) {
           // this is the simple case.
     }
     ...
}
```

alejandro-colomar: No, we cant, because we would need both conf and ctx in the same function, but we only have one data parameter.
We had conf inside ctx, but if we don't allocate ctx we can't do that:

```c
static void
nxt_http_return_send_ready(nxt_task_t *task, void *obj, void *data)
{
    ...
    nxt_http_action_t       *action;
    nxt_http_request_t      *r;
    nxt_http_return_ctx_t   *ctx;
    nxt_http_return_conf_t  *conf;

    r = obj;
    ctx = data;
    action = ctx->action;
    conf = action->u.conf;

    ...
}

```

hongzhidao: It seems questions are based on if we plan to get rid of `loc_is_const`.

```
    conf = action->u.conf;

    if (!conf->loc_is_const) {
        ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
                                     &ctx->location);
            goto fail;
        }
    }
``

> Now the question is: is it worth it to have the extra ..._simple() function to be able to skip the allocation of ctx?

I prefer not to introduce *_simple() anyway. In the future we may support outputing content by "return.body" option.
It also need to support variables, so it's better to only introduce *_send_ready.

Comment (alejandro-colomar) [reply src/nxt_http_return.c:118]: > It seems questions are based on if we plan to get rid of `loc_is_const`.

Not only that.  I can try to get rid of `loc_is_const`.  But if we get rid of ctx, we need to use `conf->location`, and that's a problem because we don't have `conf`.

> > Now the question is: is it worth it to have the extra ..._simple() function to be able to skip the allocation of ctx?
> 
> I prefer not to introduce *_simple() anyway. In the future we may support outputing content by "return.body" option. It also need to support variables, so it's better to only introduce *_send_ready.

Makes sense to me.

Comment (alejandro-colomar) [reply src/nxt_http_return.c:118]: Okay, this is what we can get:

```diff
diff --git a/src/nxt_http_return.c b/src/nxt_http_return.c
index 37d78e8..b6c5672 100644
--- a/src/nxt_http_return.c
+++ b/src/nxt_http_return.c
@@ -11,7 +11,6 @@ typedef struct {
     nxt_http_status_t  status;
     nxt_var_t          *location;
     nxt_str_t          encoded;
-    uint8_t            loc_is_const;
 } nxt_http_return_conf_t;
 
 
@@ -57,15 +56,12 @@ nxt_http_return_init(nxt_mp_t *mp, nxt_http_action_t *action,
 
     nxt_conf_get_string(acf->location, &str);
 
-    var = nxt_var_compile(&str, mp, 0);
-    if (nxt_slow_path(var == NULL)) {
+    conf->location = nxt_var_compile(&str, mp, 0);
+    if (nxt_slow_path(conf->location == NULL)) {
         return NXT_ERROR;
     }
 
-    conf->location = var;
-    conf->loc_is_const = nxt_var_is_const(var);
-
-    if (conf->loc_is_const) {
+    if (nxt_var_is_const(conf->location)) {
         nxt_var_raw(conf->location, &str);
         return nxt_http_return_encode(mp, &conf->encoded, &str);
     }
@@ -110,14 +106,11 @@ nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
     r->status = conf->status;
     r->resp.content_length_n = 0;
 
-    if (conf->loc_is_const) {
+    if (conf->location == NULL || nxt_var_is_const(conf->location)) {
         ctx->encoded = conf->encoded;
 
         nxt_http_return_send_ready(task, r, ctx);
 
-    } else if (conf->location == NULL) {
-        nxt_http_return_send_ready(task, r, ctx);
-
     } else {
         ret = nxt_var_query_init(&r->var_query, r, r->mem_pool);
         if (nxt_slow_path(ret != NXT_OK)) {
@@ -183,7 +176,7 @@ nxt_http_return_send_ready(nxt_task_t *task, void *obj, void *data)
 
     if (conf->location != NULL) {
 
-        if (!conf->loc_is_const) {
+        if (!nxt_var_is_const(conf->location)) {
             ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
                                          &ctx->location);
             if (nxt_slow_path(ret == NXT_ERROR)) {
```

hongzhidao: Ok, I'll review it after they are updated.

alejandro-colomar: Done.

hongzhidao: Looks good.
And it seems you can do the optimization of on-demand ctx.

```
    action = ctx->action;
    conf = action->u.conf;

    if (conf->location != NULL) {

        if (!nxt_var_is_const(conf->location)) {
            ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
                                         &ctx->location);
                goto fail;
            }
        }
    }
```

TO

```
    ctx = data;

    if (ctx != NULL) {
         // it means location is set.
        action = ctx->action;
        conf = action->u.conf;
        //  conf->location is always not null.
    }
```

src/nxt_http_return.c
```diff
@@ -177,15 +174,16 @@ nxt_http_return_send_ready(nxt_task_t *task, void *obj, void *data)
...
     conf = action->u.conf;
 
-    if (!conf->loc_is_const) {
-        ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
-                                     &ctx->location);
-        if (nxt_slow_path(ret == NXT_ERROR)) {
-            goto fail;
+    if (conf->location != NULL) {
+
+        if (!nxt_var_is_const(conf->location)) {
```
hongzhidao, line 179: Since If it's const, the `ctx.location` is empty.
encode only ctx.location is not empty.
```
if (ctx->location.length > 0) {
            ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
                                         &ctx->location);
                goto fail;
            }
        }
```

src/nxt_http_return.c
```diff
@@ -51,20 +50,18 @@ nxt_http_return_init(nxt_mp_t *mp, nxt_http_action_t *action,
...
+    conf->location = nxt_var_compile(&str, mp, 0);
+    if (nxt_slow_path(conf->location == NULL)) {
         return NXT_ERROR;
     }
 
-    conf->location = var;
-    conf->loc_is_const = nxt_var_is_const(var);
-
-    if (conf->loc_is_const) {
+    if (nxt_var_is_const(conf->location)) {
```
hongzhidao, line 64: Tiny optimization:
```
    if (nxt_var_is_const(conf->location)) {
        nxt_var_raw(conf->location, &str);
        if (str.length > 0) {
            return nxt_http_return_encode(mp, &conf->encoded, &str);
       }
    }

return NXT_OK;
```

alejandro-colomar: That would optimize for the following case:

```json
"action": {
    "return": 301,
    "location": "$host"
}
```

at the cost of adding a conditional (an operation or so) in the common case.

Since that case is quite useless, I prefer keeping it simple, and optimize for the common case.  Anyway, `nxt_http_return_encode()` will return quite fast when the string is empty.

src/nxt_http_return.c
alejandro-colomar: Ahh, great catch!

That implicitly optimizes for the following case at no cost:

```json
"action": {
    "return": 301,
    "location": "$arg_foo"
}
```

src/nxt_http_return.c
alejandro-colomar: Yeah!!

Now we can optimize it :)

```diff
diff --git a/src/nxt_http_return.c b/src/nxt_http_return.c
index 09c6cef..e3dd02a 100644
--- a/src/nxt_http_return.c
+++ b/src/nxt_http_return.c
@@ -15,7 +15,6 @@ typedef struct {
 
 
 typedef struct {
-    nxt_http_action_t  *action;
     nxt_str_t          location;
     nxt_str_t          encoded;
 } nxt_http_return_ctx_t;
@@ -37,7 +36,6 @@ nxt_http_return_init(nxt_mp_t *mp, nxt_http_action_t *action,
     nxt_http_action_conf_t *acf)
 {
     nxt_str_t               str;
-    nxt_var_t               *var;
     nxt_http_return_conf_t  *conf;
 
     conf = nxt_mp_zget(mp, sizeof(nxt_http_return_conf_t));
@@ -97,17 +95,23 @@ nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
         return NULL;
     }
 
-    ctx = nxt_mp_zget(r->mem_pool, sizeof(nxt_http_return_ctx_t));
-    if (nxt_slow_path(ctx == NULL)) {
-        goto fail;
+    if (conf->location == NULL) {
+        ctx = NULL;
+
+    } else {
+        ctx = nxt_mp_zget(r->mem_pool, sizeof(nxt_http_return_ctx_t));
+        if (nxt_slow_path(ctx == NULL)) {
+            goto fail;
+        }
     }
 
-    ctx->action = action;
     r->status = conf->status;
     r->resp.content_length_n = 0;
 
-    if (conf->location == NULL || nxt_var_is_const(conf->location)) {
-        ctx->encoded = conf->encoded;
+    if (ctx == NULL || nxt_var_is_const(conf->location)) {
+        if (ctx != NULL) {
+            ctx->encoded = conf->encoded;
+        }
 
         nxt_http_return_send_ready(task, r, ctx);
 
@@ -164,18 +168,13 @@ nxt_http_return_send_ready(nxt_task_t *task, void *obj, void *data)
 {
     nxt_int_t               ret;
     nxt_http_field_t        *field;
-    nxt_http_action_t       *action;
     nxt_http_request_t      *r;
     nxt_http_return_ctx_t   *ctx;
-    nxt_http_return_conf_t  *conf;
 
     r = obj;
     ctx = data;
-    action = ctx->action;
-    conf = action->u.conf;
-
-    if (conf->location != NULL) {
 
+    if (ctx != NULL) {
         if (ctx->location.length > 0) {
             ret = nxt_http_return_encode(r->mem_pool, &ctx->encoded,
                                          &ctx->location);
```

src/nxt_http_return.c
hongzhidao: Yeah, we don't need to do it.

```
static nxt_int_t
nxt_http_return_encode(nxt_mp_t *mp, nxt_str_t *encoded,
    const nxt_str_t *location)
{
    nxt_uint_t  encode;

    if (nxt_is_complex_uri_encoded(location->start, location->length)) {
        // it contains the case of empty string.
        *encoded = *location;

        return NXT_OK;
    }
```

hongzhidao (approved):

PR #690, opened 2022-05-04
Supporting UNIX sockets in python ASGI.

alejandro-colomar:
This PR depends on <https://github.com/nginx/unit/pull/655>.

Cc: @echolimazulu


tippexs (approved):

tippexs (approved):
Looks good to me. We should check with the Python Team about `! Not sure if we need to Py_INCREF(Py_None).` and do a refactoring if neede. With `Py_INCREF(Py_None)` we alligned with the scopes defintion.

tippexs:
Implemenation details https://asgi.readthedocs.io/en/latest/specs/www.html
https://docs.python.org/3/c-api/none.html

PR #693, opened 2022-05-09
Support URI $ sub-delimiters

ac000:
This pull request adds support for $ sub-delimiters, or more generally it allows to specify a _$_ in URIs that would otherwise clash with variables.

All the tests under the test/ directory pass, well except for a couple of chroot ones which also fail under master.

It comprises three commits

```
Andrew Clayton (3):
      Support URI $ sub-delimiters.
      Tests: Remove some no longer invalid tests.
      Tests: Add tests for $ sub-delimiters.
```

#### Support URI $ sub-delimiters.

This allows to specify things like

```json
"location": "https://$host/path/$$sub1$$sub2"
```
in the Unit config which would translate to something like

```
https://example.com/path/$sub1$sub2
```

#### Tests: Remove some no longer invalid tests.

This removes some tests that were testing invalid paths that were using $$, which of course if now valid.

#### Tests: Add tests for $ sub-delimiters.

This adds some new tests for using $$ in URIs

Summary of changes

```
 docs/changes.xml              |  6 ++++++
 src/nxt_http_variables.c      | 16 ++++++++++++++++
 src/nxt_var.c                 |  8 ++++++++
 test/test_return.py           | 15 +++++++++++++++
 test/test_static_chroot.py    |  3 ---
 test/test_static_variables.py |  4 ----
 6 files changed, 45 insertions(+), 7 deletions(-)
```

### DIFF
--- a/docs/changes.xml
+++ b/docs/changes.xml
@@ -45,6 +45,12 @@ supporting empty strings in the "location" option of the "return" action.
 
 <change type="feature">
 <para>
+translate '$$' to a literal '$' during variable substitution.
+</para>
+</change>
+
+<change type="feature">
+<para>
 variables support in the "location" option of the "return" action.
 </para>
 </change>
--- a/src/nxt_http_variables.c
+++ b/src/nxt_http_variables.c
@@ -13,6 +13,8 @@ static nxt_int_t nxt_http_var_uri(nxt_task_t *task, nxt_var_query_t *query,
     nxt_str_t *str, void *ctx);
 static nxt_int_t nxt_http_var_host(nxt_task_t *task, nxt_var_query_t *query,
     nxt_str_t *str, void *ctx);
+static nxt_int_t nxt_http_var_dollar(nxt_task_t *task, nxt_var_query_t *query,
+    nxt_str_t *str, void *ctx);
 
 
 static nxt_var_decl_t  nxt_http_vars[] = {
@@ -26,6 +28,10 @@ static nxt_var_decl_t  nxt_http_vars[] = {
 
     { nxt_string("host"),
       &nxt_http_var_host,
+      0 },
+
+    { nxt_string("$"),
+      &nxt_http_var_dollar,
       0 },
 };
 
@@ -74,6 +80,16 @@ nxt_http_var_host(nxt_task_t *task, nxt_var_query_t *query, nxt_str_t *str,
     r = ctx;
 
     *str = r->host;
+
+    return NXT_OK;
+}
+
+
+static nxt_int_t
+nxt_http_var_dollar(nxt_task_t *task, nxt_var_query_t *query, nxt_str_t *str,
+    void *ctx)
+{
+    nxt_str_set(str, "$");
 
     return NXT_OK;
 }
--- a/src/nxt_var.c
+++ b/src/nxt_var.c
@@ -362,6 +362,14 @@ nxt_var_next_part(u_char *start, size_t length, nxt_str_t *part,
             return NULL;
         }
 
+        /* Allow $$ to represent a literal '$' */
+        if (*p == '$') {
+            part->start = p;
+            part->length = 1;
+
+            return p + 1;
+        }
+
         if (*p == '{') {
             bracket = 1;
 
--- a/test/test_static_chroot.py
+++ b/test/test_static_chroot.py
@@ -227,6 +227,3 @@ class TestStaticChroot(TestApplicationProto):
         assert 'error' in self.update_action(
             temp_dir + '/assets$uri', temp_dir + '/assets/d$r$uri'
         )
-        assert 'error' in self.update_action(
-            temp_dir + '/assets$uri', temp_dir + '/assets/$$uri'
-        )
--- a/test/test_static_variables.py
+++ b/test/test_static_variables.py
@@ -73,7 +73,3 @@ class TestStaticVariables(TestApplicationProto):
 
     def test_static_variables_invalid(self, temp_dir):
         assert 'error' in self.update_share(temp_dir + '/assets/d$r$uri')
-        assert 'error' in self.update_share(temp_dir + '/assets/$$uri')
-        assert 'error' in self.update_share(
-            [temp_dir + '/assets$uri', temp_dir + '/assets/dir', '$$uri']
-        )
--- a/test/test_variables.py
+++ b/test/test_variables.py
@@ -115,12 +115,58 @@ class TestVariables(TestApplicationProto):
         update_pass("applications")
         assert self.get(url='/3')['status'] == 404
 
+    def test_variables_dollar(self):
+        assert 'success' in self.conf(
+            {
+                "listeners": {"*:7080": {"pass": "routes"}},
+                "routes": [{"action": {"return": 301}}],
+            }
+        )
+
+        assert 'success' in self.conf(
+            '"https://${host}${uri}path$$sub"', 'routes/0/action/location'
+        ), 'location with $'
+        assert self.get()['headers']['Location'] == 'https://localhost/path$sub'
+
+        assert 'success' in self.conf(
+            '"https://$host$uri/$$sub1$$sub2"', 'routes/0/action/location'
+        ), 'location with 2 $s'
+        assert (
+            self.get()['headers']['Location'] == 'https://localhost//$sub1$sub2'
+        )
+
+        assert 'success' in self.conf(
+            '"https://$host/routes$uri$method$$sub"', 'routes/0/action/location'
+        ), 'location with variables followed by a $'
+        assert (
+            self.get()['headers']['Location']
+            == 'https://localhost/routes/GET$sub'
+        )
+
+        assert 'success' in self.conf(
+            '"https://$host$uri$$$host"', 'routes/0/action/location'
+        ), 'location with $ immediately followed by a variable'
+        assert (
+            self.get()['headers']['Location'] == 'https://localhost/$localhost'
+        )
+
+        assert 'success' in self.conf('"$${}"', 'routes/0/action/location')
+        assert (self.get()['headers']['Location'] == '$%7B%7D')
+
+        assert 'success' in self.conf('"$${host}"', 'routes/0/action/location')
+        assert (self.get()['headers']['Location'] == '$%7Bhost%7D')
+
+        assert 'success' in self.conf('"$$$$"', 'routes/0/action/location')
+        assert (self.get()['headers']['Location'] == '$$')
+
     def test_variables_invalid(self):
         def check_variables(routes):
             assert 'error' in self.conf(
                 routes, 'listeners/*:7080/pass'
             ), 'invalid variables'
 
+        check_variables("\"${$}\"")
+        check_variables("\"${$uri}\"")
         check_variables("\"routes$\"")
         check_variables("\"routes${\"")
         check_variables("\"routes${}\"")


docs/changes.xml
```diff
@@ -31,6 +31,12 @@ NGINX Unit updated to 1.27.0.
          date="" time=""
          packager="Andrei Belov &lt;defan@nginx.com&gt;">
 
+<change type="feature">
+<para>
+support $ sub-delimiters in URIs by using "$$".
```
alejandro-colomar, line 36: I'm not sure if we should add it to the changelog as adding support for $ in URIs, or if we should add it as supporting literal $ after variable resolution with $$.  Opinions?  Cc: @artemkonev

src/nxt_http_variables.c
```diff
@@ -77,3 +83,13 @@ nxt_http_var_host(nxt_task_t *task, nxt_var_query_t *query, nxt_str_t *str,
 
     return NXT_OK;
 }
+
+
+static nxt_int_t
+nxt_http_var_dollar(nxt_task_t *task, nxt_var_query_t *query, nxt_str_t *str,
+    void *ctx)
+{
+    *str = (nxt_str_t) nxt_string("$");
```
alejandro-colomar, line 92: Maybe we can use here `nxt_str_set()` to avoid the cast?

```c
alx@asus5775:~/src/nginx/unit$ grepc nxt_string


./src/nxt_string.h:108:

#define nxt_string(str)       { nxt_length(str), (u_char *) str }
alx@asus5775:~/src/nginx/unit$ grepc nxt_str_set


./src/nxt_string.h:113:

#define nxt_str_set(str, text)                                                \
    do {                                                                      \
        (str)->length = nxt_length(text);                                     \
        (str)->start = (u_char *) text;                                       \
    } while (0)
```

test/test_return.py
```diff
@@ -178,6 +178,21 @@ def test_return_location_edit(self):
...
+        ), 'location with $ sub-delimiter'
+        assert self.get()['headers']['Location'] == 'https://localhost/path$sub'
+
+        assert 'success' in self.conf(
+            '"https://$host$uri/$$sub1$$sub2"', 'routes/0/action/location'
+        ), 'location with 2 $ sub-delimiters'
+        assert self.get()['headers']['Location'] == 'https://localhost//$sub1$sub2'
+
+        assert 'success' in self.conf(
+            '"https://$host/routes$uri$method$$sub"', 'routes/0/action/location'
```
alejandro-colomar, line 192: Would you add some test with `$$$somevar` to make sure we don't break that?

src/nxt_http_variables.c
ac000: Good catch!

docs/changes.xml
ac000: Yes, I think you're right as this is the more general case with $ sub-delimiters being a more specific use of this. I'll also update the commit subject...

test/test_return.py
ac000: Sure, will do.

ac000:
Pushed up a V2 with the following changes

- Commit subjects changed
-  Use of nxt_set_str() as suggested by @alejandro-colomar 
-  Changed the changelog entry to be more general based on the suggestion by @alejandro-colomar 
- Added a test cover $$$var as suggested by @alejandro-colomar

src/nxt_http_variables.c
```diff
@@ -27,6 +29,10 @@ static nxt_var_decl_t  nxt_http_vars[] = {
     { nxt_string("host"),
       &nxt_http_var_host,
       0 },
+
+    { nxt_string("$$"),
```
alejandro-colomar, line 33: Other variables are using here the variable name without the `$` prefix.  Do we need to use `$$` here for some reason, or can we use simply `$`?  Does it complicate the code in some other place (or maybe make it simpler)?  Does it have any implications?

test/test_return.py
```diff
@@ -178,6 +178,26 @@ def test_return_location_edit(self):
...
+        ), 'location with variables followed by a $ sub-delimiter'
+        assert self.get()['headers']['Location'] == 'https://localhost/routes/GET$sub'
+
+        assert 'success' in self.conf(
+            '"https://$host$uri$$$host"', 'routes/0/action/location'
```
alejandro-colomar, line 197: We could also add some test to check that `${$}` fails to configure.  If that behavior ever changes, we better know.

alejandro-colomar: @andrey-zelenkov are there pytests that are specific to variables instead of having to use some feature to test this?

alejandro-colomar: Another test that might be interesting: `${$host}` (should fail) :-)

src/nxt_http_variables.c
ac000: Actually I had it using just '$' originally, easy to adjust.

test/test_return.py
ac000: Yes, ${$host} fails, that bit of code is unchanged and bails out as soon as it sees a character other than [a-z_] between the curly braces.

A quick grep on the tests doesn't show a test for that case.

I'll add tests for those two failure cases, though they are technically the same failure case (non [a-z_] character after the '{'), but I'll wait until there's consensus on where the tests should go.

alejandro-colomar:
> All the tests under the test/ directory pass, well except for a couple of chroot ones which also fail under master.

Hmm, the chroot tests work for me (Debian Sid), both in master and ac/sub-delims (and master also works in the build bot which tries a lot of systems).  What system do you have?  Do you run `pytest` from the root of the repository or from `test/`? (You should run it from the root).

ac000:
@alejandro-colomar

Heh, I know nothing about pytest!. Took me a while to even figure out it was something I needed to install and run.

So I was running it in the test/ directory (on Fedora 35 and now 36). But yes, I see now that if I run it from the repository root, all the tests pass and the output even looks nicer, so win-win!

Thanks!

alejandro-colomar:
> @alejandro-colomar
> 
> Heh, I know nothing about pytest!. Took me a while to even figure out it was something I needed to install and run.
> 
> So I was running it in the test/ directory (on Fedora 35 and now 36). 

Yup, I guessed it was that, because the exact same thing happened to me a couple of months ago :P

We could have this documented somewhere.  In my projects I use an `INSTALL` file for these things.  @artemkonev Where would you document this?  Similar things are on the website, but this is maybe too specific to programmers and not for users?

> But yes, I see now that if I run it from the repository root, all the tests pass and the output even looks nicer, so win-win!
> 
> Thanks!

:)

lcrilly:
Something for the contribution guide, @artemkonev ?

alejandro-colomar:
Another standard (at least standardised by GNU) thing is to add a 'check' target in the Makefile, that should handle the complexity whatever it is.  We could add it to our Makefile (through the <auto/> scripts)

I.e., a user can expect `make check` to run whatever is needed to run the tests, which in our case is just `pytest`.

See: <https://www.gnu.org/software/make/manual/html_node/Standard-Targets.html>

ac000:
Good call. I was thinking that myself. I'll take a look (unless you beat me to it!). It would be good if could fail gracefully, if the pytest command isn't found or it fails to run properly, I also had to install golang to get it going, it could print out some useful message about what packages it needs installed to run.

ac000:
I've put up a draft pull request at https://github.com/nginx/unit/pull/695 for discussion.

We should still definitely document this someplace.

test/test_return.py
andrey-zelenkov: Since these tests are not Location specific then it's better to move all these tests in the `test_variables_dollar()` method inside of `test_variables.py` (with all similar tests for $host, $uri, etc). test_static_variables.py is more about the tests related to the static feature while sub-delimiters look more general and related to the variables overall.

test/test_return.py
```diff
@@ -193,6 +193,26 @@ def test_return_location_edit(self):
             == ""
         ), 'location with empty variable'
 
+        assert 'success' in self.conf(
```
andrey-zelenkov, line 196: Also, you can fold tests and source code patches into a single one.  The reason why most of the tests are designed as separate commits is that they have different authors. If the author is the same then it's ok to commit all together (it's up to you).

test/test_return.py
```diff
@@ -193,6 +193,26 @@ def test_return_location_edit(self):
...
+            '"https://${host}${uri}path$$sub"', 'routes/0/action/location'
+        ), 'location with $ sub-delimiter'
+        assert self.get()['headers']['Location'] == 'https://localhost/path$sub'
+
+        assert 'success' in self.conf(
+            '"https://$host$uri/$$sub1$$sub2"', 'routes/0/action/location'
+        ), 'location with 2 $ sub-delimiters'
+        assert self.get()['headers']['Location'] == 'https://localhost//$sub1$sub2'
```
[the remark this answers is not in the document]
ac000: Hi @andrey-zelenkov 

I wasn't sure how hard and fast that rule was and it was only a few chars over.

I'll apply that patch. Thanks!

test/test_return.py
ac000: Thanks, all things being equal I think I'd keep them separate, so each commit is doing one logical change.

test/test_return.py
```diff
@@ -193,6 +193,26 @@ def test_return_location_edit(self):
...
+        assert 'success' in self.conf(
+            '"https://$host/routes$uri$method$$sub"', 'routes/0/action/location'
+        ), 'location with variables followed by a $ sub-delimiter'
+        assert self.get()['headers']['Location'] == 'https://localhost/routes/GET$sub'
+
+        assert 'success' in self.conf(
+            '"https://$host$uri$$$host"', 'routes/0/action/location'
+        ), 'location with sub-delimiter immediately followed by a variable'
+        assert self.get()['headers']['Location'] == 'https://localhost/$localhost'
+
```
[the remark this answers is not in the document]
ac000: Good suggestions.

_$${}_ gives
```
$ curl -I http://localhost:8080/
Location: $%7B%7D
```

and

_$$$$_ gives

$ curl -I http://localhost:8080/
Location: $$
```

That looks right. I'll add tests for those also.

**Comment by VBart** [comment]
1. It's pretty rare case when someone needs literal `$` in web-related strings, so there is no need to make it easier and shorter
2. `$$` complicates variables syntax, adding an exception to the existing rules, and, as a consequence, it will complicate any configuration parsing/producing code/tools that should work with Unit

Taking into account both of these points, I tend to think that `$sign` is better than `$$`.

**Comment by ac000** [comment]
@hongzhidao 

I like your idea, but unfortunately it doesn't work. The problem is we need to skip past the second '$' _and_ have the two dollars replaced by a single '$' by setting up a _nxt_var_sub_t_ for it. The original string is not modified at this point and variable substitution is done at a later stage by _nxt_var_query_finish()_

You can tweak things a little bit in nxt_var_next_part() and given a string of

```
path$$sub
```

You invariably get

```
path$$sub
```
for the translation, or
```
{
        "detail": "Unknown variable \"sub\" in the \"location\" value."
}
```
or unitd in an infinite loop always finding a '$'.

**Comment by ac000** [comment]
Here's some text I wrote up while looking at this

```C
/*
 * A description of variable substitution in nginx unit
 */

/* Variables are registered via */
nxt_int_t nxt_var_register(nxt_var_decl_t *decl, size_t n);

/* variable processing is carried out by */
nxt_var_t *nxt_var_compile(nxt_str_t *str, nxt_mp_t *mp, nxt_bool_t strz);

/*
 * *str is a nxt_str_t that contains the string to be processed
 * *mp is a memory pool from which a nxt_var_t will be allocated
 * strz says whether the string is nul terminated
 */

/*
 * The first thing nxt_var_compile() does is to determine how many
 * variables are in the string.
 *
 * This is so it knows how much memory to allocate for the below nxt_var_t
 */

/* Allocates and sets up a nxt_var_t struct of */
struct nxt_var_s {
    size_t              length;
    nxt_uint_t          vars;
    uint8_t             strz;  /* 1 bit */
    u_char              data[];

/*
    nxt_var_sub_t       subs[vars];
    u_char              raw[length];
*/
};
/*
 * data[] is a flexible array member and points to n * nxt_var_sub_t structs
 * followed by a copy of the original string being processed.
 */

/*
 * Next it finds all the variables and looks them up in the variables hash
 * table. For each valid variable it fills out a nxt_var_sub_t structure.
 */
typedef struct {
    uint32_t            index;
    uint32_t            length;
    uint32_t            position;
} nxt_var_sub_t;

/*
 * index    index into the nxt_http_vars array for the variable to substitute.
 * length   length of the string to be subtituted, e.g 7 for ${host}
 * position position in the string where to start substituting.
 */

#if 0
/* * denotes a variable */
nxt_var_compile: baz$$foo$$${host}/$$bar.html#id
         - nxt_var_compile: baz
*        - nxt_var_compile: $
  sub: @ 3, len 2
         - nxt_var_compile: foo
*        - nxt_var_compile: $
  sub: @ 8, len 2
*        - nxt_var_compile: host
  sub: @ 10, len 7
         - nxt_var_compile: /
*        - nxt_var_compile: $
  sub: @ 18, len 2
         - nxt_var_compile: bar.html#id
Found 4 variables
nxt_var_compile: routes/str
         - nxt_var_compile: routes/str
Found 0 variables
#endif

/* Actual variable substitution happens in */
static void nxt_var_query_finish(nxt_task_t *task, nxt_var_query_t *query);
```

test/test_return.py
ac000: Yes, this is more generally about being able to specify a '$' in strings which are processed for variables. I will move the tests into _test_variables.py_. Thanks!

test/test_return.py
andrey-zelenkov: I realised that tests will be broken between these two commits: `Support URI $ sub-delimiters.` and `Tests: Remove some no longer invalid tests.`. This is not a good practice (for example, we use Buildbot and if it runs tests in between these patches then everyone who subscribed to its reports will get a lot of emails). So, please combine these 2 patches into a single one. You can leave the adding of the new tests as a separate patch.

Also, please use lowercase letter and past time after the ":" sign in the commit messages. Like "Tests: added ...". I would also recommend to quote dollar sign to emphasize this as a separate sign.

So, you will get the following set:
1) `Support URI $ sub-delimiters.`
2) `Tests: added tests for "$" sub-delimiters.`

test/test_return.py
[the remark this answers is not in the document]
andrey-zelenkov: To be honest I thought something about `$${host}` in the first case but I forgot to specify a variable before sending a review  :) `$${}` is a useful test too so please keep it and add one more test using pattern above with any variable you like.

test/test_return.py
ac000: Understood. I shall make those adjustments. Thanks!.

test/test_return.py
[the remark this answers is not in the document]
ac000: OK, I'll that a test for that.

alejandro-colomar:
Hi, Valentin!

> 1. It's pretty rare case when someone needs literal `$` in web-related strings,

True.

> so there is no need to make it easier and shorter

No need, that's right.
But let's consider both, and see if it may be worth the extra readability in exchange for the extra complexity.

> 
>     2. `$$` complicates variables syntax,

True for computers; however, the opposite is true for humans, since:

- make(1) uses `$$` for `$`.  (BTW, for curiosity: I tested that `${$}` != `$$`).
- printf(3) uses `%%` for `%`.

And probably many more other similar examples exist out there, so this would be very intuitive, I think.

> adding an exception to the existing rules, and, as a consequence, it will complicate any configuration parsing

Yes, that's my biggest concern:

```c
        if (*p == '$') {
            part->start = p;
            part->length = 1;

            return p + 1;
        }
```

It is adding an extra check (conditional + jump) for every variable used in the configuration.
The good part is that the check is only run at configuration time, not at request time, so not really critical (for a big config with 1000 variables, I'd say this is still pretty negligible in terms of performance; we're probably spending much more time in allocations, I guess).


> /producing code/tools that should work with Unit

Hmm, yes.  But too much?  Do you have any such tool to have a look at?  I expect such a tool to use something like perl(1).

The following PCRE should match variables: `\$(\$|{[a-z_]+}|[a-z]+)`

```
alx@asus5775:~/tmp$ cat vars 
x$$x
x${a}x
x${.x}x
x$.{x}x
x${ab}x
x${x.x}x
x$c.x
x$cd.x
x$.x
alx@asus5775:~/tmp$ pcregrep --color=always '\$(\$|{[a-z_]+}|[a-z]+)' vars 
x$$x
x${a}x
x${ab}x
x$c.x
x$cd.x
```
The coloring is as expected.

> 
> 
> Taking into account both of these points, I tend to think that `$sign` is better than `$$`.

I don't have a strong opinion on either side; but `$sign` is ugly :).  Please consider the above and let me know your opinions.

VBart:
[..]
> True for computers; however, the opposite is true for humans, since:
> 
>     * make(1) uses `$$` for `$`.  (BTW, for curiosity: I tested that `${$}` != `$$`).
> 
>     * printf(3) uses `%%` for `%`.
> 
> 
> And probably many more other similar examples exist out there, so this would be very intuitive, I think.
> 

But remember who are Unit users: devops, webdevs, sysadmins. I'm not sure that many of them are very familiar with nuances of using make/printf.

To emphasize how insignificant this `$$` use case for a web-server, I should point out that nginx having similar variables syntax still has no built-in way to represent literal `$` at all.

[..]
> > /producing code/tools that should work with Unit
> 
> Hmm, yes. But too much? Do you have any such tool to have a look at?

Well, I guess even unit-cli and Unit website syntax highlighting haven't been designed yet to recognise `$$`. Unit API is perfect for automation, and users definitely wrote and will write a lot of various scripts to work with it. Each time, when they decide to work with variables in config somehow, they should be aware of this `$$`, which is syntactically different from any other variables.

[..]
> I don't have a strong opinion on either side; but $sign is ugly :). Please consider the above and let me know your opinions.

I agree that `$$` looks good and in other cases I'd prefer it over `$sign`, but taking into account how rare this will be used, I'd better to hide literal `$` among other variables.

alejandro-colomar:
@VBart 

[..]
> But remember who are Unit users: devops, webdevs, sysadmins. I'm not sure that many of them are very familiar with nuances of using make/printf.
> 
> To emphasize how insignificant this `$$` use case for a web-server, I should point out that nginx having similar variables syntax still has no built-in way to represent literal `$` at all.
> 

[...]
> Well, I guess even unit-cli and Unit website syntax highlighting haven't been designed yet to recognise `$$`. Unit API is perfect for automation, and users definitely wrote and will write a lot of various scripts to work with it. Each time, when they decide to work with variables in config somehow, they should be aware of this `$$`, which is syntactically different from any other variables.
> 
> [..]
> 
> > I don't have a strong opinion on either side; but $sign is ugly :). Please consider the above and let me know your opinions.
> 
> I agree that `$$` looks good and in other cases I'd prefer it over `$sign`, but taking into account how rare this will be used, I'd better to hide literal `$` among other variables.

Makes sense.  Let's add `$sign` for now.  If in the future (like 10 yr. from now), when we know we won't be breaking (or making more difficult) other's tools, we are always in time to add `$$` as an alias, if we see a need.

ac000:
> Makes sense. Let's add $sign for now.

Is 'sign' the right name to use though? As we'd now have a variable called 'sign' which could be anything.

Perhaps *${dollar}* would be better? And if it's something that is barely used then having the couple of extra characters to type shouldn't be a problem.

VBart:
> > Makes sense. Let's add $sign for now.
> 
> Is 'sign' the right name to use though? As we'd now have a variable called 'sign' which could be anything.
> 
> Perhaps _${dollar}_ would be better? And if it's something that is barely used then having the couple of extra characters to type shouldn't be a problem.

`$dollar` is ok, but when reading and voicing the `$` itself, this sounds in mind like duplication. So that's why `$sign` is suggested, as it sounds like "(dollar)sign".

tippexs:
Hi Folks - following the discussion like all the points and the actual effort we put in every single feature and change! That makes sure we will keep the quality high.💪🏼

I have looked around and found this 

https://www.toptal.com/designers/htmlarrows/currency/dollar-sign/

as `&dollar;` is the HTML representation of $ I would like to suggest using `$dollar` for our implementation to make it a little bit aligned to something that already exists.

alejandro-colomar:
> For completeness, here's my updated 'sub-delims' branch which I think addresses all the above feedback.

Thanks!  Looks very good to me.

If you're ok with it, I'm going to host it in my own git server too.  I've called the branch `ac/dollar-dollar` (rebased against the current master).

<http://www.alejandro-colomar.es/src/alx/nginx/unit.git/log/?h=ac/dollar-dollar>

> 
> Next I'll dust off my 'simple-dollar' branch and submit that instead, maybe as a new pull-request...

Sure.  You can close this one when you want.

PR #695, opened 2022-05-13
Build: Add a make check target.

ac000:
While unit comes with a test suite, it is not entirely obvious how to
run it!.

As suggested by @alejandro-colomar add a make check target which is a
common target for running such things. This is added to the Makefile in
the repository root. It will build unit first if required.

If pytest fails to run for some reason (but not due to tests failing)
then a message is displayed saying what packages need to be installed on
some common platforms.

On FreeBSD and Fedora, pytest can be run as pytest or pytest-3, on
Debian it is pytest-3, so make tries running pytest first then pytest-3
if pytest wasn't found.

### DIFF
--- a/auto/make
+++ b/auto/make
@@ -430,6 +430,12 @@ cat << END > Makefile
 
 include $NXT_MAKEFILE
 
+PYTEST := $NXT_PYTEST_NAME
+
+.PHONY: check
+check: unitd
+	\$(PYTEST)
+
 .PHONY: clean
 clean:
 		rm -rf $NXT_BUILD_DIR *.dSYM Makefile
--- a/auto/pytest
+++ b/auto/pytest
@@ -1,0 +1,58 @@
+
+# Copyright (C) Andrew Clayton
+
+
+$echo -n "checking for pytest ... "
+cat << END >> $NXT_AUTOCONF_ERR
+----------------------------------------
+checking for pytest
+END
+
+
+# Allow error exit status.
+set +e
+
+NXT_PYTEST_NAME=
+
+PYTEST_DEFAULT_NAME=pytest
+PYTEST_FOUND_STR="not found"
+PYTEST_GO=
+
+PYTEST_CMDS="
+    pytest
+    pytest-3
+"
+
+for pyt_cmd in $PYTEST_CMDS
+do
+    if `/bin/sh -c "($pyt_cmd --version)" 2>&1 >> $NXT_AUTOCONF_ERR 2>&1`
+    then
+        NXT_PYTEST_NAME=$pyt_cmd
+        break
+    fi
+done
+
+if `/bin/sh -c "(go version)" 2>&1 >> $NXT_AUTOCONF_ERR 2>&1`
+then
+    PYTEST_GO=go
+fi
+
+if [ ! -z $NXT_PYTEST_NAME ]
+then
+    PYTEST_FOUND_STR="found"
+fi
+
+if [ ! -z $NXT_PYTEST_NAME ] && [ -z $PYTEST_GO ]
+then
+    PYTEST_FOUND_STR=${PYTEST_FOUND_STR}" but is missing golang"
+fi
+
+$echo $PYTEST_FOUND_STR
+
+if [ -z $NXT_PYTEST_NAME ]
+then
+    NXT_PYTEST_NAME=$PYTEST_DEFAULT_NAME
+fi
+
+# Stop on error exit status again.
+set -e
--- a/configure
+++ b/configure
@@ -163,6 +163,7 @@ cat << END >> $NXT_AUTO_CONFIG_H
 END
 
 . auto/test_build
+. auto/pytest
 . auto/sources
 . auto/save
 


ac000:
I didn't create a changelog entry, does it need one?

alejandro-colomar:
I'd keep the Makefile simple:

```mk
check: unitd
	$(PYTEST)
```

Where PYTEST would resolve to pytest or one of the alt names such as pytest-3.  If no pytest is found, it should resolve to pytest, so that we get `sh: 1: pytest: not found`.

I like the standard error message of `sh: 1: cmd: not found`.  It's clear and concise.  The first thing to do in such a case is `apt-file find /bin/pytest` or equivalent in other systems.  Reading the `./configure` `stderr` for warnings is another common step.  Then, our (yet missing) documentation would be the other thing to look at.

I think those 3 things should be enough, and we shouldn't add too much to the Makefile itself.

I'd use `\$(PYTEST)` so that one can run `make check PYTEST=pytest-3`.

What do you think?

ac000:
> Where $(PYTEST) would resolve to pytest or one of the alt names such as pytest-3. If no pytest is found, it should resolve to pytest, so that we get sh: 1: pytest: not found.

So just to double check, you're saying to set PYTEST in the Makefile to the pytest executable, defaulting to pytest (for example if pytest isn't installed)? But then you can override that from the make command?

When pytest is installed but no golang you get a screen full of errors, at first I just thought it was generally broken until I looked closer and noticed the references to 'go' and figured I needed to also install golang.

alejandro-colomar:
> > Where $(PYTEST) would resolve to pytest or one of the alt names such as pytest-3. If no pytest is found, it should resolve to pytest, so that we get sh: 1: pytest: not found.
> 
> So just to double check, you're saying to set PYTEST in the Makefile to the pytest executable, defaulting to pytest (for example if pytest isn't installed)? But then you can override that from the make command?

Not exactly.  Test which pytests are available within the `./configure` (and `auto/` scripts), the same way we test which compilers are available, and then copy to the makefile the result from those checks.  The makefile shouldn't use `pytest` directly, but `PYTEST := pytest` (or similar).

> 
> When pytest is installed but no golang you get a screen full of errors, at first I just thought it was generally broken until I looked closer and noticed the references to 'go' and figured I needed to also install golang.

Hmm, that's a bit of a problem.  But, I expect that by having it in the documentation (which we don't have yet), and a bit of intuition, programmers should know what's the problem (as you did, even without the docs).

ac000:
OK, I see. Thanks.

ac000:
V2 of this pull-request has the following change

- The help text has been removed from auto/make and the logic for finding the pytest command to run has been moved to a new auto/pytest, where we also do a wee check for golang.

I have put this configure check at the end of the other "checking for ..." tests and you will now see 1 of 3 messages

-  checking for pytest ... not found

No pytest installed/found.

- checking for pytest ... found but is missing golang

pytest found but no golang installed/found and mirrors a similar message
```
checking for pthread spinlock zero initial value ... found but is not working
```

- checking for pytest ... found

pytest and golang found

With pytest not installed running make check gives

```
make: pytest: No such file or directory
make: *** [Makefile:8: check] Error 127
```
and from build/autoconf.err we have
```
----------------------------------------
checking for pytest
/bin/sh: line 1: pytest: command not found
/bin/sh: line 1: pytest-3: command not found
/bin/sh: line 1: go: command not found
```

auto/make
```diff
@@ -430,6 +430,12 @@ cat << END > Makefile
 
 include $NXT_MAKEFILE
 
+PYTEST := $NXT_PYTEST_NAME
+
+.PHONY: check
+check: unitd
+	@\$(PYTEST)
```
alejandro-colomar, line 437: Don't hide the command (we're showing most commands in the Makefile, with few exceptions such as rmdir(1)).

ac000: Sure, OK

ac000:
Rebased with master (not that there were any conflicts).

ac000:
Sure, having a readme in there would be useful. But having having 'make check' is also useful and pretty standard. It was the first thing I tried.

andrey-zelenkov:
Sorry, I just can't get what problem we are trying to solve with this patch.

For example, the presence of the "go" doesn't guarantee that tests with the go module will work. It's necessary but not enough. Thus users may be misled by suggested output. Even more, the presence of the "pytest" itself doesn't guarantee that our tests will work (tests work with specific versions of pytest indicated in [requirements.txt](https://github.com/nginx/unit/blob/master/test/requirements.txt)).

And if we are going to add a check for a "go", then we also need to add the same checks for "javac", "node", ssl, etc. But all these checks are already implemented in our test module.

So from my side I would suggest to add good `README` file (in case if we are trying to solve problem with poor documented test process).

alejandro-colomar:
@andrey-zelenkov The thing with go is that it's required to run all pytests, I think.  Even the core Unit pytests.  Not just for the go module.

But we could just omit that test, and assume that users read the `test/requirements.txt` file (but please add a `test/README`).  BTW, I'd say that requirements.txt should contain go.  I think it doesn't work without go (but we should check).

We could also specify those dependencies in the Build-Depends (<https://github.com/nginx/unit/blob/master/pkg/deb/debian/control.in>) @defanator .

ac000:
Without golang installed 

```
$ pytest
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/usr/lib/python3.10/site-packages/_pytest/main.py", line 267, in wrap_session
INTERNALERROR>     config.hook.pytest_sessionstart(session=session)
INTERNALERROR>   File "/usr/lib/python3.10/site-packages/pluggy/_hooks.py", line 265, in __call__
INTERNALERROR>     return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
INTERNALERROR>   File "/usr/lib/python3.10/site-packages/pluggy/_manager.py", line 80, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
INTERNALERROR>   File "/usr/lib/python3.10/site-packages/pluggy/_callers.py", line 60, in _multicall
INTERNALERROR>     return outcome.get_result()
INTERNALERROR>   File "/usr/lib/python3.10/site-packages/pluggy/_result.py", line 60, in get_result
INTERNALERROR>     raise ex[1].with_traceback(ex[2])
INTERNALERROR>   File "/usr/lib/python3.10/site-packages/pluggy/_callers.py", line 39, in _multicall
INTERNALERROR>     res = hook_impl.function(*args)
INTERNALERROR>   File "/home/andrew/programming/unit/test/conftest.py", line 204, in pytest_sessionstart
INTERNALERROR>     option.available['modules']['go'] = check_go()
INTERNALERROR>   File "/home/andrew/programming/unit/test/unit/check/go.py", line 5, in check_go
INTERNALERROR>     process = TestApplicationGo.prepare_env('empty')
INTERNALERROR>   File "/home/andrew/programming/unit/test/unit/applications/lang/go.py", line 65, in prepare_env
INTERNALERROR>     process = subprocess.run(args, env=env, cwd=temp_dir)
INTERNALERROR>   File "/usr/lib64/python3.10/subprocess.py", line 501, in run
INTERNALERROR>     with Popen(*popenargs, **kwargs) as process:
INTERNALERROR>   File "/usr/lib64/python3.10/subprocess.py", line 966, in __init__
INTERNALERROR>     self._execute_child(args, executable, preexec_fn, close_fds,
INTERNALERROR>   File "/usr/lib64/python3.10/subprocess.py", line 1842, in _execute_child
INTERNALERROR>     raise child_exception_type(errno_num, err_msg, err_filename)
INTERNALERROR> FileNotFoundError: [Errno 2] No such file or directory: 'go'
```

On three different systems, Fedora, Debian and FreeBSD, installing pytest3 and golang got the tests going, which is why I was specifically checking for go also...

andrey-zelenkov:
@ac000 Thank you for pointing out this issue. But this error doesn't mean that the `go` is ***required*** for the tests. Tests are written (at that moment) using pytest framework that doesn't require golang. So, if you check the error then you can see that it happens in the `check_go()` function which just relies on the existence of the `go` in the system (and it's true for the default systems that we ship). To be honest there are many assumptions in the tests that we do not check properly (existing of the `ps`, `go`, various flags, etc). It's not good and it can be fixed if it really hurts. But if it's not a real use case, that at that moment we can focus on more prioritised tasks in tests (i.e poor coverage, parallelization, missing README, etc).

Anyway, thank you for raising this, let me know if the following patch works for you (for example `pytest test/test_variables.py` should work without installed `go` with this patch):

```diff --git a/test/unit/applications/lang/go.py b/test/unit/applications/lang/go.py
--- a/test/unit/applications/lang/go.py
+++ b/test/unit/applications/lang/go.py
@@ -9,6 +9,11 @@ from unit.option import option
 class TestApplicationGo(TestApplicationProto):
     @staticmethod
     def prepare_env(script, name='app', static=False):
+        try:
+            subprocess.check_output(['which', 'go'])
+        except:
+            return None
+
         temp_dir = option.temp_dir + '/go/'

         if not os.path.exists(temp_dir):
```

And going back to my original concern: I still can't understand the argumentation of this pull request.

Issue #697, opened 2022-05-17
Undefined befavior introduced in "variables in location"


alejandro-colomar:
```c
$ grepc -tfd nxt_http_return
./src/nxt_http_return.c:71:
nxt_http_action_t *
nxt_http_return(nxt_task_t *task, nxt_http_request_t *r,
    nxt_http_action_t *action)
{
    ...

    if (conf->location == NULL) {
        nxt_str_null(&loc);

    } else {
        ...
    }

    nxt_debug(task, "http return: %d (loc: \"%V\")", conf->status, &loc);

    ...
}
```

```c
$ grepc -tm nxt_debug
./src/nxt_log.h:98:
#define nxt_debug(task, ...)                                                  \
    do {                                                                      \
        nxt_log_t  *_log = (task)->log;                                       \
                                                                              \
        if (nxt_slow_path(_log->level == NXT_LOG_DEBUG || nxt_debug)) {       \
            _log->handler(NXT_LOG_DEBUG, _log, __VA_ARGS__);                  \
        }                                                                     \
    } while (0)


...
```

```c
$ grepc -tt nxt_log_t
./src/nxt_main.h:26:
typedef struct nxt_log_s             nxt_log_t;


./src/nxt_log.h:27:
struct nxt_log_s {
    ...
    nxt_log_handler_t      handler;
    ...
};
```

```c
$ grepc -tt nxt_log_handler_t
./src/nxt_log.h:22:
typedef void nxt_cdecl (*nxt_log_handler_t)(nxt_uint_t level, nxt_log_t *log,
    const char *fmt, ...);
```

```c
$ grep -rn '(nxt_uint_t \w\+, nxt_log_t \*\w\+,'
...
src/nxt_log.h:40:NXT_EXPORT void nxt_cdecl nxt_log_handler(nxt_uint_t level, nxt_log_t *log,
src/nxt_log.c:67:nxt_log_handler(nxt_uint_t level, nxt_log_t *log, const char *fmt, ...)
...
```

```c
$ grepc -tfd nxt_log_handler
./src/nxt_log.c:66:
void nxt_cdecl
nxt_log_handler(nxt_uint_t level, nxt_log_t *log, const char *fmt, ...)
{
    ...
    p = nxt_vsprintf(p, end, fmt, args);
    ...
}
```

```c
$ grepc -tfd nxt_vsprintf
./src/nxt_sprintf.c:97:
u_char *
[...]

PR #699, opened 2022-05-17
Uncover memcpy(0, 0, 0) Undefined Behavior

alejandro-colomar:
This patch should NOT be applied to unit.  It's only a testing patch that uncovers (most) hidden cases of `memcpy(NULL, NULL, 0)` by aborting unit.

Please peruse the patch to test until unit doesn't abort.


alejandro-colomar:
It's the same issue as <https://github.com/nginx/unit/issues/697>; I just found it to be present in other places too.

hongzhidao:
> This patch should NOT be applied to unit. It's only a testing patch that uncovers (most) hidden cases of memcpy(NULL, NULL, 0) by aborting unit.

What does this mean? Do I need to review it?

Issue #675, opened 2022-04-13
Support `$` as a URI sub-delimiter


alejandro-colomar:
`$` is a reserved character for URIs, for use as a subdelimiter.  However, we already use that character for Unit variables.  To be able to Represent a literal `$`, we need some variable or character sequence that evaluates to the `$` itself.

For more details, including RFC references, see the PR where this issue was first raised:
https://github.com/nginx/unit/pull/660

@VBart proposed adding a variable called something like `$sign`.  That's easy, as it only requires adding a new variable with a static content.

I suggest an alternative, `$$` (as in [_make_(1)](https://man7.org/linux/man-pages/man1/make.1.html) for example: <https://www.gnu.org/software/make/manual/make.html#Basics-of-Variable-References>).  Implementing it will probably be more complex, probably needing to change the parsing code, but the result might be nicer for the user.

alejandro-colomar:
I should note that while discussing some completely unrelated thing, I just met a URI in the wild that had a `$` subdelimiter, so they are actually in use in some places.

ac000:
So, yes, using ${sign} to represent a literal _$_ is straightforward and the patch looks like

```diff
diff --git a/src/nxt_http_variables.c b/src/nxt_http_variables.c
index 1c0b561..0adcb8c 100644
--- a/src/nxt_http_variables.c
+++ b/src/nxt_http_variables.c
@@ -13,6 +13,8 @@ static nxt_int_t nxt_http_var_uri(nxt_task_t *task, nxt_var_query_t *query,
     nxt_str_t *str, void *ctx);
 static nxt_int_t nxt_http_var_host(nxt_task_t *task, nxt_var_query_t *query,
     nxt_str_t *str, void *ctx);
+static nxt_int_t nxt_http_var_dollar(nxt_task_t *task, nxt_var_query_t *query,
+    nxt_str_t *str, void *ctx);
 
 
 static nxt_var_decl_t  nxt_http_vars[] = {
@@ -27,6 +29,10 @@ static nxt_var_decl_t  nxt_http_vars[] = {
     { nxt_string("host"),
       &nxt_http_var_host,
       0 },
+
+    { nxt_string("sign"),
+      &nxt_http_var_dollar,
+      0 },
 };
 
 
@@ -77,3 +83,13 @@ nxt_http_var_host(nxt_task_t *task, nxt_var_query_t *query, nxt_str_t *str,
 
     return NXT_OK;
 }
+
+
+static nxt_int_t
[...]

PR #702, opened 2022-05-18
Add a $dollar variable that translates to a literal '$'

ac000:
This pull-request is for supporting a new variable called 'dollar' that can be
used for specifying _$_ 's in URIs etc. One such use case being $ sub-delimiters.

This pull-request comprises the following commits

- Var: added a $dollar variable that translates to a '$'.
- Tests: added tests for translating $sign into a literal $.

Wasn't sure if it _really_ needed tests as the current variable tests should
cover it, but added just a couple of new tests to ensure we are getting $'s
out. That commit can simply be dropped if it's not wanted.

So if you wanted to specify a URL like

    https://example.com/path/15$1588/9925$2976.html

that would go in the unit config like

    https://example.com/path/15${dollar}1588/9925${dollar}2976.html

All tests pass.

This relates to this [issue](https://github.com/nginx/unit/issues/675).

### DIFF
--- a/test/test_variables.py
+++ b/test/test_variables.py
@@ -114,6 +114,27 @@ class TestVariables(TestApplicationProto):
         check_user_agent('', 404)
         check_user_agent('no', 404)
 
+    def test_variables_dollar(self):
+        assert 'success' in self.conf(
+            {
+                "listeners": {"*:7080": {"pass": "routes"}},
+                "routes": [{"action": {"return": 301}}],
+            }
+        )
+
+        def check_dollar(location, expect):
+            assert 'success' in self.conf(
+                '"' + location + '"',
+                'routes/0/action/location',
+            )
+            assert self.get()['headers']['Location'] == expect
+
+        check_dollar(
+            'https://${host}${uri}path${dollar}dollar',
+            'https://localhost/path$dollar',
+        )
+        check_dollar('path$dollar${dollar}', 'path$$')
+
     def test_variables_many(self):
         self.conf_routes("\"routes$uri$method\"")
         assert self.get(url='/5')['status'] == 206, 'many'
--- a/docs/changes.xml
+++ b/docs/changes.xml
@@ -62,6 +62,13 @@ increased the applications' startup timeout.
 </para>
 </change>
 
+<change type="feature">
+<para>
+added a new variable, $dollar, that translates to a literal "$" during
+variable substitution.
+</para>
+</change>
+
 </changes>
 
 
--- a/src/nxt_http_variables.c
+++ b/src/nxt_http_variables.c
@@ -7,6 +7,8 @@
 #include <nxt_http.h>
 
 
+static nxt_int_t nxt_http_var_dollar(nxt_task_t *task, nxt_str_t *str,
+    void *ctx, uint16_t field);
 static nxt_int_t nxt_http_var_method(nxt_task_t *task, nxt_str_t *str,
     void *ctx, uint16_t field);
 static nxt_int_t nxt_http_var_request_uri(nxt_task_t *task, nxt_str_t *str,
@@ -41,6 +43,9 @@ static nxt_int_t nxt_http_var_cookie(nxt_task_t *task, nxt_str_t *str,
 
 static nxt_var_decl_t  nxt_http_vars[] = {
     {
+        .name = nxt_string("dollar"),
+        .handler = nxt_http_var_dollar,
+    }, {
         .name = nxt_string("method"),
         .handler = nxt_http_var_method,
     }, {
@@ -93,6 +98,15 @@ nxt_int_t
 nxt_http_register_variables(void)
 {
     return nxt_var_register(nxt_http_vars, nxt_nitems(nxt_http_vars));
+}
+
+
+static nxt_int_t
+nxt_http_var_dollar(nxt_task_t *task, nxt_str_t *str, void *ctx, uint16_t field)
+{
+    nxt_str_set(str, "$");
+
+    return NXT_OK;
 }
 
 


alejandro-colomar:
[Just my obsession for consistency talking; feel free to disagree]

My opinion is that:

```c
static nxt_var_decl_t  nxt_http_vars[] = {
    { nxt_string("method"),
      &nxt_http_var_method,
      0 },
    { nxt_string("uri"),
      &nxt_http_var_uri,
      0 },
    { nxt_string("host"),
      &nxt_http_var_host,
      0 },

    /* literal $ sign */
    { nxt_string("sign"),
      &nxt_http_var_dollar,
      0 },
};
```
looks... inconsistent.

Of course, `nxt_http_var_dollar()` is the only name that makes sense for the function.  And that asks for `"dollar"` being the actual name for the variable.

In addition, consistency with other languages that have defined it similarly, as @tippexs found, would also be nice, and since they used `&dollar;`, I see another reason to go for `$dollar` here.

But, you write it, so do whatever looks better to you; both make sense to me (and to Valentin it's nicer the other way it seems :)

 BTW, I'd remove the `/* literal $ sign */` comment; I think it should be obvious.

ac000:
Hey, I'm all for $dollar, it was my [suggestion](https://github.com/nginx/unit/pull/693#issuecomment-1130308254) originally...

I think I named it like that

1) It says what it is actually doing
2) I half anticipated we'd use $dollar

So, I'll switch it over to _dollar_.

alejandro-colomar (approved):

test/test_variables.py
```diff
@@ -115,6 +115,27 @@ def update_pass(prefix):
         update_pass("applications")
         assert self.get(url='/3')['status'] == 404
 
+    def test_variables_dollar(self):
```
andrey-zelenkov, line 118: Please apply the following style fixes for the tests (nothing significant, just making life easier if we decide to add more tests in the future).

Also I would suggest to move this test a little bit closer to the similar test for $host, $method, etc (to match organization in C code). This is not a requirement.

```
diff --git a/test/test_variables.py b/test/test_variables.py
--- a/test/test_variables.py
+++ b/test/test_variables.py
@@ -53,6 +53,27 @@ class TestVariables(TestApplicationProto
         check_host('www.localhost', 404)
         check_host('localhost1', 404)
 
+    def test_variables_dollar(self):
+            {
+                "listeners": {"*:7080": {"pass": "routes"}},
+                "routes": [{"action": {"return": 301}}],
+            }
+        )
+
+        def check_dollar(location, expect):
+            assert 'success' in self.conf(
+                '"' + location + '"',
+                'routes/0/action/location',
+            )
+            assert self.get()['headers']['Location'] == expect
+
+        check_dollar(
+            'https://${host}${uri}path${dollar}dollar',
+            'https://localhost/path$dollar',
+        )
+        check_dollar('path$dollar${dollar}', 'path$$')
+
     def test_variables_many(self):
         self.conf_routes("\"routes$uri$method\"")
         assert self.get(url='/5')['status'] == 206, 'many'
@@ -115,27 +136,6 @@ class TestVariables(TestApplicationProto
         update_pass("applications")
         assert self.get(url='/3')['status'] == 404
 
-    def test_variables_dollar(self):
-        assert 'success' in self.conf(
-            {
-                "listeners": {"*:7080": {"pass": "routes"}},
-                "routes": [{"action": {"return": 301}}],
-            }
-        )
-
-        assert 'success' in self.conf(
-            '"https://${host}${uri}path${dollar}sub"',
-            'routes/0/action/location'
-        ), 'location with $'
-        assert self.get()['headers']['Location'] == 'https://localhost/path$sub'
-
-        assert 'success' in self.conf(
-            '"path$dollar${dollar}"', 'routes/0/action/location'
-        ), 'path followed by $$'
-        assert (
-            self.get()['headers']['Location'] == 'path$$'
-        )
-
     def test_variables_invalid(self):
         def check_variables(routes):
             assert 'error' in self.conf(

```

ac000: Ah, OK I see. Thanks!

ac000:
Yes, there was something strange going with the GitHub markdown, I fixed it before by doing

```$ 's```

rather than

```$'s```

but it came back briefly when I refreshed the page, but now I fully reloaded it, it isn't showing funny for me now.

I'll have another play with it.

**Comment by ac000** [comment]
It's that very first '$' in the first paragraph that seems to cause issues.

If I have just '$' it's messed up, if I have '\$' it's messed up. If I put '\\$' I get '\$' and everything else looks fine.

What I actually wrote was
```
If I have just '$' it's messed up, if I have '\$' it's messed up. If I put '\\$' I get '\$' and everything else looks fine.
```

At the moment I have it as

```_$_ 's```

which is displaying OK for me in Firefox. Just tried it in epiphany and it looks OK too.

Heh, that will look somewhat messed up if the 'preview' is anything to go by, but I guess proves that GitHub is doing something funky! So I've shown what I actually wrote also.

AFAIK '$' isn't anything special in markdown...

alejandro-colomar:
Ahh, sorry, I didn't expect github to show it funny :p

I meant in the commit message for the tests, where it says:

`If you need to specify a $ in a URI you can now use '$dollar/${dollar}'.`

The last .../... reads a bit weird to me.  Maybe quoting both alternatives separately would read better?

ac000:
Sure, I just assumed that '/' was universally understood as 'or', but I'll spell it out...

alejandro-colomar:
> Sure, I just assumed that '/' was universally understood as 'or', but I'll spell it out...

Yes, but since you're quoting it all at once, it reads (to me) as if the `/` was part of a literal.

ac000:
Well, should be unambiguous now... ;-)

ac000:
- Rebased against master
- Updated changelog entry text to more match surrounding entries

andrey-zelenkov (approved):

ac000:
Rebased against current master.

artemkonev (approved):

PR #706, opened 2022-05-26
Var: Added `$request_uri` (as in NGINX).

alejandro-colomar:
The commit is unfinished.  r->request_uri doesn't exist yet.


test/test_variables.py
```diff
@@ -17,6 +17,7 @@ def setup_method(self):
                     "5GET": [{"action": {"return": 206}}],
                     "GETGET": [{"action": {"return": 207}}],
                     "localhost": [{"action": {"return": 208}}],
+                    "9*?q": [{"action": {"return": 209}}],
```
alejandro-colomar, line 20: @andrey-zelenkov 

I don't understand why I need to put here `"9*?q"`, and `"9%2A?q"` doesn't work with `$request_uri`, since the contents of that variable are encoded.


See my own test:

```
{
	"listeners": {
		"*:81": {
			"pass": "routes/ru"
		}
	},

	"routes": {
		"ru": [{
			"action": {
				"return": 301,
				"location": "$request_uri"
			}
		}]
	}
}
```
```
$ curl -i http://localhost:81/*foo%2Abar?baz#arg
HTTP/1.1 301 Moved Permanently
Location: /*foo%2Abar?baz
Server: Unit/1.27.0
Date: Mon, 30 May 2022 16:04:30 GMT
Content-Length: 0

```

alejandro-colomar: @hongzhidao  Ahh, probably variable contents are normalized before comparing against fields in the configuration.

Issue #715, opened 2022-06-03
Ruby/Rails: config.ru in request.path


lluuaapp:
Hi!

Since the lates release (1.27) all my rails/rack app gets an extra `config.ru` prefix added to the `request.path`. The relatively simple config is relatively simple:

```
  "applications": {
    "app_name": {
      "type": "ruby",
      "processes": 2,
      "working_directory": "/usr/local/app_name",
      "user": "www-data",
      "group": "www-data",
      "script": "/usr/local/app_name/config.ru",
      "environment": {
        "RAILS_ENV": "production",
        "SECRET_KEY_BASE": "blah"
      }
    }
```

Seems the new index name feature has broken something there?

Please tell me if you need any further information. Reverted back to 1.26.1 and it works again.

ac000:
Hi

My ruby is almost non-existent but I don't think I'm seeing what you're seeing.

Here's my config

```JSON
{
    "listeners": {
        "*:8080": {
            "pass": "applications/test"
        }
    },
    "applications": {
        "test": {
            "type": "ruby",
            "script": "/var/tmp/config.ru"
        }
    }
}
```

Here's my config.ru

```ruby
app = Proc.new do |env|                                                         
    p env                                                                       
    ["200", {                                                                   
        "Content-Type" => "text/plain",                                         
    }, ["Hello World"]]                                                         
end                                                                             
                                                                                
run app
```

Here's my request

```
$ curl http://localhost:8080/test
Hello World
```

Here's 'env'
```
{"SERVER_SOFTWARE"=>"Unit/1.27.0", "SCRIPT_NAME"=>"config.ru", "rack.version"=>[1, 3],
"rack.input"=>#<NGINX_Unit_Stream_IO_Read:0x00007ff183ed1120>, 
"rack.errors"=>#<NGINX_Unit_Stream_IO_Error:0x00007ff183ed0ec8>, "rack.multithread"=>false,
"rack.multiprocess"=>true, "rack.run_once"=>false, "rack.hijack?"=>false, "rack.hijack"=>nil, 
"rack.hijack_io"=>nil, "REQUEST_METHOD"=>"GET", "REQUEST_URI"=>"/test", 
"PATH_INFO"=>"/test", "QUERY_STRING"=>"", "SERVER_PROTOCOL"=>"HTTP/1.1",
"REMOTE_ADDR"=>"127.0.0.1", "SERVER_ADDR"=>"127.0.0.1", "SERVER_NAME"=>"localhost",
[...]

PR #719, opened 2022-06-09
Ruby: fixed contents of SCRIPT_NAME.

alejandro-colomar:
Having the basename of the script pathname was incorrect.  While
we don't have something more accurate, the best thing to do is to
have it empty (which should be most of the time).

Fixes (git) commit 0032543fa65f454c471c968998190b027c1ff270,
which was 'Ruby: added the Rack environment parameter "SCRIPT_NAME".'.

Closes: <https://github.com/nginx/unit/issues/715>

### DIFF
--- a/docs/changes.xml
+++ b/docs/changes.xml
@@ -75,6 +75,12 @@ increased the applications' startup timeout.
 </para>
 </change>
 
+<change type="bugfix">
+<para>
+force SCRIPT_NAME in Ruby to always be an empty string.
+</para>
+</change>
+
 </changes>
 
 
--- a/src/ruby/nxt_ruby.c
+++ b/src/ruby/nxt_ruby.c
@@ -29,7 +29,6 @@ typedef struct {
 static nxt_int_t nxt_ruby_start(nxt_task_t *task,
     nxt_process_data_t *data);
 static VALUE nxt_ruby_init_basic(VALUE arg);
-static VALUE nxt_ruby_script_basename(nxt_str_t *script);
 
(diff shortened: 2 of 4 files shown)


andrey-zelenkov (approved):

docs/changes.xml
```diff
@@ -5,6 +5,13 @@
...
 
 
+<change type="bugfix">
+<para>
+fixed contents of SCRIPT_NAME in Ruby to always be an empty string.
+</para>
+</change>
+
+
 <changes apply="unit-php
```
andrey-zelenkov, line 15: This should be placed into 1.28.0 changes.

alejandro-colomar: Rebased to master, where 1.28.0 changes exist, and moved there.

tippexs (approved):

docs/changes.xml
```diff
@@ -62,6 +62,12 @@ increased the applications' startup timeout.
 </para>
 </change>
 
+<change type="bugfix">
+<para>
+fixed contents of SCRIPT_NAME in Ruby to always be an empty string.
```
artemkonev, line 67: I suggest "force SCRIPT_NAME in Ruby to always be an empty string".

andrey-zelenkov: Probably s/force/forced/ for consistency (past time used in other places)?

alejandro-colomar: Hmm, I'd say no, because in `bugfix` we usually use past to refer to the issue that happened in the past, but this sentence refers to what happens after the fix.

PR #723, opened 2022-06-15
Put function attributes where ISO C2x specifies.

alejandro-colomar:
GCC is very generous in where attributes can be placed.  C2x, the
first ISO C standard that will standardize attributes, in order to
be less ambiguous, decided to only allow attributes to be placed
in very specific places.  Function attributes go at the very
beginning of a function prototype (it also allows them to go after
the function identifier and just before the parameter list's
opening parenthesis, but that's a side effect of being compatible
with C++'s syntax, and should probably be never used in C, and
anyway I don't like it for many reasons, including that it breaks
the ability to grep for functions using a regex).

See the following test that helps understand this:

$ cat inline.c
[[gnu::always_inline]] static inline void a(void);
static [[gnu::always_inline]] inline void b(void);
static inline [[gnu::always_inline]] void c(void);
static inline void [[gnu::always_inline]] d(void);
static inline void e [[gnu::always_inline]] (void);
static inline void f(void)[[gnu::always_inline]];
static inline void g(void);

static inline void a(void) { return; }
static inline void b(void) { return; }
static inline void c(void) { return; }
static inline void d(void) { return; }
static inline void e(void) { return; }
static inline void f(void) { return; }
static inline void g(void) { return; }

$ clang -Wall -Wextra -Werror -std=c2x inline.c -c
inline.c:2:8: error: an attribute list cannot appear here
static [[gnu::always_inline]] inline void b(void);
       ^~~~~~~~~~~~~~~~~~~~~~
inline.c:3:15: error: an attribute list cannot appear here
static inline [[gnu::always_inline]] void c(void);
              ^~~~~~~~~~~~~~~~~~~~~~
inline.c:4:22: error: 'always_inline' attribute cannot be applied to types
static inline void [[gnu::always_inline]] d(void);
                     ^
inline.c:6:29: error: attribute 'always_inline' ignored, because it cannot be applied to a type [-Werror,-Wignored-attributes]
static inline void f(void)[[gnu::always_inline]];
                            ^
4 errors generated.

$ cc -Wall -Wextra -Werror -std=c2x inline.c -c
inline.c:2:1: error: ‘always_inline’ attribute does not apply to types [-Werror=attributes]
    2 | static [[gnu::always_inline]] inline void b(void);
      | ^~~~~~
inline.c:2:31: error: expected identifier or ‘(’ before ‘inline’
    2 | static [[gnu::always_inline]] inline void b(void);
      |                               ^~~~~~
inline.c:3:1: error: ‘always_inline’ attribute does not apply to types [-Werror=attributes]
    3 | static inline [[gnu::always_inline]] void c(void);
      | ^~~~~~
inline.c:3:38: error: expected identifier or ‘(’ before ‘void’
    3 | static inline [[gnu::always_inline]] void c(void);
      |                                      ^~~~
inline.c:4:1: error: ‘always_inline’ attribute does not apply to types [-Werror=attributes]
    4 | static inline void [[gnu::always_inline]] d(void);
      | ^~~~~~
inline.c:6:1: error: ‘always_inline’ attribute does not apply to types [-Werror=attributes]
    6 | static inline void f(void)[[gnu::always_inline]];
      | ^~~~~~
cc1: all warnings being treated as errors

### DIFF
--- a/auto/clang
+++ b/auto/clang
@@ -147,7 +147,7 @@ nxt_feature_incs=
 nxt_feature_libs=
 nxt_feature_test="#include <stdlib.h>
 
-                  void *f(void) __attribute__ ((__malloc__));
+                  __attribute__ ((__malloc__)) void *f(void);
 
                   void *f(void) {
                       return malloc(1);
@@ -183,7 +183,7 @@ nxt_feature_name=NXT_HAVE_GCC_ATTRIBUTE_UNUSED
 nxt_feature_run=
 nxt_feature_incs=
 nxt_feature_libs=
-nxt_feature_test="static void f(void) __attribute__ ((__unused__));
+nxt_feature_test="__attribute__ ((__unused__)) static void f(void);
 
                   static void f(void)
                   {
--- a/src/nxt_clang.h
+++ b/src/nxt_clang.h
@@ -8,7 +8,7 @@
 #define _NXT_CLANG_H_INCLUDED_
 
 
-#define nxt_inline     static inline __attribute__((always_inline))
+#define nxt_inline     __attribute__((always_inline)) static inline
 #define nxt_noinline   __attribute__((noinline))
 #define nxt_cdecl
 
--- a/src/nxt_conf_validation.c
+++ b/src/nxt_conf_validation.c
@@ -76,9 +76,8 @@ static nxt_int_t nxt_conf_vldt_error(nxt_conf_validation_t *vldt,
     const char *fmt, ...);
 static nxt_int_t nxt_conf_vldt_var(nxt_conf_validation_t *vldt, nxt_str_t *name,
     nxt_str_t *value);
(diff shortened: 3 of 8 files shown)


hongzhidao:
Thanks.

I know very little about C2x, but for now, it's not a required compiler, right?
```
C2x is an informal name for the next (after C17) major C language standard revision.[1][2] It is expected to be voted on in 2023 and would therefore be C23.[3]
```
Could you share why this change is made, from the perspective of value?
1. It has worked well and passed tests in many machine environments.
2. If it's changed, we have to review, test, and publish them, it increases time cost at least.

Since such submissions have been on the rise lately, I wonder about the benefits of doing so.
Let's discuss #722 together. The change is fine, but it was fine. Why change some code that has been working well for a long time?

Unit has many features not yet supported. For example:
1. I'd like to see it supports gzip in the file serving. It's a good feature for users who want to use file serving.
2. supporting rewrite feature.
3. supporting responding to simple text.
etc.

This is just my personal opinion, welcome to discuss :)

Issue #669, opened 2022-04-01
Problem when restarting unix socket as listener


echolimazulu:
**Hello Nginx Team and Contributors,**

This is a separation of the unix socket restart problem from the socket cleanup problem (https://github.com/nginx/unit/issues/643) when the program is closed, as these are technically and conceptually different problems.

**Issue:**
`bind(\"unix:/var/run/unit.sock\") failed (98: Address already in use)`

echolimazulu:
In the next hour there will be a PR for this request.

echolimazulu:
The solution to the problem is proposed in the **PR**:
https://github.com/nginx/unit/pull/670

echolimazulu:
**Repository to reproduce the issue:**
https://github.com/echolimazulu/unit-bind-aaiu

hongzhidao:
Hi, @echolimazulu 

Thanks for your patch.
Actually, it's much more complicated. It also happens during reconfiguration.
Just take a look at the case:
```
{
    "listeners": {
        "unix:/tmp/test.sock": {
            "pass": "routes"
        }
    },
    "routes": [
        {
            "action": {
                "return": 205
            }
        }
    ]
}
```
Then
```
# conf.json
{
    "pass": "routes"
}

> curl -X DELETE http://127.1:8000/config/listeners/unix:%2Ftmp%2Ftest.sock
> curl -X PUT -d@conf.json http://127.1:8000/config/listeners/unix:%2Ftmp%2Ftest.sock
It failed
```

echolimazulu:
Good morning @hongzhidao,

Thanks for the option you suggested.

Reproducing the specified list of commands with the patch (**PR:** https://github.com/nginx/unit/pull/670) applied, I get an error:
```
# cat conf.json
{
    "listeners": {
        "unix:/tmp/test.sock": {
            "pass": "routes"
        }
    },
    "routes": [
        {
            "action": {
                "return": 205
            }
        }
    ]
}

# curl -X PUT --data-binary @conf.json --unix-socket /var/run/control.unit.sock http://localhost/config/
{
	"success": "Reconfiguration done."
}

# curl -X DELETE --unix-socket /var/run/control.unit.sock http://localhost/config/listeners/unix:%2Ftmp%2Ftest.sock

# cat /var/lib/unit/conf.json
{"listeners":{},"routes":[{"action":{"return":205}}]}

[...]

PR #670, opened 2022-04-01
Socket: replacement of called functions, permissions, reusability

echolimazulu:
**Solution to the problem:**
https://github.com/nginx/unit/issues/669

**Repository to reproduce the issue:**
https://github.com/echolimazulu/unit-bind-aaiu

Hello @VBart, @mar0x, @hongzhidao, @alejandro-colomar, @tippexs.

Please tell me what you think. In the process of testing this fix, I no longer observe the previously encountered problem.

I tried to take into account all the previously voiced comments regarding the code, commits, indents and other points.

I did not propose the implementation of locks, cleaning sockets and locks when the program is closed, cleaning when updating the configuration, since this is a much larger number of changes. Addressing several different issues and extending functionality that is not needed for: https://github.com/nginx/unit/issues/669

PR #735, opened 2022-07-19
unlink(2) Unix sockets after use.

alejandro-colomar:


alejandro-colomar:
I'm getting EACCESS from unlink(2) :(

alejandro-colomar:
My guess is that the socket has been passed to a thread that doesn't have permission to unlink it.  Maybe we need to intervene from the main thread to unlink(2) when a thread tries to close(2) a socket?  Using BPF?  Or sending the socket back to the main thread and letting it close it?

src/nxt_socket.c
```diff
@@ -285,6 +285,20 @@ nxt_socket_close(nxt_task_t *task, nxt_socket_t s)
...
 
+#if (NXT_HAVE_UNIX_DOMAIN)
+    socklen_t           len;
+    struct sockaddr_un  sa = {0};
+
+    len = sizeof(struct sockaddr_un) - 1;
+    if ((getsockname(s, (struct sockaddr *)&sa, &len) == 0)
+        && (sa.sun_family == AF_UNIX))
+    {
+        ((char *) &sa)[len] = '\0';
```
ac000, line 296: The opening **{** should be on the preceding line I believe.

Also, won't sa.sun_len be null-terminated from the 0 initialisation of sa? Or do you fear some compilers won't get it right?

I take it the -1 on the length was to do with null-terminating sa.sun_path?

src/nxt_socket.c
```diff
@@ -285,6 +285,20 @@ nxt_socket_close(nxt_task_t *task, nxt_socket_t s)
...
+        nxt_debug(task, "socket unlink(%s)", sa.sun_path);
+        (void) unlink(sa.sun_path);
```
ac000, line 298: Right, the problem here is that it seems to be the router process doing the unlink(2) and when unit is running as root the router process is running as an unprivileged user (nobody here).

I wonder if it would be possible to setup a callback function like there is for the isolation or cgroup stuff for cleaning things up when an application is terminated. The socket is created in the main unit process so hopefully it still has the details...

Or chown(2) the socket to the unprivileged user... the sockets are being created 666 regardless (here anyway).

alejandro-colomar: Re: callback:  If the file is owned by root, then only the main process can remove it.  I thought about sending the sfd back to the main process so that it's the main process the one that unlink(2)s and close(2)s it.  But that's not an easy thing.  I'd prefer to avoid that.

Re: chown(2):  Hmm, I like the idea.  If it's currently 666 I guess there's no added insecurity by chown(2)ing it to nobody.  The directory permissions are a problem, though.

alejandro-colomar: Re: callback:  If the file is owned by root, then only the main process can remove it.  I thought about sending the sfd back to the main process so that it's the main process the one that unlink(2)s and close(2)s it.  But that's not an easy thing.  I'd prefer to avoid that, if we can.

Re: chown(2):  Hmm, I like the idea.  If it's currently 666 I guess there's no added insecurity by chown(2)ing it to nobody.  The directory permissions are a problem, though.

I don't know...  is there any way to tell the kernel: "I want to unlink this file, but don't remove it from the filesystem until the last fd has been closed"?

Maybe the answer to that is using abstract sockets in Linux, and leaving other systems with non-unlinked sockets?

src/nxt_socket.c
alejandro-colomar: From the coding style <https://nginx.org/en/docs/dev/development_guide.html#code_style>:

> Conditionals and Loops
> 
> The “if” keyword is separated from the condition by one space. Opening brace is located on the same line, or on a dedicated line if the condition takes several lines.

alejandro-colomar: Re: NUL-termination:  Oh yes, that's a glitch.  I first terminated it by one method, then changed it, and forgot to remove the old code.

My concern with `{0}` is that I'm not sure if I can trust the kernel not writing extra bytes with garbage, as I've never had to use getsockname(2) before and am not used to it.  I guess that letting garbage into user-space data could be a security risk and probably the kernel doesn't do that, but wanted to be sure.  So, can you confirm that I can just remove the `... = '\0';` assignment?

The behavior of `getsockname(2)` regarding the string should be documented in the manual page.  I read it several times and it says nothing about it.  Could you please send a patch if you know how it works?

src/nxt_socket.c
alejandro-colomar: I guess you referred to abstract sockets the other day, right?  For anyone else reading this, see unix(7):

```
DESCRIPTION
       [...]
   Address format
       [...]

       Three  types  of  address  are distinguished in the sock‐
       addr_un structure:

       [...]

       *  abstract: an abstract socket address is  distinguished
          (from  a pathname socket) by the fact that sun_path[0]
          is a null byte ('\0').  The socket’s address  in  this
          namespace is given by the additional bytes in sun_path
          that  are  covered  by the specified length of the ad‐
          dress structure.  (Null bytes in the name have no spe‐
          cial significance.)  The name has no  connection  with
          filesystem pathnames.  When the address of an abstract
          socket  is  returned,  the returned addrlen is greater
          than sizeof(sa_family_t) (i.e., greater than  2),  and
          the  name of the socket is contained in the first (ad‐
          drlen - sizeof(sa_family_t)) bytes of sun_path.

       [...]

   Abstract sockets
       Socket permissions have no meaning for abstract  sockets:
       the  process  umask(2)  has no effect when binding an ab‐
       stract socket, and changing the ownership and permissions
       of the object (via fchown(2) and fchmod(2)) has no effect
       on the accessibility of the socket.

       Abstract sockets automatically disappear  when  all  open
       references to the socket are closed.

       The  abstract socket namespace is a nonportable Linux ex‐
       tension.

[...]

VERSIONS
       SCM_CREDENTIALS  and  the  abstract namespace were intro‐
       duced with Linux 2.2 and should not be used  in  portable
       programs.  (Some BSD‐derived systems also support creden‐
       tial passing, but the implementation details differ.)
```

alejandro-colomar:
This PR attempts to fix #669, and attempts to be a simpler alternative implementation of #670 .

src/nxt_socket.c
ac000: RE if. Ah, I'll need to change some of my if's that have been combined and now fall into that category...

Actually, I didn't think to look at the nginx coding style. The unit CONTRIBUTING.md does say this

> please read the NGINX Unit code guidelines to know more about coding conventions

Is that actually referring to the nginx coding style document?

Regarding

> My concern with {0} is that I'm not sure if I can trust the kernel not writing extra bytes with garbage, as I've never had to use getsockname(2) before and am not used to it.

I would certainly hope it doesn't. On Linux I think everything is good, from reading the unix(7) man page, you get back a null-terminated sun_path.

However the BUGS section has a caveat to that in that if you supplied a pathname of 108 characters with no nul byte, the kernel will add one which will exceed sizeof(sun_path) and depending on how you set addrlen you may not get back the nul byte. More interestingly is what it says about other implementations where they don't require a null-terminated string when binding a socket in which case it basically suggests doing what you've done ;).

Hmm, I wonder if we do any checking on the path length when creating the socket...

alejandro-colomar: Re: unix(7):  Hmm, it seems (although the text isn't entirely clear, that's what I infer) that systems respect the NUL byte if you pass it, so as long as we guarantee that we create sockets with a terminating NUL byte, I don't need to care at all about it.

src/nxt_socket.c
ac000: > Re: callback: If the file is owned by root, then only the main process can remove it. I thought about sending the sfd back to the main process so that it's the main process the one that unlink(2)s and close(2)s it. But that's not an easy thing. I'd prefer to avoid that, if we can.

Yes, what I was hoping was that you might be able to do something like what the isolation and cgroup code do by hooking into the nxt_main_process_cleanup() function for when an application terminates. This is run from the main unit process. E.g

```
$ sudo /opt/unit/sbin/unitd --control 127.0.0.1:8080 --no-daemon
2022/07/20 17:13:27 [info] 5220#5220 unit 1.28.0 started
^C[5220] Cleaning up cgroups for (5255/"sqroot" prototype)
[5220] Cleaning up cgroups for (5224/"add" prototype)
[5220] Cleaning up cgroups for (5247/"divide" prototype)
[5220] Cleaning up cgroups for (5229/"multiply" prototype)
[5220] Cleaning up cgroups for (5226/"subtract" prototype)
```

In the socket case I guess you'd want to do it at reconfigure/shutdown time. Now I don't know if a facility already exists or if it might not be too much work to add one, just another idea...

> I don't know... is there any way to tell the kernel: "I want to unlink this file, but don't remove it from the filesystem until the last fd has been closed"?

Not as far as I know, the closest paradigm I can think of is the open file, unlink it (so it **can't** be easily accessed) and then close. But of course that doesn't help here...

> Maybe the answer to that is using abstract sockets in Linux, and leaving other systems with non-unlinked sockets?

Heh, that would certainly nicely solve the problem on Linux!

src/nxt_socket.c
ac000: Yes, that is my understanding also. Linux will add one if it's missing, some others won't, but if you put one, then it _should_ stay there.

alejandro-colomar:
Better fixed by #740

Issue #742, opened 2022-08-05
The $runstatedir is not being created at runtime


alejandro-colomar:
See the following configuration.  I used a prefix of `/opt/local/unit` to make sure there were no previous existing directories or stuff that was left by a previous unit installation, and so that I can wipe it easily, but this could perfectly use system paths.

The issue is that the system is allowed (and will normally do so) to wipe `/run` (or `/var/run`) at [re]boot time.  Since we use various files in /run (at least, the pid file and the control socket), the FHS says that we should use a directory within /run:

<https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#runRuntimeVariableData>
> Programs may have a subdirectory of /run; this is encouraged for programs that use more than one run-time file.

```
alx@asus5775:~/src/nginx/unit$ ./configure --prefix=/opt/local/unit \
                                           --modules=/opt/local/unit/lib/unit/modules \
                                           --state=/opt/local/unit/var/lib/unit \
                                           --pid=/opt/local/unit/var/run/unit/unit.pid \
                                           --log=/opt/local/unit/var/log/unit/unit.log \
                                           --control=unix:/opt/local/unit/var/run/unit/control.unit.sock \
                               | tail -n 27
Unit configuration summary:

  bin directory: ............. "/opt/local/unit/bin"
  sbin directory: ............ "/opt/local/unit/sbin"
  lib directory: ............. "/opt/local/unit/lib"
  include directory: ......... "/opt/local/unit/include"
  man pages directory: ....... "/opt/local/unit/share/man"
  modules directory: ......... "/opt/local/unit/lib/unit/modules"
  state directory: ........... "/opt/local/unit/var/lib/unit"
  tmp directory: ............. "/opt/local/unit/tmp"

  pid file: .................. "/opt/local/unit/var/run/unit/unit.pid"
  log file: .................. "/opt/local/unit/var/log/unit/unit.log"

  control API socket: ........ "unix:/opt/local/unit/var/run/unit/control.unit.sock"

  non-privileged user: ....... "nobody"
  non-privileged group: ...... ""

  IPv6 support: .............. YES
  Unix domain sockets support: YES
  TLS support: ............... NO

  process isolation: ......... USER NS PID NET UTS CGROUP

  debug logging: ............. NO

alx@asus5775:~/src/nginx/unit$ make -j >/dev/null; echo $?
0
alx@asus5775:~/src/nginx/unit$ sudo make install
install -d /opt/local/unit/sbin
[...]

PR #743, opened 2022-08-05
Added default values for pathnames.

alejandro-colomar:
This allows one to simply run `./configure` and expect it to
produce sane defaults for an install.

Previously, without specifying `--prefix=...`, `make install`
would simply fail, recommending to set `--prefix` or `DESTDIR`,
but that recommendation was incomplete at best, since it didn't
set many of the subdirs needed for a good organization.

Setting `DESTDIR` was even worse, since that shouldn't even affect
an installation (it is required to be transparent to the
installation).

/usr/local is the historic Unix standard path to use for
installations from source made manually by the admin of the
system.  Some package managers (Homebrew, I'm looking specifically
at you) have abused that path to install their things, but 1) it's
not our fault that someone else incorrectly abuses that path (and
they seem to be fixing it for newer archs; e.g., they started
using /opt/homebrew for Apple Silicon), 2) there's no better path
than /usr/local, 3) we still allow changing it for systems where
this might not be the desired path (MacOS Intel with hombrew), and
4) it's _the standard_.

See a related conversation with Ingo (OpenBSD maintainer):

On 7/27/22 16:16, Ingo Schwarze wrote:
> Hi Alejandro,
[...]
>
> Alejandro Colomar wrote on Sun, Jul 24, 2022 at 07:07:18PM +0200:
>> On 7/24/22 16:57, Ingo Schwarze wrote:
>>> Alejandro Colomar wrote on Sun, Jul 24, 2022 at 01:20:46PM +0200:
>
>>>> /usr/local is for sysadmins to build from source;
>
>>> Doing that is *very* strongly discouraged on OpenBSD.
>
>> I guess that's why the directory was reused in the BSDs to install ports
>> (probably ports were installed by the sysadmin there, and by extension,
>> ports are now always installed there, but that's just a guess).
>
> Maybe.  In any case, the practice of using /usr/local for packages
> created from ports is significantly older than the recommendation
> to refrain from using upstream "make install" outside the ports
> framework.
>
>   * The FreeBSD ports framework was started by Jordan Hubbard in 1993.
>   * The ports framework was ported from FreeBSD to OpenBSD
>     by Niklas Hallqvist in 1996.
>   * NetBSD pkgsrc was forked from FreeBSD ports by Alistair G. Crooks
>     and Hubert Feyrer in 1997.
>
> I failed to quickly find Jordan's original version, but rev. 1.1
> of /usr/ports/infrastructure/mk/bsd.port.mk in OpenBSD (dated Jun 3
> 22:47:10 1996 UTC) already said
>
>    LOCALBASE ?= /usr/local
>    PREFIX    ?= ${LOCALBASE}
>
[...]
>> I had a discussion in NGINX Unit about it, and
>> the decission for now has been: "support prefix=/usr/local for default
>> manual installation through the Makefile, and let BSD users adjust to
>> their preferred path".
>
> That's an *excellent* solution for the task, thanks for doing it
> the right way.  By setting PREFIX=/usr/local by default in the
> upstream Makefile, you are minimizing the work for *BSD porters.
>
> The BSD ports frameworks will typically run the upstreak "make install"
> with the variable DESTDIR set to a custom value, for example
>
>    DESTDIR=/usr/ports/pobj/groff-1.23.0/fake-amd64
>
> so if the upstream Makefile sets PREFIX=/usr/local ,
> that's perfect, everything gets installed to the right place
> without an intervention by the person doing the porting.
>
> Of course, if the upstream Makefile would use some other PREFIX,
> that would not be a huge obstacle.  All we have to do in that case
> is pass the option --prefix=/usr/local to the ./configure script,
> or something equivalent if the software isn't using GNU configure.
>
>> We were concerned that we might get collisions
>> with the BSD port also installing in /usr/local, but that's the least
>> evil (and considering BSD users don't typically run `make install`, it's
>> not so bad).
>
> It's not bad at all.  It's perfect.
>
> Of course, if a user wants to install *without* the ports framework,
> they have to provide their own --prefix.  But that's not an issue
> because it is easy to do, and installing without a port is discouraged
> anyway.

===

Directory variables should never contain a trailing slash (I've
learned that the hard way, where some things would break
unexpectedly).  Especially, make(1) is likely to have problems
when things have double slashes or a trailing slash, since it
treats filenames as text strings.  I've removed the trailing slash
from the prefix, and added it to the derivate variables just after
the prefix.  pkg-config(1) also expects directory variables to have
no trailing slash.

===

I also removed the code that would set variables as depending on
the prefix if they didn't start with a slash, because that is a
rather non-obvious behavior, and things should not always depend
on prefix, but other dirs such as $(runstatedir), so if we keep
a similar behavior it would be very unreliable.  Better keep
variables intact if set, or use the default if unset.

===

Since now the variables are set only after parsing the options,
the auto/help script call needs to be delayed until after setting
the variables.

===

I used a subdirectory under the standard /var/lib for NXT_STATE,
instead of a homemade "state" dir that does the same thing.

===

Modified the Makefile to create some dirs that weren't being
created, and also remove those that weren't being removed in
uninstall, probably because someone forgot to add them.

===

A summary of the default config is:

Unit configuration summary:

  bin directory: ............. "/usr/local/bin"
  sbin directory: ............ "/usr/local/sbin"
  lib directory: ............. "/usr/local/lib"
  include directory: ......... "/usr/local/include"
  man pages directory: ....... "/usr/local/share/man"
  modules directory: ......... "/usr/local/lib/unit/modules"
  state directory: ........... "/usr/local/var/lib/unit"
  tmp directory: ............. "/usr/local/tmp"

  pid file: .................. "/usr/local/var/run/unit/unit.pid"
  log file: .................. "/usr/local/var/log/unit/unit.log"

  control API socket: ........ "unix:/usr/local/var/run/unit/control.unit.sock"

===

Following GNU coding standards and the Filesystem Hierarchy Standard:
<https://www.gnu.org/prep/standards/html_node/Directory-Variables.html>
<https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html>


alejandro-colomar:
This cannot be merged before fixing <https://github.com/nginx/unit/issues/742>, since it would create a broken default.

alejandro-colomar:
Rebased to master.

alejandro-colomar:
Rebased to master.  Added tags to commit logs.  Added some discussion to the second commit log: `2fdbaaaa Ensured $NXT_TMP is created at 'make install'`.

alejandro-colomar:
I'm closing this.  I'll push the updated patches to RB.

alejandro-colomar:
Merged.

PR #746, opened 2022-08-10
Add support for a user configuration file.

alejandro-colomar:
Add support for a user-specified user configuration file, normally
under $sysconfdir, that is, `/etc`.  I'd recommend using `/etc/unit/unitd.conf.json` for the default pathname.

The part for setting the file's default pathname, and for changing it, will be implemented in the auto
scripts, but since it will depend on the PR that adds a default
prefix, that code is not yet written in this commit.  TODO.

### DIFF
--- a/src/nxt_controller.c
+++ b/src/nxt_controller.c
@@ -174,6 +174,12 @@ nxt_controller_prefork(nxt_task_t *task, nxt_process_t *process, nxt_mp_t *mp)
     nxt_conf_ver = 12500;
 
     ret = nxt_controller_file_read(task, rt->conf, &ctrl_init.conf, mp);
+
+    if (nxt_slow_path(ret == NXT_DECLINED)) {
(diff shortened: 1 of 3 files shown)


ac000:
I wonder if it even needs to be a file? Or could it just be stored internally to unit? That way users won't go thinking they can or should be editing it.

Or if it needs to be an actual file, maybe sticking it under the docroot e.g /usr/share/doc/unit would be better...

alejandro-colomar:
> I wonder if it even needs to be a file? Or could it just be stored internally to unit? That way users won't go thinking they can or should be editing it.

I think a real file would make it simpler.

> 
> Or if it needs to be an actual file, maybe sticking it under the docroot e.g /usr/share/doc/unit would be better...

Maybe.  But then I would certainly add a flag to enable reading that file, since noone expects that a file in $docdir would change behavior of a program.

alejandro-colomar:
<http://www.alejandro-colomar.es/src/alx/nginx/unitcli.git/tree/bin/setup-unit>

PR #747, opened 2022-08-15
Socket: Created control socket & pid file directories.

ac000:
@alejandro-colomar reported an issue on GitHub whereby Unit would fail
to start due to not being able to create the control socket (a Unix
Domain Socket)

  2022/08/05 20:12:22 [alert] 21613#21613 bind(6,
  unix:/opt/local/unit/var/run/unit/control.unit.sock.tmp) failed (2:
  No such file or directory)

This could happen if the control socket was set to a directory that
doesn't exist. A common place to put the control socket would be under
/run/unit, and while /run will exist, /run/unit may well not (/run
is/should be cleared on each boot).

The pid file would also generally go under /run/unit, though this is
created after the control socket, however it could go someplace else so
we should also ensure its directory exists.

This commit ensures that the full directory paths to where the control
socket and pid file will be located exists before trying to create them.

Closes: https://github.com/nginx/unit/issues/742
Reported-by: Alejandro Colomar <alx.manpages@gmail.com>
Reviewed-by: Alejandro Colomar <alx.manpages@gmail.com>
Tested-by: Alejandro Colomar <alx.manpages@gmail.com>
Signed-off-by: Andrew Clayton <andrew@digital-domain.net>

### DIFF
--- a/src/nxt_fs.c
+++ b/src/nxt_fs.c
@@ -273,6 +273,31 @@ nxt_fs_mkdir_all(const u_char *dir, mode_t mode)
 }
 
 
+nxt_int_t
+nxt_fs_mkdir_parent(const u_char *path, mode_t mode)
+{
+    char       *ptr, *dir;
+    nxt_int_t  ret;
+
+    dir = nxt_strdup(path);
+    if (nxt_slow_path(dir == NULL)) {
+        return NXT_ERROR;
+    }
+
+    ret = NXT_OK;
+
+    ptr = strrchr(dir, '/');
+    if (nxt_fast_path(ptr != NULL)) {
+        *ptr = '\0';
+        ret = nxt_fs_mkdir((const u_char *) dir, mode);
+    }
+
+    nxt_free(dir);
+
+    return ret;
+}
+
+
 static nxt_int_t
 nxt_fs_mkdir(const u_char *dir, mode_t mode)
 {
--- a/src/nxt_controller.c
+++ b/src/nxt_controller.c
@@ -666,6 +666,14 @@ nxt_runtime_controller_socket(nxt_task_t *task, nxt_runtime_t *rt)
 #endif
     ls->handler = nxt_controller_conn_init;
 
+#if (NXT_HAVE_UNIX_DOMAIN)
+    if (ls->sockaddr->u.sockaddr.sa_family == AF_UNIX) {
+        const char *path = ls->sockaddr->u.sockaddr_un.sun_path;
+
+        nxt_fs_mkdir_parent((const u_char *) path, 0755);
+    }
+#endif
+
     if (nxt_listen_socket_create(task, rt->mem_pool, ls) != NXT_OK) {
         return NXT_ERROR;
     }
--- a/src/nxt_fs.h
+++ b/src/nxt_fs.h
@@ -36,6 +36,7 @@ typedef struct {
 } nxt_fs_mount_t;
 
 
+nxt_int_t nxt_fs_mkdir_parent(const u_char *path, mode_t mode);
 nxt_int_t nxt_fs_mkdir_all(const u_char *dir, mode_t mode);
 nxt_int_t nxt_fs_mount(nxt_task_t *task, nxt_fs_mount_t *mnt);
 void nxt_fs_unmount(const u_char *path);
--- a/src/nxt_runtime.c
+++ b/src/nxt_runtime.c
@@ -1369,6 +1369,8 @@ nxt_runtime_pid_file_create(nxt_task_t *task, nxt_file_name_t *pid_file)
 
     file.name = pid_file;
 
+    nxt_fs_mkdir_parent(pid_file, 0755);
+
     n = nxt_file_open(task, &file, O_WRONLY, O_CREAT | O_TRUNC,
                       NXT_FILE_DEFAULT_ACCESS);
 


ac000:
I've included a bunch of tags at the end this commit to _demonstrate_ what
they would look like, they can be removed before committing...

The tags serve several functions, including

1) Signed-off-by is the main one and says that the author (and the person(s)
   who handle the patch on its way to the repotitory) have the right to submit
   the patch under the terms of the projects license.
   
2) Give credit to contributors who review code, report bugs, test patches etc.

3) Provide an audit trail of how a patch made it into the project.

alejandro-colomar (approved):

hongzhidao:
In my feelings, in the function `nxt_socket_bind()`, there are a few places that are worth reworking.
```
#if NXT_HAVE_UNIX_DOMAIN
    if (sa->u.sockaddr.sa_family == AF_UNIX) {
        char  *ptr;
        char  *socket_dir;

        socket_dir = strdup(sa->u.sockaddr_un.sun_path);
        if (nxt_slow_path(socket_dir == NULL)) {
            return NXT_ERROR;
        }

        ptr = strrchr(socket_dir, '/');
        *ptr = '\0';

        nxt_fs_mkdir_all((const u_char *) socket_dir, 0777);

        nxt_free(socket_dir);
    }
#endif

```
1. We usually use wrapper functions instead of C native functions if possible, such as `strchr()` to `nxt_strchr()`.
2. It's possible that calling `nxt_fs_mkdir_all()` failed.
3. What about abstract unix socket?
4. This will make `nxt_socket_bind()` dependent on `nxt_fs_mkdir_all()`. Or say that making `socket` dependent on `file`. It seems it's unreasonable.  And users will be possible to create directories by reconfiguration.

alejandro-colomar:
> In my feelings, in the function `nxt_socket_bind()`, there are a few places that are worth reworking.
> 
> ```
> #if NXT_HAVE_UNIX_DOMAIN
>     if (sa->u.sockaddr.sa_family == AF_UNIX) {
>         char  *ptr;
>         char  *socket_dir;
> 
>         socket_dir = strdup(sa->u.sockaddr_un.sun_path);
>         if (nxt_slow_path(socket_dir == NULL)) {
>             return NXT_ERROR;
>         }
> 
>         ptr = strrchr(socket_dir, '/');
>         *ptr = '\0';
> 
>         nxt_fs_mkdir_all((const u_char *) socket_dir, 0777);
> 
>         nxt_free(socket_dir);
>     }
> #endif
> ```
> 
>     1. We usually use wrapper functions instead of C native functions if possible, such as `strchr()` to `nxt_strchr()`.

We use the wrappers when we use `nxt_str_t`, but in this case we're using "native" C strings, so it is simpler to use the native libc functions.

> 
>     2. It's possible that calling `nxt_fs_mkdir_all()` failed.

True, although if it fails, the next bind(2) is necessarily going to fail.  We can safely omit the error check, I think.

> 
>     3. What about abstract unix socket?

I don't remember, but I think we don't support abstract unix sockets in the control socket.  Do we?

> 
>     4. This will make `nxt_socket_bind()` dependent on `nxt_fs_mkdir_all()`. Or say that making `socket` dependent on `file`. It seems it's unreasonable.  And users will be able to create directories by reconfiguration.

I don't fully understand this.  Creating a socket in `/foo/bar/sock` necessarily requires that we can create the `/foo/bar/` directory.  Otherwise the socket creation will fail.

Yes, users will be able to create directories by reconfiguring, but that's why we strongly recommend that the control socket is only r/w by root.

hongzhidao:
> 4. This will make `nxt_socket_bind()` dependent on `nxt_fs_mkdir_all()`. Or say that making `socket` dependent on `file`. It seems it's unreasonable.  And users will be able to create directories by reconfiguration.
I don't fully understand this. 

I mean every function should only do its job as what the function name is. The `nxt_socket_bind()` should only do socket bind, but not create directories.

hongzhidao:
> 2. It's possible that calling `nxt_fs_mkdir_all()` failed.
True, although if it fails, the next bind(2) is necessarily going to fail. We can safely omit the error check, I think.

That `nxt_socket_bind()` will depend on the outside again. I prefer to make functions individual and reliable.

alejandro-colomar:
> > 4. This will make `nxt_socket_bind()` dependent on `nxt_fs_mkdir_all()`. Or say that making `socket` dependent on `file`. It seems it's unreasonable.  And users will be able to create directories by reconfiguration.
> >    I don't fully understand this.
> 
> I mean every function should only do its job as what the function name is. The `nxt_socket_bind()` should only do socket bind, but not create directories.

Then it might make sense to add a separate function that creates the dir and is called always before `nxt_socket_bind()`.  But yes, they need to be called together, so I'd just make it depend on mkdir().

hongzhidao:
> 3. What about abstract unix socket?
I don't remember, but I think we don't support abstract unix sockets in the control socket. Do we?

`nxt_socket_bind()` is a generic function, We can't assume that someone won't call it.

alejandro-colomar:
What bind does is bind a socker file descriptor to a filesystem path.  That filesystem path includes a file and its parent directory.  I don't think creating a file is too different from creating a full path (dir + file).  So I don't think it is really a different operation.

hongzhidao:
Just to wonder if DevOps can do the thing of creating directories by themself?

alejandro-colomar:
> Just to wonder if DevOps can do the thing of creating directories by themself?

Do you mean manually create the directory?

It's not reasonable, because at reboot, the OS destroys everything under `/run`, so it would require that the directory is manually recreated at every reboot.

hongzhidao:
> What bind does is bind a socket file descriptor to a filesystem path. That filesystem path includes a file and its parent directory. I don't think creating a file is too different from creating a full path (dir + file). So I don't think it is really a different operation.

What if a system administrator only wants the socket to run in a specific directory?

alejandro-colomar:
> > What bind does is bind a socket file descriptor to a filesystem path. That filesystem path includes a file and its parent directory. I don't think creating a file is too different from creating a full path (dir + file). So I don't think it is really a different operation.
> 
> What if a system administrator only wants the socket to run in a specific directory?


Sockets should be created under some subdirectory of `/run`.  The sysadmin should specify the specific directory that s/he wants in the unit config.

alejandro-colomar:
If the sysadmin wants to limit the ability of unitd to create certain directories, it can limit the permissions of the user that will run it, so that it's not root.  If unitd is run under root, it will have root privileges.  So, normal filesystem permission situation.  Nothing dangerous, as I see it.

hongzhidao:
From a function design point of view, why is `nxt_socket_bind()` able to create any directories if possible?
It should only do socket bind as its function name.

ac000:
> > 3. What about abstract unix socket?
> 
> I don't remember, but I think we don't support abstract unix sockets in the control socket. Do we?

No, I'm pretty sure it doesn't support being an abstract socket, it uses a different code path to the listeners.

hongzhidao:
> Actually, let me double check that...

It's better to make sure if it's reasonable to create directories in `nxt_socket_bind()`.

hongzhidao:
I found some information I didn't know before.
https://github.com/nginx/unit/issues/742
```
alx@asus5775:~/src/nginx/unit$ sudo /opt/local/unit/sbin/unitd 
2022/08/05 20:12:22 [alert] 21613#21613 bind(6, unix:/opt/local/unit/var/run/unit/control.unit.sock.tmp) failed (2: No such file or directory)
unitd fails to create the socket (because it assumes the parent dir exists always), and exits.

unitd should create the parent directory previous to attempting to bind(2) the socket.
```

1. > unitd fails to create the socket (because it assumes the parent dir exists always), and exits.

I think that's a reasonable limit, unit is not responsible for managing related unix socket directory, it only works with UNIX socket file.

> unitd should create the parent directory previous to attempting to bind(2) the socket.

Just to think that if it's a reasonable requirement to help users create any socket directory internally?

hongzhidao:
Let's separate it into two requirements.

1. If it reasonable to limit the creation of socket directory? For example:
```
{
       "listeners": {
              "unix:/a/b/c/some.sock": {}
       }
}
```
If it has root privilege, the above configuration is valid or not? I prefer invalid. unix is not responsible for managing unix socket directory.

2. If it impossible to meet your needs without creating a directory internally? Is there other way to do it?

alejandro-colomar:
> Let's separate it into two requirements.
> 
>     1. If it reasonable to limit the creation of socket directory? For example:
> 
> 
> ```
> {
>        "listeners": {
>               "unix:/a/b/c/some.sock": {}
>        }
> }
> ```

We're not (yet) modifying that, right?  @ac000 this change doesn't affect listeners, right?

I think this patch only changes the behavior for the control socket (because it's the one that blocked my patch, but I think we should also allow that in the long term).

> 
> If it has root privilege, the above configuration is valid or not? I prefer invalid. unix is not responsible for managing unix socket directory.

It should be valid for root. The Filesystem Hierarchy Standard requires that socket files should be created under an app-specific directory of `/run`:

<https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s15.html>

And `/run` is nuked by the OS at reboot, so it's impossible (except by using rc init scripts, systemd, or a docker entrypoint, depending on your system) to create the directory outside of unitd.  I think it's our responsibility to create the directory.

> 
>     2. If it impossible to meet your needs without creating a directory internally? Is there other way to do it?

The other ways are writing that in the same init scripts that run unitd.

hongzhidao:
> We're not (yet) modifying that, right? @ac000 this change doesn't affect listeners, right?

For devs, `nxt_socket_bind()` is a generic function, we can't predict who uses it or not.

alejandro-colomar:
> > We're not (yet) modifying that, right? @ac000 this change doesn't affect listeners, right?
> 
> For devs, `nxt_socket_bind()` is a generic function, we can't predict who uses it or not.

Whoever needs to run that function should also want to run mkdir.  We're not creating a problem to users of the function.  We're just making life easier to them.

ac000:
> > Let's separate it into two requirements.
> > ```
> > 1. If it reasonable to limit the creation of socket directory? For example:
> > ```
> > ```
> > {
> >        "listeners": {
> >               "unix:/a/b/c/some.sock": {}
> >        }
> > }
> > ```
> 
> We're not (yet) modifying that, right? @ac000 this change doesn't affect listeners, right?

Correct.

> I think this patch only changes the behavior for the control socket (because it's the one that blocked my patch, but I think we should also allow that in the long term).

Yes, this change only affects the control socket.

ac000:
OK, so after some code splunking, I have found the following...

In unit there are currently two calls to bind(2)

Controller socket bind(2) path
```
main()
  nxt_runtime_create()
    nxt_runtime_conf_init()
      nxt_runtime_controller_socket()
        nxt_listen_socket_create()
          nxt_socket_bind()			<== src/nxt_socket.c
            bind(2)
```

Listen socket bind(2) path
```
main()
  nxt_runtime_create()
    nxt_runtime_initial_start()
      nxt_main_process_start()
        nxt_main_process_port_create()
          nxt_port_enable()
            nxt_main_port_socket_handler()
              nxt_main_listening_socket()	<== src/nxt_main_porcess.c
                bind(2)
```

There are two calls to nxt_socket_bind(), the above one and
```
[
#define nxt_conn_connect(engine, c)                                           \ 
    nxt_work_queue_add(&engine->socket_work_queue, nxt_conn_sys_socket,       \ 
                       c->socket.task, c, c->socket.data) 
nxt_conn_sys_socket()
]
  nxt_conn_socket()
    nxt_socket_bind()

$ grep -rn "nxt_conn_connect(" src/*.c
src/nxt_conn_proxy.c:117:        nxt_conn_connect(task->thread->engine, peer);
src/nxt_conn_proxy.c:224:    nxt_conn_connect(task->thread->engine, p->peer);
src/nxt_conn_proxy.c:846:    nxt_conn_connect(task->thread->engine, peer);
src/nxt_h1proto.c:2192:    nxt_conn_connect(task->thread->engine, c);
```

So it looks like nxt_socket_bind() is also used for connecting to proxies and HTTP CONNECTs. I'm guessing at least some of them _could_ be Unix domain sockets...

What I think I'll do is factor out the directory making code into a helper function, make the pid file code use it and also call it in nxt_runtime_controller_socket() for the control socket.

ac000:
This version factors out the directory making code into a new nxt_fs_mkdir_dirname() function in src/nxt_fs.c

I've re-introduced the check for if we find a '/' in the path as this function may be used more generally and it will return NXT_OK in such cases, assuming that it's the current working directory that's being used which will almost certainly exist.

@alejandro-colomar I've removed your Tested-by and S-o-b tags as this has been somewhat reworked...

hongzhidao:
Just to wonder if these automatically created directories need to be cleaned up if unitd quits?

alejandro-colomar:
> Just to wonder if these automatically created directories need to be cleaned up after unitd quits?

No, they need not be cleaned up.  The OS nukes `/run` at each boot, which is the reason we need to create them.  Also, trailing directories are usually not a problem (unless they have the same name as a file that is trying to be created, but that would be a problem of the person that organized that, not ours.

alejandro-colomar:
https://linux.codidact.com/posts/286875

alejandro-colomar:
Regarding the trailing directories: it's a minor issue, especially in the standard path, which is `/run`, since it's nuked at reboot.

One might say that in non-standard paths that may be a problem, but usually, having trailing directories is not a big issue.

Still, it may be an issue for some.  But then, those "some" shouldn't be asking unit to use such a path (e.g., `/bin/unit/control.unit.sock`).

But, since we don't cleanup (and I will argue that we shouldn't either) socket files, it would be nonesense to cleanup dirs.  `--control=/bin/control.unit.sock` would be as brain-damaged as this, or even more, and that is already possible with current unit.  We shouldn't think about what brain-damaged users can do with our program, as long as it's not introducing a serious vulnerability.

Other software, such as avahi, also create the runtime directory for the pid file and a socket, as this patch is doing.  So it looks good to me.

`Reviewed-by: Alejandro Colomar <alx.manpages@gmail.com>`

I'll add my tested-by tag soon when I test it.

src/nxt_fs.c
```diff
@@ -273,6 +273,32 @@ nxt_fs_mkdir_all(const u_char *dir, mode_t mode)
...
+nxt_int_t
+nxt_fs_mkdir_dirname(const u_char *path, mode_t mode)
+{
+    int   err;
+    char  *ptr;
+    char  *dir;
+
+    dir = nxt_strdup(path);
+    if (nxt_slow_path(dir == NULL)) {
+        return NXT_ERROR;
```
alejandro-colomar, line 285: Small issue; you forgot to catch errors.  It is caught here, but ignored later outside of `nxt_fs_mkdir_dirname()`.

alejandro-colomar: Or was it on purpose, just because a failure here doesn't really matter?

ac000: Yes, catch the memory error rather than segfault. But if we don't actually create the directory then that is no different to previously.

ac000: So in src/nxt_runtime.c::nxt_runtime_exit() there is

```C
    if (rt->type == NXT_PROCESS_MAIN) {                                         
        if (rt->pid_file != NULL) {                                             
            nxt_file_delete(rt->pid_file);                                      
        }                                                                       
                                                                                
#if (NXT_HAVE_UNIX_DOMAIN)                                                      
        {                                                                       
            nxt_sockaddr_t   *sa;                                               
            nxt_file_name_t  *name;                                             
                                                                                
            sa = rt->controller_listen;                                         
                                                                                
            if (sa->u.sockaddr.sa_family == AF_UNIX) {                          
                name = (nxt_file_name_t *) sa->u.sockaddr_un.sun_path;          
                (void) nxt_file_delete(name);                                   
            }                                                                   
        }                                                                       
#endif                                                                          
    }
```

This removes the pid file and control socket.

I could simply create an analogous function to nxt_fs_mkdir_dirname, e.g nxt_fs_rmdir_dirname, that would delete the directories (yes, they will often be the same)...

**However**, I am now convinced that we should **not** do that.

In the cgroups patch I do try and clean up cgroup directories

1) These can change from config to config.
2) I know at which point I should stop removing directories, e.g '/sys/fs/cgroup'

In the control socket/pid file case

1) These should not be changing.
2) I don't really know at which point in the file system hierarchy I should stop removing directories.

For (2) in the case of `/run/unit/control.sock` that could result in three rmdir(2)'s
```
rmdir("/run/unit");
rmdir("/run");
rmdir("/");
```
Of course no real harm will likely be done in this case as '/run' and '/' are most likely not going to be empty, but it just seems a little wrong to try.

On a side note, it's a pity that `rt->listen_sockets->elts` seems to be empty
```gdb
(gdb) p *(nxt_listen_socket_t *)rt->listen_sockets->elts
$3 = {
  socket = 0,
  backlog = 0,
  work_queue = 0x0,
  handler = 0x0,
  sockaddr = 0x0,
  count = 0,
  flags = 0 '\000',
  read_after_accept = 0 '\000',
  ipv6only = 0 '\000',
  socklen = 0 '\000',
  address_length = 0 '\000'
```

But I guess these are set in some other process after being forked from the main process...

alejandro-colomar: I agree that we should not even try, even if we can.  It's not possible to know exactly where to stop, and I don't feel happy trying to *rmdir(2)* everything up to `/`.

Also, GNU standards for Makefiles say that a Makefile doesn't need to try to delete directories, because they could have existed previously:

>  There is no need to delete parent directories that were created with ‘mkdir -p’, since they could have existed anyway.
<https://www.gnu.org/prep/standards/html_node/Standard-Targets.html>

That is said for targets that remove stuff, such as `clean`.  All the time, it talks about removing files, which is what makes sense removing.

Also, as we've seen, other programs don't remove dirs either.

ac000:
- Rebased with master
- Use nxt_int_t in nxt_fs_mkdir_dirname()

src/nxt_fs.c
```diff
@@ -273,6 +273,32 @@ nxt_fs_mkdir_all(const u_char *dir, mode_t mode)
 }
 
 
+nxt_int_t
+nxt_fs_mkdir_dirname(const u_char *path, mode_t mode)
+{
+    char       *ptr;
+    char       *dir;
```
alejandro-colomar, line 280: You could just compress the declaration of ptr and dir into a single line.  That's common here.

ac000: Ah, so it is...

ac000:
Added a paragraph to the commit message about the new nxt_fs_mkdir_dirname() function.

ac000:
Changes:

- Call nxt_fs_mkdir() vs nxt_fs_mkdir_all() in nxt_fs_mkdir_dirname()
- Rename nxt_fs_mkdir_dirname() to nxt_fs_mkdir_parent()
- Create directories specifically with mode 0755

alejandro-colomar:
Hi Andrew!

Thanks for the change.  For next time, please first rebase and push, and then add your changes and push again (if it's easy for you, of course).  Otherwise, the interdiff gets all of the changes to master, which is slightly unreadable.  :)

Cheers,
Alex

ac000:
Ah, OK. I will try and remember that!

alejandro-colomar:
> Changes:
> 
>     * Call nxt_fs_mkdir() vs nxt_fs_mkdir_all() in nxt_fs_mkdir_dirname()
> 
>     * Rename nxt_fs_mkdir_dirname() to nxt_fs_mkdir_parent()
> 
>     * Create directories specifically with mode 0755
> 
> 
> @alejandro-colomar removed your Tested-by & Reviewed-by tags due to changes...


All of the changes look good to me.  Tested and reviewed again. :-)

ac000:
No functional changes.

Rebased with master and updated email addresses to @nginx.com

PR #754, opened 2022-09-11
change entrypoint order

aapjeisbaas:
I would like to be able to run some shell scripts before loading my config.json
Otherwise my startup fails because the script downloads the index.php that is referenced in the config.json


tippexs:
Good Morning, thanks for your PR. The ordering of the cert, conf, shell script are does make sense. In your case I would like you to you the shell script for doing the bootstrapping of your application and later apply the unit configuration by sending it the controller api using `curl`.

This is the way I did that with NodeJS and that should just work for you.

PR #775, opened 2022-10-24
feat: improve docker-entrypoint.sh

jd1378:
Hi
I was using this file in one of my custom docker images, and I think we could improve it a bit by switching to sh, which is more widely available than bash, and to include soft links in the find command results, so we can create softlinks in docker entry directory as well

### DIFF
--- a/pkg/docker/docker-entrypoint.sh
+++ b/pkg/docker/docker-entrypoint.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/bin/sh
 
 set -e
 
--- a/pkg/docker/docker-entrypoint.sh
+++ b/pkg/docker/docker-entrypoint.sh
@@ -5,8 +5,8 @@ set -e
 curl_put()
 {
     RET=$(/usr/bin/curl -s -w '%{http_code}' -X PUT --data-binary @$1 --unix-socket /var/run/control.unit.sock http://localhost/$2)
-    RET_BODY=$(printf '%s' "$RET" | cut -c 1)
-    RET_STATUS=$(echo $RET | /usr/bin/tail -c 4)
+    RET_BODY=$(echo "$RET" | sed 's/.\{3\}$//')
+    RET_STATUS=$(echo "$RET" | tail -1 | grep -ohe '...$')
     if [ "$RET_STATUS" -ne "200" ]; then
         echo "$0: Error: HTTP response status code is '$RET_STATUS'"
         echo "$RET_BODY"
--- a/pkg/docker/docker-entrypoint.sh
+++ b/pkg/docker/docker-entrypoint.sh
@@ -1,11 +1,11 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 
 set -e
 
 curl_put()
 {
-    RET=`/usr/bin/curl -s -w '%{http_code}' -X PUT --data-binary @$1 --unix-socket /var/run/control.unit.sock http://localhost/$2`
-    RET_BODY=${RET::-3}
+    RET=$(/usr/bin/curl -s -w '%{http_code}' -X PUT --data-binary @$1 --unix-socket /var/run/control.unit.sock http://localhost/$2)
+    RET_BODY=$(printf '%s' "$RET" | cut -c 1)
     RET_STATUS=$(echo $RET | /usr/bin/tail -c 4)
     if [ "$RET_STATUS" -ne "200" ]; then
         echo "$0: Error: HTTP response status code is '$RET_STATUS'"
@@ -18,7 +18,7 @@ curl_put()
     return 0
 }
 
-if [ "$1" = "unitd" -o "$1" = "unitd-debug" ]; then
+if [ "$1" = "unitd" ] || [ "$1" = "unitd-debug" ]; then
     if /usr/bin/find "/var/lib/unit/" -mindepth 1 -print -quit 2>/dev/null | /bin/grep -q .; then
         echo "$0: /var/lib/unit/ is not empty, skipping initial configuration..."
     else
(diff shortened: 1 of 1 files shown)


pkg/docker/docker-entrypoint.sh
```diff
@@ -1,11 +1,11 @@
-#!/usr/bin/env bash
```
alejandro-colomar, line 1: The shebang for sh(1) should be  `#!/bin/sh`.

pkg/docker/docker-entrypoint.sh
```diff
@@ -1,11 +1,11 @@
...
+#!/usr/bin/env sh
 
 set -e
 
 curl_put()
 {
-    RET=`/usr/bin/curl -s -w '%{http_code}' -X PUT --data-binary @$1 --unix-socket /var/run/control.unit.sock http://localhost/$2`
-    RET_BODY=${RET::-3}
+    RET=$(/usr/bin/curl -s -w '%{http_code}' -X PUT --data-binary @$1 --unix-socket /var/run/control.unit.sock http://localhost/$2)
+    RET_BODY=$(printf '%s' "$RET" | cut -c 1)
```
alejandro-colomar, line 8: Can't we just have an `echo(1)` here?

jd1378: Sorry but I don't know how it exactly  works to change this

pkg/docker/docker-entrypoint.sh
jd1378: done

pkg/docker/docker-entrypoint.sh
alejandro-colomar, line 8: I don't fully understand bash's `${RET::-3}`. I agree with you it's quite cryptic, and prefer your approach with pipes and standard commands.

But could you please explain why those two are equivalent?  The bash code is removing the last 3 characters of a string, and your code is keeping only the first character.

Please show which kind of string is expected/valid at this point, and why both are equivalent.

alejandro-colomar: I meant `$(echo "$RET" | cut -c 1)`.  Isn't that equivalent to what you wrote?  Why do we need printf(1)?

pkg/docker/docker-entrypoint.sh
jd1378: but wouldnt it make it less portable?

alejandro-colomar: I think not.  In case of `bash(1)`, yes, because some systems install `bash(1)` in a weird place, so `env(1)` always knows where it is.

However, `sh(1)` is always (AFAIK) available at `/bin/sh`.

I just looked at POSIX, and it seems that it allows `sh(1)` to be in a different location.  But that would break lots of scripts, so I don't think any reasonable system puts `sh(1)` in a different place.  BTW, I guess there's a reason most sources refer to `sh(1)` as `/bin/sh` instead of `sh(1)`.

pkg/docker/docker-entrypoint.sh
alejandro-colomar: If you don't know how echo works: it just prints its arguments, separated by space (but normally you just pass a signle argument, so it prints it literally).

jd1378: so I figured what its doing in original file
what I did was clearly nonsense

so what it does is simply get the string without the last 3 characters, because the last 3 chars are the http response code

jd1378: okay I fixed it
I hope its clear now

pkg/docker/docker-entrypoint.sh
jd1378: thanks for the info

thresheek:
Why do you need symlinks in the docker-entrypoint.d directory? Where are they pointing at - how exactly does that work?

pkg/docker/docker-entrypoint.sh
```diff
@@ -1,12 +1,12 @@
...
 
 set -e
 
 curl_put()
 {
     RET=$(/usr/bin/curl -s -w '%{http_code}' -X PUT --data-binary @$1 --unix-socket /var/run/control.unit.sock http://localhost/$2)
-    RET_BODY=$(printf '%s' "$RET" | cut -c 1)
-    RET_STATUS=$(echo $RET | /usr/bin/tail -c 4)
+    RET_BODY=$(echo "$RET" | sed 's/.\{3\}$//')
+    RET_STATUS=$(echo "$RET" | tail -1 | grep -ohe '...$')
```
alejandro-colomar, line 9: I think `tail -1` is an extension (I didn't even know about it), and is not documented in coreutils.

POSIX only guarantees `-n 1`.

pkg/docker/docker-entrypoint.sh
```diff
@@ -1,12 +1,12 @@
...
 set -e
 
 curl_put()
 {
-    RET=`/usr/bin/curl -s -w '%{http_code}' -X PUT --data-binary @$1 --unix-socket /var/run/control.unit.sock http://localhost/$2`
-    RET_BODY=${RET::-3}
-    RET_STATUS=$(echo $RET | /usr/bin/tail -c 4)
+    RET=$(/usr/bin/curl -s -w '%{http_code}' -X PUT --data-binary @$1 --unix-socket /var/run/control.unit.sock http://localhost/$2)
+    RET_BODY=$(echo "$RET" | sed 's/.\{3\}$//')
+    RET_STATUS=$(echo "$RET" | tail -1 | grep -ohe '...$')
```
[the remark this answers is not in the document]
jd1378: thanks for the info

I didn't need to change RET_STATUS in the first place I think, so I reverted it
I only changed RET_BODY now

pkg/docker/docker-entrypoint.sh
jd1378: done

pkg/docker/docker-entrypoint.sh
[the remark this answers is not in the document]
alejandro-colomar: It's looking good!

alejandro-colomar (approved):

jd1378:
> Why do you need symlinks in the docker-entrypoint.d directory? Where are they pointing at - how exactly does that work?

because that way I can keep my script inside my `/var/www` dir, create a symlink, and it would run it as if it was executed in `/var/www` dir, from what I understand, so I wouldn't need to cd inside the script.
correct me if I'm having wrong assumptions here

thresheek:
That does not make any sense to me on multiple levels.

With scripts/cert+keys under /var/www, you're basically exposing your configuration through a directory typically used as a publically accessible webroot, mixed with your static/application files.

But ok, even if any other directory is used to store those files, I don't get why you'd want an extra step of creating symlinks instead of just specifying that directory with specific configuration files in a volume mount?

I don't mind making the script POSIX-compliant (although I don't really see much point in doing so, since we don't ship any non-bash images), but the symlink part does not seem to provide any benefit.

alejandro-colomar:
> That does not make any sense to me on multiple levels.
> 
> With scripts/cert+keys under /var/www, you're basically exposing your configuration through a directory typically used as a publically accessible webroot, mixed with your static/application files.

Yeah, `/var/www` is a bad directory for that.

But I'd think it would be fine putting a script in `/usr/local/bin` or a similar place, and then symlinking to wherever the script needs to be put.

> But ok, even if any other directory is used to store those files, I don't get why you'd want an extra step of creating symlinks instead of just specifying that directory with specific configuration files in a volume mount?

The files could be embedded in the image at build time.  I did that a long time ago with some images of mine.  Not sure if it was ideal, but it worked for me.

> I don't mind making the script POSIX-compliant (although I don't really see much point in doing so, since we don't ship any non-bash images), but the symlink part does not seem to provide any benefit.

Agree, it makes sense to split the patch in two.  I can do that while committing.

I don't think it's important being sh(1)-compliant, but the changes in this case are using `$()` instead of ``, and also using Unix tools instead of cryptic shell features, which are good changes, even for a bash(1) script.  In fact, maybe we want to go back to a bash(1) shebang some day to use things like `-o pipefail`, which I'm not sure if we want in this script, but the contents of the script being as sh(1)-compatible as possible are a good thing for readability anyway.

If you agree, I'll merge the POSIX changes already, since they look good to me.

thresheek:
> But I'd think it would be fine putting a script in /usr/local/bin or a similar place, and then symlinking to wherever the script needs to be put.

That only complicates things, since /docker-entrypoint.d/ still stays and violates FHS. I don't like it as well but that's a well-known and common directory in docker world.

> The files could be embedded in the image at build time. I did that a long time ago with some images of mine. Not sure if it was ideal, but it worked for me.

Sure, there are many ways to do things, but I'd rather promote clear separation between configuration and application files.

> If you agree, I'll merge the POSIX changes already, since they look good to me.

Sure.  Maybe also use a full path to sed binary while at it to conform with the other parts of the script. Thanks!

alejandro-colomar:
> > But I'd think it would be fine putting a script in /usr/local/bin or a similar place, and then symlinking to wherever the script needs to be put.
> 
> That only complicates things, since /docker-entrypoint.d/ still stays and violates FHS. I don't like it as well but that's a well-known and common directory in docker world.
> 
> > The files could be embedded in the image at build time. I did that a long time ago with some images of mine. Not sure if it was ideal, but it worked for me.
> 
> Sure, there are many ways to do things, but I'd rather promote clear separation between configuration and application files.

Yeah, since @jd1378 seems to not need symlinks (since they won't offer the features he thought they had), I think we'll reject the changes for now (if someone in the future comes with a better reason, we can reconsider).

> > If you agree, I'll merge the POSIX changes already, since they look good to me.
> 
> Sure. Maybe also use a full path to sed binary while at it to conform with the other parts of the script. Thanks!

Okay, I'll merge those changes, and use the full paths (BTW, I'm curious: why are we using full paths?  Are they reliable?).

thresheek:
Full paths are used so an incorrect PATH set through environment when container is launched won't mess with the entrypoint script.

alejandro-colomar:
The POSIX compatibility changes have been merged.

jd1378:
thanks a lot for the explanation

about being  POSIX-compliant: I have made an alpine image with unit, and to run this script, I had to make it compliant

Issue #795, opened 2022-11-30
Critical Undefined Behavior: write past the end of an array


alejandro-colomar:
While trying to fix a bug, I found these:

```c
$ grepc -tfd nxt_clone_credential_setgroups
#./src/nxt_clone.c:37:
nxt_int_t
nxt_clone_credential_setgroups(nxt_task_t *task, pid_t child_pid,
    const char *str)
{
    int     fd, n;
    u_char  *p, *end;
    u_char  path[PATH_MAX];

    end = path + PATH_MAX;
    p = nxt_sprintf(path, end, "/proc/%d/setgroups", child_pid);
    *p = '\0';

    if (nxt_slow_path(p == end)) {
        nxt_alert(task, "error write past the buffer: %s", path);
        return NXT_ERROR;
    }

    ...
}
```

```c
$ grepc -tfd nxt_discovery_modules
./src/nxt_application.c:205:
static nxt_buf_t *
nxt_discovery_modules(nxt_task_t *task, const char *path)
{
    ...

    *p++ = ']';

    if (nxt_slow_path(p > end)) {
        nxt_alert(task, "discovery write past the buffer");
        goto fail;
    }

    ...
}
```

```c
$ grepc -tfd nxt_clone_credential_map_set
./src/nxt_clone.c:133:
nxt_int_t
nxt_clone_credential_map_set(nxt_task_t *task, const char* mapfile, pid_t pid,
    nxt_int_t default_container, nxt_int_t default_host,
    nxt_clone_credential_map_t *map)
{
    ...

    if (map->size > 0) {
        ...

    } else {
        ...

        end = mapinfo + len;
        p = nxt_sprintf(mapinfo, end, "%d %d 1",
                        default_container, default_host);
        *p = '\0';

        if (nxt_slow_path(p == end)) {
            nxt_alert(task, "write past mapinfo buffer");
            nxt_free(mapinfo);
            return NXT_ERROR;
        }
    }

    ...
}
```

Writing past the end is invoking Critical Undefined Behavior, and checking after it has happened is too late.  We should move the checks to *before* the writing.

alejandro-colomar:
However, that doesn't fix the other important problem with this code:  we're not really detecting string truncation at all.

`nxt_sprintf()` needs to be fixed to be able to report truncation.  A good example of how to do that, IMO, would be `stpecpy()`, which returns `end + 1` on truncation: <https://software.codidact.com/posts/285946>

CC:
@andrey-zelenkov 
[...]

PR #805, opened 2022-12-10
Add nxt_stpecpy(), and use it in isolation code to replace (and optimize) unsafe code

alejandro-colomar:
```
/*
 * SYNOPSIS
 *      char *stpecpy(char *dst, char past_end[0], const char *restrict src);
 *
 * ARGUMENTS
 *      dst     Destination string.  Will be null-terminated after the
 *              call.
 *
 *      past_end
 *              Pointer to one past the last byte in the 'dst' buffer.
 *              This is used to truncate the output string if
 *              necessary.
 *
 *      src     Source string to copy from.  This string should be
 *              terminated with a NUL byte.
 *
 * DESCRIPTION
 *      This function is similar to stpcpy(3), but it truncates the
 *      output string as necessary.
 *
 *      You can chain calls to this function as you would do with
 *      stpcpy(3).  Only after the last call it is necessary to check
 *      for truncation (but don't forget to do so):
 *
 *          past_end = buf + sizeof(buf);
 *          p = buf;
 *          p = stpecpy(p, past_end, "foo");
 *          p = stpecpy(p, past_end, "bar");
 *          p = stpecpy(p, past_end, "baz");
 *          if (p == past_end) {
 *              --p;
 *              // Handle truncation.
 *          }
 *
 *      This function is an improvement over strscpy(9) and strlcpy(3)
 *      (and strlcat(3)), both in terms of usability and performance.
 *
 * RETURN VALUE
 *      This function returns a pointer to the 'end' of the destination
 *      string, that is, a pointer to its NUL byte.
 *
 *      However, if the destination string has been truncated, it
 *      returns a pointer to one past the end (past_end).
 *
 * ERRORS
 *      This function has no errors.  If strlcpy(3)-like behavior is
 *      desired, that is, if you want to crash if the input string is
 *      invalid (doesn't contain a terminating NUL byte), you may use
 *      stpecpyx() instead:
 *
 *      inline char *
 *      stpecpyx(char *dst, const char *restrict src, char past_end[0])
 *      {
 *          if (unlikely(src[strlen(src)] != '\0'))
 *              raise(SIGSEGV);
 *
 *          return stpecpy(dst, src, past_end);
 *      }
 *
 * CAVEATS
 *      Truncation is normally a bug, and you should detect it.  Only
 *      if a truncated string is not a problem, it makes sense to ignore
 *      the return value of this function.
 */

This function is in the public domain, and its design can be found in
the Codidact forum, in the following link.

Link: <https://software.codidact.com/posts/285946/287522#answer-287522>
```

### DIFF
--- a/src/nxt_cgroup.c
+++ b/src/nxt_cgroup.c
@@ -8,9 +8,9 @@
 #include <nxt_cgroup.h>
 
 
-static int nxt_mk_cgpath_relative(nxt_task_t *task, const char *dir,
+static char *nxt_mk_cgpath_relative(nxt_task_t *task, const char *dir,
     char *cgpath);
-static nxt_int_t nxt_mk_cgpath(nxt_task_t *task, const char *dir,
+static char *nxt_mk_cgpath(nxt_task_t *task, const char *dir,
     char *cgpath);
 
 
@@ -18,6 +18,7 @@ nxt_int_t
 nxt_cgroup_proc_add(nxt_task_t *task, nxt_process_t *process)
 {
     int        len;
+    char       *p, *past_end;
     char       cgprocs[NXT_MAX_PATH_LEN];
     FILE       *fp;
     nxt_int_t  ret;
@@ -29,8 +30,8 @@ nxt_cgroup_proc_add(nxt_task_t *task, nxt_process_t *process)
         return NXT_OK;
     }
 
-    ret = nxt_mk_cgpath(task, process->isolation.cgroup.path, cgprocs);
-    if (nxt_slow_path(ret == NXT_ERROR)) {
+    p = nxt_mk_cgpath(task, process->isolation.cgroup.path, cgprocs);
+    if (nxt_slow_path(p == NULL)) {
         return NXT_ERROR;
     }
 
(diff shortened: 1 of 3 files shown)


alejandro-colomar:
BTW, before you ask, this is a perfect case for why `restrict` is necessary.  

```
$ grepc -tfd nxt_stpecpy
./src/nxt_string.c:154:
char *
nxt_stpecpy(char *dst, const char *restrict src, char *past_end)
{
    char *p;

    if (dst == past_end)
        return past_end;

    p = memccpy(dst, src, '\0', past_end - dst);
    if (p != NULL)
        return p - 1;

    /* truncation detected */
    past_end[-1] = '\0';
    return past_end;
}
```

It tells the programmer that the first and last pointers to the function are pointers to the same object, while the second one points to a distinct object.  It helps readability (at the same time that it allows the compiler to better understand the code, and also be able to warn about silly mistakes).

alejandro-colomar:
I think I will change the function prototype, to pass the `past_end` pointer next to its `dst` relative.  It changes the "standard" way of putting the limits after all other parameters, but it would improve alignment and therefore readability, and also, there's no standard function accepting `past_end`, so I have a bit of freedom there, in therms of consistency.

alejandro-colomar:
Cc: @igorsysoev  @VBart  @mar0x

alejandro-colomar:
I'm closing temporarily until the next release.

ac000:
I do object to calling the current code _unsafe_.

Can you point out how in its current form it's _unsafe_? 

Or do you really mean, it's hard to get right?

What about the other dozen or so uses of snprintf(3)?

alejandro-colomar:
> I do object to calling the current code _unsafe_.

There are different levels of unsafety, and they are different to different people.

> 
> Can you point out how in its current form it's _unsafe_?

There are no bugs or vulnerabilities in your code, if that's the question.

> 
> Or do you really mean, it's hard to get right?

Yes, I meant that, which to me is the same as unsafe.  I'm sorry if this usage of the term caused any confusion/troubles, and I reworded to clarify.

> 
> What about the other dozen or so uses of snprintf(3)?

Oh, that one is really broken.  I only had a fast look and found several bugs (UB).  Reported here: <https://github.com/nginx/unit/issues/795#issuecomment-1345400420>.

PR #806, opened 2022-12-11
Update readme.md, explaining how to obtain the path/to/control.unit.sock and also fixed wrong ports on samples

0x6f677548:


ac000 (changes requested):
Hi,

Please undo this change in the previous commit and then just drop this one.

ac000 (changes requested):
Seeing as '8080' is used in a couple of places. I'd just change the '8000' in the PUT command

```shell
# curl -X PUT -d '{"127.0.0.1:8000": {"pass": "applications/helloworld"}}'  \
       --unix-socket /path/to/control.unit.sock http://localhost/config/listeners
```


to '8080'.

0x6f677548:
> Seeing as '8080' is used in a couple of places. I'd just change the '8000' in the PUT command
> 
> ```shell
> # curl -X PUT -d '{"127.0.0.1:8000": {"pass": "applications/helloworld"}}'  \
>        --unix-socket /path/to/control.unit.sock http://localhost/config/listeners
> ```
> 
> to '8080'.

either 8000 or 8080 should be ok if used consistently through the samples. My point was that both ports were being used on the same guide (readme.md) which could be confusing for some newcomers.

ac000:
> either 8000 or 8080 should be ok if used consistently through the samples. My point was that both ports were being used on the same guide (readme.md) which could be confusing for some newcomers.

Yes, exactly!

So rather than change the 8080's to 8000, change the single 8000 to 8080...

README.md
```diff
@@ -123,7 +133,7 @@ Now Unit accepts requests at the specified IP and port, passing them to the
 application process. Your app works!
 
 ``` console
-$ curl 127.0.0.1:8080
+$ curl 127.0.0.1:8000
 
       Hello, PHP on Unit!
 ```
```
ac000, line 139: Drop this hunk.

ac000 (changes requested):

ac000 (changes requested):
Change the last commit to change the port to 8080 in the curl PUT command.

0x6f677548:
I accidentally deleted patch branch while rebasing. Re-submitted new pull request #816

PR #811, opened 2022-12-16
Tools: using double quotes for the apostrophe.

lcrilly:


lcrilly:
Pushed directly to Mercurial

PR #816, opened 2022-12-20
Update readme.md, explaining how to obtain the path/to/control.unit.sock and also fixed wrong ports on samples

0x6f677548:
This was pull request 806, but I wrongly deleted the patch branch while rebasing :)

- Added text explaining how to get the path to control sock
- Fixed curl example (wrong port)


ac000:
https://github.com/nginx/unit/pull/816/commits/f86600c011cdd7f9abfac6e1ce5af6497b623883 looks fine, though we'd probably need to change the summary to something like 'Used port 8080 consistently in README.md.' (yes, past tense, slightly awkward I know...)

The other commit I'm not 100% sure about. For me (self compiled) I get

```
$ /opt/unit/sbin/unitd --version
unit version: 1.29.0
configured as ./configure --prefix=/opt/unit
```

However with our Fedora package I do get (and the README does look like it's geared towards using a packaged Unit and not self compiled from source).

```
$ /sbin/unitd --version
unit version: 1.29.0
configured as ./configure --prefix=/usr --state=/var/lib/unit --control=unix:/var/run/unit/control.sock --pid=/var/run/unit/unit.pid --log=/var/log/unit/unit.log --tmp=/var/tmp --user=unit --group=unit --tests --openssl --njs --modules=/usr/lib64/unit/modules --libdir=/usr/lib64 --cc-opt='-O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -fPIC' --ld-opt='-Wl,-z,relro -Wl,-z,now -pie'
```

I'm also unsure if the examples in the README are assuming if you have run 'setup-unit welcome' which for me gives (once I've installed the unit package and started it...)

```
$ sudo ./setup-unit welcome
{
}

You may want to try the following commands now:

Check out current unitd configuration:
    sudo curl --unix-socket /var/run/unit/control.sock http://localhost/config

Browse the welcome page:
    curl http://localhost:56991/
```

Although the setup-unit script starts unit on a random port which obviously doesn't match the examples.

So yes, your suggestion may well be a valid suitable portable way of getting the location of the control socket.

0x6f677548:
Changed the summary of 6f63cc8 

I don't know any other alternative to get the pre-defined control socket other than "unitd --version". I guess the other alternative would be to instruct the user to change the control socket to something known, using "unitd --control socket", but that might add additional complexity to new users testing the platform for the 1st time.

lcrilly:
`unitd --help` is a more reliable way of detecting the default control socket. It's also a useful thing to see when you're getting started.

The long term solution here is to introduce a command line tool that will take care of locating the control socket. We don't currently ship [unitc](https://github.com/nginx/unit/blob/master/tools/unitc) with the packages but this would simplify the quickstart guide.

Further to @ac000 's comment above, we might also consider replacing the current quickstart guide with `setup-unit welcome` which also takes care of much of this and shows the correct curl(1) command.

ac000:
> I don't know any other alternative to get the pre-defined control socket other than "unitd --version". I guess the other 

On Linux with the iproute tools, you can simply do

```shell
$ sudo ss -xlp | grep unitd
u_str LISTEN 0      511                /var/run/unit/control.sock.tmp 216125            * 0    users:(("unitd",pid=44738,fd=6),("unitd",pid=44737,fd=6),("unitd",pid=44731,fd=6))
```

0x6f677548:
> `unitd --help` is a more reliable way of detecting the default control socket. It's also a useful thing to see when you're getting started.
> 
It depends. I guess that would assume the config wasn't changed from default values, which might not be the future case in a container, for example. Right now, the provided containers are using the default values, but if that changes in the future, ```unitd --help``` would induce the user in error.

0x6f677548:
> > I don't know any other alternative to get the pre-defined control socket other than "unitd --version". I guess the other
> 
> On Linux with the iproute tools, you can simply do
> 
> ```shell
> $ sudo ss -xlp | grep unitd
> u_str LISTEN 0      511                /var/run/unit/control.sock.tmp 216125            * 0    users:(("unitd",pid=44738,fd=6),("unitd",pid=44737,fd=6),("unitd",pid=44731,fd=6))
> ```

yes, but that would imply the user is having iproute tools installed.

lcrilly:
> It depends. I guess that would assume the config wasn't changed from default values, which might not be the future case in a container, for example. Right now, the provided containers are using the default values, but if that changes in the future, unitd --help would induce the user in error.

All true, but not relevant here.

If the user was able to change the defaults then they are capable of reading the documentation. In which case they are not following the quickstart guide.

The quickstart guide should optimize for the first-time user, where it is safe to assume that all defaults are in use.

I'll repeat my assertion that `unitd --help` is a more convenient way of obtaining the default values than `unitd --version` is.

callahad:
Thank you so much, and I'm sorry this has lingered for so long. I'm going to close this PR for now, and promise that a complete overhaul of our docs is coming this year.

Issue #815, opened 2022-12-20
Latest Docker Image Error Python failed to call 'asyncio.get_event_loop'


treyrich:
Upgrading our docker image from 1.28.0-python3.10 to 1.29.0-python3.11 results in the following log output:
```
2022/12/20 18:38:19 [info] 21#21 unit 1.29.0 started
2022/12/20 18:38:19 [info] 28#28 discovery started
2022/12/20 18:38:19 [notice] 28#28 module: python 3.11.1 "/usr/lib/unit/modules/python3.unit.so"
2022/12/20 18:38:19 [info] 24#24 controller started
2022/12/20 18:38:19 [notice] 24#24 process 28 exited with code 0
2022/12/20 18:38:19 [info] 33#33 router started
2022/12/20 18:38:19 [info] 33#33 OpenSSL 1.1.1n  15 Mar 2022, 101010ef
{
	"certificates": {},
	"config": {
		"listeners": {},
		"applications": {}
	},

	"status": {
		"connections": {
			"accepted": 0,
			"active": 0,
			"idle": 0,
			"closed": 0
		},

		"requests": {
			"total": 0
		},

		"applications": {}
	}
}
/usr/local/bin/docker-entrypoint.sh: Looking for certificate bundles in /docker-entrypoint.d/...
/usr/local/bin/docker-entrypoint.sh: Looking for configuration snippets in /docker-entrypoint.d/...
/usr/local/bin/docker-entrypoint.sh: Applying configuration /docker-entrypoint.d/config.json
2022/12/20 18:38:19 [info] 44#44 "webapp" prototype started
2022/12/20 18:38:19 [info] 46#46 "webapp" application started
2022/12/20 18:38:19 [alert] 0#46 [unit] Python failed to call 'asyncio.get_event_loop'
2022/12/20 18:38:20 [notice] 44#44 app process 46 exited with code 1
2022/12/20 18:38:20 [warn] 33#33 failed to start application "webapp"
2022/12/20 18:38:20 [alert] 33#33 failed to apply new conf
2022/12/20 18:38:20 [notice] 24#24 process 44 exited with code 0
/usr/local/bin/docker-entrypoint.sh: Error: HTTP response status code is '500'
 }error": "Failed to apply new configuration."
```

I've confirmed that this is a bug by replicating it with the simplest python application (copied from Unit docs):
```python

async def application(scope, receive, send):

    await send({"type": "http.response.start", "status": 200})

[...]

Issue #560, opened 2021-06-20
Python: asgi app stucks on asyncio loop


ingria:
I'm trying to use unit for my application, which is currently running on uvicorn (asgi3). 

```python
service = MyBackgroundService()

async def application(scope, receive, send):
    if scope['type'] == 'lifespan':
        message = await receive()
        if message['type'] == 'lifespan.startup':
            await send({"type": "lifespan.startup.complete"})
            await service.bootUp()
            logger.info('This line is never displayed')
        return

    # handle http request
```

Under the hood of `MyBackgroundService.bootUp()` is an [infinite loop](https://github.com/LonamiWebs/Telethon/blob/4b16183d2bbe80cbf4dabdb266a8015c5bf975cc/telethon/network/mtprotosender.py#L500-L507), which is constanly running and receiving new data. The idea is that this service is running asynchronously "in background" while asgi `application` is interacting with it.

The problem is that the app stucks on that `recv()` loop and becomes unresponsive (doesn't process any requests and cannot be stopped with anything except `kill -9`).

I'm not sure whether the problem is in my code or in unit, but it works just fine with uvicorn. My guess is that there's something wrong with event loops (maybe app and unit are using different loops?).

App config:

```json
"01_01": {
    "type": "python 3",
    "module": "worker",
    "callable": "application",
    "protocol": "asgi",
    "working_directory": "/home/user/dir",
    "path": "/home/user/dir",
    "user": "user",
    "environment": {
        "SOME_VARS": "some_value"
    }
}
```

mar0x:
Hello,

`telethon.network.MTProtoSender.connect()->_connect()` [creates tasks](https://github.com/LonamiWebs/Telethon/blob/4b16183d2bbe80cbf4dabdb266a8015c5bf975cc/telethon/network/mtprotosender.py#L259-L264) to run send and receive loops. So, it is not 100% correct saying _it is an infinite loop_ unless you have your own loop inside `bootUp()` method. According to [ASGI Lifespan specification](https://asgi.readthedocs.io/en/latest/specs/lifespan.html) your application needs to await further `shutdown` message in `lifespan` scope.

The fact your application running on uvicorn, but blocks on Unit points on possible problem in Unit ASGI implementation and requires deeper investigation. Please share as much information about your application as possible. Sources or tiny mockup which replicates the issue would be perfect.

Thanks for your attention to Unit. I’m looking forward to your reply.

ingria:
I've created a mockup project:

[...]

PR #826, opened 2023-01-16
Fix/python asyncio

88Ocelot:
Fixes for python asgi

### DIFF
--- a/src/python/nxt_python_asgi.c
+++ b/src/python/nxt_python_asgi.c
@@ -17,7 +17,7 @@
 
 
 static PyObject *nxt_python_asgi_get_func(PyObject *obj);
-static int nxt_python_asgi_ctx_data_alloc(void **pdata, int main);
+static int nxt_python_asgi_ctx_data_alloc(void **pdata);
 static void nxt_python_asgi_ctx_data_free(void *data);
 static int nxt_python_asgi_startup(void *data);
 static int nxt_python_asgi_run(nxt_unit_ctx_t *ctx);
@@ -199,9 +199,9 @@ nxt_python_asgi_init(nxt_unit_init_t *init, nxt_python_proto_t *proto)
 
     return NXT_UNIT_OK;
 }
-#if NXT_HAVE_ASYNCIO_RUNNER
+#if (NXT_HAVE_ASYNCIO_RUNNER)
 static int
-nxt_python_asgi_ctx_data_alloc(void **pdata, int main)
+nxt_python_asgi_ctx_data_alloc(void **pdata)
 {
     uint32_t                i;
     PyObject                *asyncio, *loop, *obj, *runner_ref, *runner, *loop_ref;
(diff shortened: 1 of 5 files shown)


src/python/nxt_python_asgi.c
```diff
@@ -199,7 +199,153 @@ nxt_python_asgi_init(nxt_unit_init_t *init, nxt_python_proto_t *proto)
 
     return NXT_UNIT_OK;
 }
+#if NXT_HAVE_ASYNCIO_RUNNER
```
alejandro-colomar, line 202: Please wrap this in parentheses:

`#if (NXT_HAVE_...)`

src/python/nxt_python_asgi.c
```diff
@@ -199,7 +199,153 @@ nxt_python_asgi_init(nxt_unit_init_t *init, nxt_python_proto_t *proto)
...
+    Py_XDECREF(ctx_data->loop_remove_reader);
+    Py_XDECREF(ctx_data->quit_future);
+    Py_XDECREF(ctx_data->quit_future_set_result);
+    close_loop_func = PyObject_GetAttrString(ctx_data->runner, "close");
+    PyObject_CallObject(close_loop_func, NULL);
+    Py_XDECREF(ctx_data->runner);
+
+    nxt_unit_free(NULL, ctx_data);
+}
+#else
```
alejandro-colomar, line 348: This code is almost identical to the other part of the `#else`.  I think it makes more sense to just wrap a few lines in preprocessor conditionals, rather than the entire functions.

src/python/nxt_python_asgi.h
```diff
@@ -22,7 +22,21 @@ typedef struct {
...
+    PyObject         *loop_create_task;
+    PyObject         *loop_call_soon;
+    PyObject         *loop_add_reader;
+    PyObject         *loop_remove_reader;
+    PyObject         *quit_future;
+    PyObject         *quit_future_set_result;
+    PyObject         **target_lifespans;
+    PyObject         *runner;
+} nxt_py_asgi_ctx_data_t;
+#else
```
alejandro-colomar, line 39: Please just add the member to the structure, wrapped in preprocessor, rather than copying the entire structure.

src/python/nxt_python_asgi.h
```diff
@@ -22,7 +22,21 @@ typedef struct {
     nxt_unit_request_info_t  *req;
     uint64_t                 content_length;
 } nxt_py_asgi_add_field_ctx_t;
-
+#ifdef NXT_HAVE_ASYNCIO_RUNNER
```
alejandro-colomar, line 25: `#id defined()` is preferable over `#ifdef`.  However, since we can use just `#if ()`, it's better to use that (as in the other place).

alejandro-colomar:
Apart from the code comments, could you please detail in the commit messages what this does, and especially why we want it?  I don't have context about this change.  However, the code changes look something reasonable; thanks for the work!

88Ocelot:
its the fix for python 3.10-3.11.
#815 

alongside latest python patch releases were made breaking change for lower level asyncio api.
https://github.com/python/cpython/issues/93453#issuecomment-1361226496

Also seems like using asyncio api `get_event_loop` for main thread was related to this issue
#560

src/python/nxt_python_asgi.c
```diff
@@ -199,10 +199,133 @@ nxt_python_asgi_init(nxt_unit_init_t *init, nxt_python_proto_t *proto)
...
+
+    nxt_python_asgi_ctx_data_free(ctx_data);
+
+    Py_XDECREF(loop);
+    Py_XDECREF(asyncio);
+
+    return NXT_UNIT_ERROR;
+}
+
+#else
```
alejandro-colomar, line 325: This function can also be compacted to a single one with small preprocessor in it.

ac000:
This has been fixed via 4b7a95469cf309b437fc1a8201a93f14c33ce3a8, 9c0a4a09978e0defbc6656c7866a7ca147236733, cbc01907fef6a028f02d9051b909def920c49f24

PR #850, opened 2023-03-07
Conditionally log the request line and the selected route

alejandro-colomar:
Ignore `Router: factored out listeners creation loop.` and `Router: factored out application creation loop.` for now.

### DIFF
--- a/src/nxt_conf_validation.c
+++ b/src/nxt_conf_validation.c
@@ -344,6 +344,9 @@ static nxt_conf_vldt_object_t  nxt_conf_vldt_http_members[] = {
         .type       = NXT_CONF_VLDT_OBJECT,
         .validator  = nxt_conf_vldt_object,
         .u.members  = nxt_conf_vldt_static_members,
+    }, {
+        .name       = nxt_string("log_route"),
+        .type       = NXT_CONF_VLDT_BOOLEAN,
     },
 
     NXT_CONF_VLDT_END
--- a/src/nxt_h1proto.c
+++ b/src/nxt_h1proto.c
@@ -507,6 +507,7 @@ nxt_h1p_conn_request_init(nxt_task_t *task, void *obj, void *data)
 
             r->conf = joint;
             skcf = joint->socket_conf;
+            r->log_route = skcf->log_route;
 
             if (c->local == NULL) {
                 c->local = skcf->sockaddr;
@@ -652,6 +653,11 @@ nxt_h1p_header_process(nxt_task_t *task, nxt_h1proto_t *h1p,
     r->method = &h1p->parser.method;
     r->path = &h1p->parser.path;
     r->args = &h1p->parser.args;
+
+    if (nxt_slow_path(r->log_route)) {
+        nxt_log(task, NXT_LOG_NOTICE, "http request line \"%V %V %V\"",
+                r->method, &r->target, &r->version);
+    }
 
     r->fields = h1p->parser.fields;
 
--- a/src/nxt_http.h
+++ b/src/nxt_http.h
@@ -189,6 +189,8 @@ struct nxt_http_request_s {
 
     nxt_http_status_t               status:16;
 
+    uint8_t                         log_route;    /* 1 bit */
+
     uint8_t                         pass_count;   /* 8 bits */
     uint8_t                         app_target;
     nxt_http_protocol_t             protocol:8;   /* 2 bits */
--- a/src/nxt_http_route.c
+++ b/src/nxt_http_route.c
@@ -1548,6 +1548,18 @@ nxt_http_route_handler(nxt_task_t *task, nxt_http_request_t *r,
 
     for (i = 0; i < route->items; i++) {
         action = nxt_http_route_match(task, r, route->match[i]);
+
+        if (nxt_slow_path(r->log_route)) {
+            const char  *sel = (action == NULL) ? "discarded" : "selected";
+
+            if (route->name.length == 0) {
+                nxt_log(task, NXT_LOG_INFO, "\"routes/%z\" %s", i, sel);
+            } else {
+                nxt_log(task, NXT_LOG_INFO, "\"routes/%V/%z\" %s",
+                        &route->name, i, sel);
+            }
+        }
+
         if (action != NULL) {
             return action;
         }
--- a/src/nxt_http_route.c
+++ b/src/nxt_http_route.c
@@ -1540,21 +1540,17 @@ static nxt_http_action_t *
 nxt_http_route_handler(nxt_task_t *task, nxt_http_request_t *r,
     nxt_http_action_t *start)
 {
+    size_t                  i;
     nxt_http_route_t        *route;
     nxt_http_action_t       *action;
-    nxt_http_route_match_t  **match, **end;
 
     route = start->u.route;
-    match = &route->match[0];
-    end = match + route->items;
 
-    while (match < end) {
-        action = nxt_http_route_match(task, r, *match);
+    for (i = 0; i < route->items; i++) {
+        action = nxt_http_route_match(task, r, route->match[i]);
         if (action != NULL) {
             return action;
         }
-
-        match++;
     }
 
     nxt_http_request_error(task, r, NXT_HTTP_NOT_FOUND);
--- a/src/nxt_http_static.c
+++ b/src/nxt_http_static.c
@@ -196,6 +196,9 @@ nxt_http_static(nxt_task_t *task, nxt_http_request_t *r,
 
         if (!nxt_str_eq(r->method, "HEAD", 4)) {
             if (action->fallback != NULL) {
+                if (nxt_slow_path(r->log_route)) {
+                    nxt_log(task, NXT_LOG_NOTICE, "\"fallback\" taken");
+                }
                 return action->fallback;
             }
 
@@ -690,6 +693,9 @@ nxt_http_static_next(nxt_task_t *task, nxt_http_request_t *r,
     }
 
     if (action->fallback != NULL) {
+        if (nxt_slow_path(r->log_route)) {
+            nxt_log(task, NXT_LOG_NOTICE, "\"fallback\" taken");
+        }
         nxt_http_request_action(task, r, action->fallback);
         return;
     }
--- a/src/nxt_router.c
+++ b/src/nxt_router.c
@@ -1513,6 +1513,12 @@ static nxt_conf_map_t  nxt_router_http_conf[] = {
         NXT_CONF_MAP_INT8,
         offsetof(nxt_socket_conf_t, discard_unsafe_fields),
     },
+
+    {
+        nxt_string("log_route"),
+        NXT_CONF_MAP_INT8,
+        offsetof(nxt_socket_conf_t, log_route),
+    },
 };
 
 
--- a/src/nxt_router.h
+++ b/src/nxt_router.h
@@ -197,6 +197,8 @@ typedef struct {
 
     nxt_str_t              body_temp_path;
 
+    uint8_t                log_route;  /* 1 bit */
+
     uint8_t                discard_unsafe_fields;  /* 1 bit */
 
     nxt_http_forward_t     *forwarded;


hongzhidao:
Hi,
Is the new PR what we just discussed to add to the requirement?
If yes, I'd suggest we separate the refactoring stuff from them. Like the below:
```
Router: factored out application creation loop.
Router: factored out listeners creation loop.
Fixed alignment (whitespace) in <nxt_router.h>.
```
Reviewers prefer to make as few decisions as possible. Since We might have a similar discussion in the future, I'd point it out for now. Changes are a decision of whether we need to do it. They are separated patches from the requirement, I know they are working well, but keeping PR as minimum and individual as possible is convenient for reviewers.

alejandro-colomar:
> Hi, Is the new PR what we just discussed to add to the requirement? If yes, I'd suggest we separate the refactoring stuff from them. Like the below:
> 
> ```
> Router: factored out application creation loop.
> Router: factored out listeners creation loop.
> Fixed alignment (whitespace) in <nxt_router.h>.
> ```

The third one is necessary.

I have now removed the first two.

> 
> Reviewers prefer to make as few decisions as possible. Since We might have a similar discussion in the future, I'd point it out for now. Changes are a decision of whether we need to do it. They are separated patches from the requirement, I know they are working well, but keeping PR as minimum and individual as possible is convenient for reviewers.
> 
> Btw, I'd suggest not introducing `nxt_http_log.h`. In Unit, we have the error and access log, if we let it be only the error log for http, the naming still be confusing for some developers.

Check again.  It is `nxt_log_http.h`:

```sh
alx@asus5775:~/src/nginx/unit/debug$ git show master..HEAD | grep -i http_log
alx@asus5775:~/src/nginx/unit/debug$ git show master..HEAD | grep -i log_http
+    nxt_log_http(task, r, NXT_LOG_NOTICE, NXT_LOG_HTTP_ROUTE_SELECTION,
+                nxt_log_http(task, r, NXT_LOG_NOTICE,
+                             NXT_LOG_HTTP_ROUTE_SELECTION,
+                nxt_log_http(task, r, NXT_LOG_NOTICE,
+                             NXT_LOG_HTTP_ROUTE_SELECTION,
    The macros nxt_log_http() and nxt_debug_http() accept a "condition",
+#include <nxt_log_http.h>
+    nxt_log_http_features_t         log_ft;
diff --git a/src/nxt_log_http.h b/src/nxt_log_http.h
+++ b/src/nxt_log_http.h
+#ifndef _NXT_LOG_HTTP_H_INCLUDED_
+#define _NXT_LOG_HTTP_H_INCLUDED_
+typedef enum nxt_log_http_features_e  nxt_log_http_features_t;
+#define nxt_log_http(task, r, level, cond, ...)                               \
+enum nxt_log_http_features_e {
+    NXT_LOG_HTTP_ROUTE_SELECTION = 0x01,
+#endif  /* _NXT_LOG_HTTP_H_INCLUDED_ */
+#include <nxt_log_http.h>
+    nxt_log_http_features_t     log_ft;
+        log_ft |= nxt_conf_get_boolean(conf) * NXT_LOG_HTTP_ROUTE_SELECTION;
+#include <nxt_log_http.h>
+    nxt_log_http_features_t  log_ft;
```

hongzhidao:
> Fixed alignment (whitespace) in <nxt_router.h>.

I thought it was a style patch, I wonder if it fixed existing issues?

>  It is nxt_log_http.h:

Unfortunately, it's still unclear. Since software architecture has a thing called "layer", from the naming, I would misunderstand it as a part of the log because of the prefix `nxt_log_`, but actually, it belongs to http. Could you move them into the `nxt_http.h` file?

alejandro-colomar:
> > Fixed alignment (whitespace) in <nxt_router.h>.
> 
> I thought it was a style patch, I wonder if it fixed existing issues?

It is for adding new members, which have a different alignment (the type name of the new member is longer than the existing type names.  If we don't align them before, the new addition wouldn't be aligned.  See:

```diff
$ git log -1 33a3bf27
commit 33a3bf2738b394ac0ea75e69f8b3dac5a723fcbc
Author: Alejandro Colomar <alx@nginx.com>
Date:   Thu Mar 2 12:44:26 2023 +0100

    Fixed alignment (whitespace) in <nxt_router.h>.
    
    This makes alignment more consistent in this file, and it will be
    necessary for inserting new elements in the next patch.
    
    Signed-off-by: Alejandro Colomar <alx@nginx.com>
```
```diff
$ git show 50f2b0c3 -- src/nxt_router.h
commit 50f2b0c31298d3c6250f322cdab7dc4d930829fb
Author: Alejandro Colomar <alx@nginx.com>
Date:   Thu Jan 26 15:07:12 2023 +0100

    Configuration: added "settings/http/log_route_selection".
    
    This adds configurability to the error log.  For now, it only has one
    toggle: "settings/http/log_route_selection".  It allows enabling and
    disabling logs related to how the router performs selection of the
    routes.
    
    The macros nxt_log_http() and nxt_debug_http() accept a "condition",
    which can be an OR of various features, and if any of them are enabled
    in the configuration, the log will be printed.  This OR will be useful
    for example for printing the request line, which we don't want to print
    by default, but we will want to print it if we're printing any of the
    conditional logs that can be printed in a request (right now there's
    only one, "log_route_selection", but in the future we will likely add at
    least one more for the URI rewrite feature).
    
    Link: <https://github.com/nginx/unit/issues/758>
    Link: <https://github.com/nginx/unit/pull/824>
    Link: <https://github.com/nginx/unit/pull/839>
    Suggested-by: Timo Stark <t.stark@nginx.com>
    Suggested-by: Liam Crilly <liam@nginx.com>
    Cc: Artem Konev <a.konev@f5.com>
    Cc: Zhidao Hong <z.hong@f5.com>
    Signed-off-by: Alejandro Colomar <alx@nginx.com>
    
    http

diff --git a/src/nxt_router.h b/src/nxt_router.h
index 078776ec..11a98864 100644
--- a/src/nxt_router.h
+++ b/src/nxt_router.h
@@ -12,6 +12,8 @@
 #include <nxt_runtime.h>
 #include <nxt_main_process.h>
 
+#include <nxt_log_http.h>
+
 typedef struct nxt_http_request_s  nxt_http_request_t;
 #include <nxt_application.h>
 
@@ -202,6 +204,8 @@ typedef struct {
     nxt_http_forward_t       *forwarded;
     nxt_http_forward_t       *client_ip;
 
+    nxt_log_http_features_t  log_ft;
+
 #if (NXT_TLS)
     nxt_tls_conf_t           *tls;
 #endif
```

> 
> > It is nxt_log_http.h:
> 
> Unfortunately, it's still unclear. Since software architecture has a thing called "layer", from the naming, I would misunderstand it as a part of the log because of the prefix `nxt_log_`, but actually, it belongs to http.

It really belongs to both.

We can discuss it again in the future if we add a new http log.

> Could you move them into the `nxt_http.h` file?

No, we can't.  I already tried a few days ago, but found the compilation error that I reported.  It would create a circular dependency between `<nxt_router.h>` and `<nxt_http.h>`.

hongzhidao:
1.  > nxt_log_http_features_t  log_ft;

Not sure if the naming is not clear enough. For any functions like nxt_var, nxt_log, the naming stood for they are already features, we don't need to add "features" to show what it is. And because of introducing the new type, we need to touch many lines of code in `nxt_router.h`, I prefer to avoid it. What about the alternative way?
(Note we only two clear features to be logged.)
```
/* Just put them into nxt_log.h, refer to NGINX */
#define NXT_LOG_HTTP_ROUTE_SELECTION    0x01
#define NXT_LOG_HTTP_URL_REWRITE              0x02 (in the future)
```
```
/* in related structures */
uint8_t           log_flags; /* or log_features, welcome to discuss them */
```

ac000:
> @hongzhidao
> 
> > 1. > nxt_log_http_features_t  log_ft;
> > 
> > Not sure if the naming is not clear enough. For any functions like nxt_var, nxt_log, the naming stood for they are already features, we don't need to add "features" to show what it is.

Naming's hard, _features_ was the best we came up with. The way I see it, _features_  in this case refers to the fact we can select different things to log. Or enabling different _features_ of the log if you like.
 
> An alternative name would be `nxt_log_http_flags_t`, but I'm not convinced by it. @ac000 what do you think of this?

_flags_ is not terrible, if a little on the nose...
 
> ...
> > What about the alternative way? (Note we only two clear features to be logged.)
> > ```
> > /* Just put them into nxt_log.h, refer to NGINX */
> > #define NXT_LOG_HTTP_ROUTE_SELECTION    0x01
> > #define NXT_LOG_HTTP_URL_REWRITE              0x02 (in the future)
> > ```
> >   
> > ```
> > /* in related structures */
> > uint8_t           log_flags; /* or log_features, welcome to discuss them */
> > ```
> 
> The good thing about using an `enum` is that a user can correlate both unambiguously, without needing a comment (and even with the comment, an enum is still more unambiguous). See how easy it is to correlate:
> 
> ```c
> $ grepc NXT_LOG_HTTP_ROUTE_SELECTION
> ./src/nxt_log_http.h:28:
> enum nxt_log_http_features_e {
>     NXT_LOG_HTTP_ROUTE_SELECTION = 0x01,
> };
> ```
> 
> So wherever we see that constant being used, we know that it will be used with variables of the `enum nxt_log_http_features_e` type.
> ...
> And wherever we see variables of the enum type, we know what values it will possibly hold.

Also, with _enum_ being a part of the language we can benefit from things like type safety.

I would definitely like it as an **enum**.

hongzhidao:
Let's rethink the function of conditional logging, I think we can make it more generic. In the original design of Unit, we plan to support TCP which can be used as a TCP proxy in the future. And for the tls/openssl part, they might also need similar functions. In short, we need a generic function of conditional logging for Unit.
Based on it, my suggestions are:
1. It makes sense to put them into `nxt_log.h`, it's an extensional function for log.
2. About the namings.
```
typedef enum {
NXT_LOG_HTTP_... = 0x01;
} nxt_log_feature_t;  /* don't plus `s` after feature. */

nxt_log_feature(task, level, flag, feature, fmt);

nxt_log_feature_t  log_flags; /* it's a bits flags */
```

alejandro-colomar:
> Let's rethink the function of conditional logging, I think we can make it more generic. In the original design of Unit, we plan to support TCP which can be used as a TCP proxy in the future. And for the tls/openssl part, they might also need similar functions. In short, we need a generic function of conditional logging for Unit. Based on it, my suggestions are:
> 
>     1. It makes sense to put them into `nxt_log.h`, it's an extensional function for log.

Makes sense.

> 
>     2. About the namings.
> 
> 
> ```
> typedef enum {
>     NXT_LOG_HTTP_... = 0x01;
> } nxt_log_feature_t;  /* don't plus `s` after feature. */
> ```

I think features with s makes more sense, because each of the `NXT_LOG_...` is one feature, so there are many feature*s*.  Does it make sense to you?

> 
> ```
> nxt_log_feature(task, level, feature, flags, fmt);
> ```

Makes sense.  Either `nxt_log_cond()` or `nxt_log_conditional()` or `nxt_log_feature()` make sense to me, so if you prefer `_feature()`, that's good for me.

> 
> ```
> nxt_log_feature_t  log_flags; /* it's a bits flags */
> ```

Also makes sense.

alejandro-colomar:
@hongzhidao Should we also move the configuration from `"settings/http/log_route_selection"` to `"settings/log_route_selection"`?

hongzhidao:
> I think features with s makes more sense, because each of the NXT_LOG_... is one feature, so there are many features.

Take other examples
```
typedef enum {
    NXT_CONF_VLDT_NULL    = 1 << NXT_CONF_NULL,
    NXT_CONF_VLDT_BOOLEAN = 1 << NXT_CONF_BOOLEAN,
    NXT_CONF_VLDT_INTEGER = 1 << NXT_CONF_INTEGER,
    NXT_CONF_VLDT_NUMBER  = (1 << NXT_CONF_NUMBER) | NXT_CONF_VLDT_INTEGER,
    NXT_CONF_VLDT_STRING  = 1 << NXT_CONF_STRING,
    NXT_CONF_VLDT_ARRAY   = 1 << NXT_CONF_ARRAY,
    NXT_CONF_VLDT_OBJECT  = 1 << NXT_CONF_OBJECT,
} nxt_conf_vldt_type_t;

typedef enum {
    NXT_HTTP_PROTO_H1 = 0,
    NXT_HTTP_PROTO_H2,
    NXT_HTTP_PROTO_DEVNULL,
} nxt_http_protocol_t;
```
We can just follow the style.

hongzhidao:
> I think features with s makes more sense, because each of the NXT_LOG_... is one feature, so there are many features.

Take other examples
```
typedef enum {
    NXT_CONF_VLDT_NULL    = 1 << NXT_CONF_NULL,
    NXT_CONF_VLDT_BOOLEAN = 1 << NXT_CONF_BOOLEAN,
    NXT_CONF_VLDT_INTEGER = 1 << NXT_CONF_INTEGER,
    NXT_CONF_VLDT_NUMBER  = (1 << NXT_CONF_NUMBER) | NXT_CONF_VLDT_INTEGER,
    NXT_CONF_VLDT_STRING  = 1 << NXT_CONF_STRING,
    NXT_CONF_VLDT_ARRAY   = 1 << NXT_CONF_ARRAY,
    NXT_CONF_VLDT_OBJECT  = 1 << NXT_CONF_OBJECT,
} nxt_conf_vldt_type_t;

typedef enum {
    NXT_HTTP_PROTO_H1 = 0,
    NXT_HTTP_PROTO_H2,
    NXT_HTTP_PROTO_DEVNULL,
} nxt_http_protocol_t;
```
We can just follow the style.

alejandro-colomar:
> > I think features with s makes more sense, because each of the NXT_LOG_... is one feature, so there are many features.
> 
> Take other examples
> 
> ```
> typedef enum {
>     NXT_CONF_VLDT_NULL    = 1 << NXT_CONF_NULL,
>     NXT_CONF_VLDT_BOOLEAN = 1 << NXT_CONF_BOOLEAN,
>     NXT_CONF_VLDT_INTEGER = 1 << NXT_CONF_INTEGER,
>     NXT_CONF_VLDT_NUMBER  = (1 << NXT_CONF_NUMBER) | NXT_CONF_VLDT_INTEGER,
>     NXT_CONF_VLDT_STRING  = 1 << NXT_CONF_STRING,
>     NXT_CONF_VLDT_ARRAY   = 1 << NXT_CONF_ARRAY,
>     NXT_CONF_VLDT_OBJECT  = 1 << NXT_CONF_OBJECT,
> } nxt_conf_vldt_type_t;
> 
> typedef enum {
>     NXT_HTTP_PROTO_H1 = 0,
>     NXT_HTTP_PROTO_H2,
>     NXT_HTTP_PROTO_DEVNULL,
> } nxt_http_protocol_t;
> ```
> 
> We can just follow the style.

Hmm, your right!

hongzhidao:
> Makes sense. Either nxt_log_cond() or nxt_log_conditional() or nxt_log_feature() make sense to me, so if you prefer _feature(), that's good for me.

My thought is that the fewer concepts or namings you create, the better. We can only introduce the new word `feature` for the feature. `conditional` is for users, but `feature` is for devs who are interested in reading source code.

hongzhidao:
> The pros are:
a. It's easy to get rid of unnecessary conditions like the above.
b. we don't need to introduce a new macro.
And NGINX has the same way.

        if (r->connection->log->log_level & NGX_LOG_DEBUG_HTTP) {
            ngx_log_error(NGX_LOG_NOTICE, r->connection->log, 0,
                          "\"%V\" does not match \"%V\"",
                          &code->name, &e->line);
        }

Btw, if you don't call it many places, you can do it like the way, then we'll see whether it's worth introducing a macro for it in the future, what do you think?

hongzhidao:
Btw, for the commit logs.
```
Router: rewrote while loop as for loop.
```
Router in Unit means the router process. The feature is about http route, I think we can use `HTTP` instead of `Router`.

alejandro-colomar:
Changes:

-  Rename `nxt_log_http()` -> `nxt_log_feature()`
-  Change `nxt_log_feature()` to accept `r->log_flags` instead of `r`
-  Reorder parameters, to put `r->log_flags` together with the flags (`cond`).

alejandro-colomar:
Changes: 

-  Reword commit messages: Router -> HTTP

alejandro-colomar:
> > The pros are:
> > a. It's easy to get rid of unnecessary conditions like the above.
> > b. we don't need to introduce a new macro.
> > And NGINX has the same way.
> 
> ```
>     if (r->connection->log->log_level & NGX_LOG_DEBUG_HTTP) {
>         ngx_log_error(NGX_LOG_NOTICE, r->connection->log, 0,
>                       "\"%V\" does not match \"%V\"",
>                       &code->name, &e->line);
>     }
> ```
> 
> Btw, if you don't call it many places, you can do it like the way, then we'll see whether it's worth introducing a macro for it in the future, what do you think?

I kept the macro, because otherwise it is a lot of nesting at call site, and it would become less readable, IMO.

hongzhidao:
The left unclear place is the change in the `nxt_http_route_handler`.
1. I suggest using `route->name.length != 0` instead of `start->u.pass->start != NULL`, since the latter might involve variables, I think the `route->name` is easier to understand.
2. It's better to optimize the code.
```
            if (start->u.pass->start == NULL) {
                nxt_log_http(task, r, NXT_LOG_NOTICE,
                             NXT_LOG_HTTP_ROUTE_SELECTION,
                             "selected: \"routes/%z/action\"", i);
            } else {
                nxt_log_http(task, r, NXT_LOG_NOTICE,
                             NXT_LOG_HTTP_ROUTE_SELECTION,
                             "selected: \"routes/%V/%z/action\"",
                             start->u.pass, i);
            }

It's the same as below.

            if (flags & feature) {
                nxt_log(...);

           } else {
                if (flags & feature) {
                      nxt_log(...);
                }
          }
```
When they are ready, could you post them into RB?

alejandro-colomar:
> The left unclear place is the change in the `nxt_http_route_handler`.
> 
>     1. I suggest using `route->name.length != 0` instead of `start->u.pass->start != NULL`, since the latter might involve variables, I think the `route->name` is easier to understand.

Yup, that was a mistake by me.  Fixed.

alejandro-colomar:
>     2. It's better to optimize the code.
> 
> 
> ```
>             if (start->u.pass->start == NULL) {
>                 nxt_log_http(task, r, NXT_LOG_NOTICE,
>                              NXT_LOG_HTTP_ROUTE_SELECTION,
>                              "selected: \"routes/%z/action\"", i);
>             } else {
>                 nxt_log_http(task, r, NXT_LOG_NOTICE,
>                              NXT_LOG_HTTP_ROUTE_SELECTION,
>                              "selected: \"routes/%V/%z/action\"",
>                              start->u.pass, i);
>             }
> 
> It's the same as below.
> 
>             if (flags & feature) {
>                 nxt_log(...);
> 
>            } else {
>                 if (flags & feature) {
>                       nxt_log(...);
>                 }
>           }
> ```

Changes:

-  Factored out conditional, to reduce one conditional (I guess the compiler would have deduced that anyway, but it reads simpler this way, I agree).

> 
> When they are ready, could you post them into RB?

I'll do that to-morrow.  :-)

alejandro-colomar:
Testing:

Since there are two identical lines about the `"fallback"` action, I applied a small cosmetic patch to differentiate them in my testing (but that was just for my testing):

```diff
alx@debian:~/src/nginx/unit/debug$ git diff
diff --git a/src/nxt_http_static.c b/src/nxt_http_static.c
index 781dce0f..21f99233 100644
--- a/src/nxt_http_static.c
+++ b/src/nxt_http_static.c
@@ -198,7 +198,7 @@ nxt_http_static(nxt_task_t *task, nxt_http_request_t *r,
             if (action->fallback != NULL) {
                 nxt_log_feature(task, NXT_LOG_NOTICE,
                                 r->log_flags, NXT_LOG_HTTP_ROUTE_SELECTION,
-                                "selected: \"fallback\" action");
+                                "select'd: \"fallback\" action");
                 return action->fallback;
             }
 
```

Clean up any old stuff, and install Unit:

```sh
alx@debian:~/src/nginx/unit/debug$ sudo killall unitd
unitd: no process found
alx@debian:~/src/nginx/unit/debug$ sudo rm -rf /opt/local/unit/router/non/
alx@debian:~/src/nginx/unit/debug$ git clean -dffx
Removing Makefile
Removing build/
alx@debian:~/src/nginx/unit/debug$ time ./configure --prefix=/opt/local/unit/router/non/ >/dev/null 2>&1; echo $?

real	0m3.316s
user	0m2.665s
sys	0m0.791s
0
alx@debian:~/src/nginx/unit/debug$ time make -j >/dev/null 2>&1; echo $?

real	0m3.847s
user	0m11.444s
sys	0m1.452s
0
alx@debian:~/src/nginx/unit/debug$ sudo make install
test -d /opt/local/unit/router/non//sbin 		|| install -d /opt/local/unit/router/non//sbin
install -p build/unitd /opt/local/unit/router/non//sbin/
test -d /opt/local/unit/router/non//var/lib/unit 		|| install -d /opt/local/unit/router/non//var/lib/unit
test -d /opt/local/unit/router/non//var/log/unit 		|| install -d /opt/local/unit/router/non//var/log/unit
test -d /opt/local/unit/router/non//var/run/unit 		|| install -d /opt/local/unit/router/non//var/run/unit
test -d /opt/local/unit/router/non//share/man/man8 		|| install -d /opt/local/unit/router/non//share/man/man8
install -p -m644 build/unitd.8 /opt/local/unit/router/non//share/man/man8/
```

Run `unitd` and listen to the log file:

```sh
alx@debian:~/src/nginx/unit/debug$ sudo /opt/local/unit/router/non/sbin/unitd 
2023/03/14 14:08:39 [info] 22271#22271 unit 1.30.0 started
alx@debian:~/src/nginx/unit/debug$ sudo tail -f /opt/local/unit/router/non/var/log/unit/unit.log &
[2] 22276
alx@debian:~/src/nginx/unit/debug$ 2023/03/14 14:08:39 [info] 22273#22273 discovery started
2023/03/14 14:08:39 [notice] 22273#22273 no modules matching: "/opt/local/unit/router/non//lib/unit/modules/*.unit.so" found
2023/03/14 14:08:39 [info] 22272#22272 controller started
2023/03/14 14:08:39 [notice] 22272#22272 process 22273 exited with code 0
2023/03/14 14:08:39 [info] 22275#22275 router started

```

Here's a config that shows a difficulty that users might have when understanding the matching feature in Unit, which @lcrilly and I discussed while talking about this feature:

```json
alx@debian:~/src/nginx/unit/debug$ cd ~/etc/unitd/
alx@debian:~/etc/unitd$ cat liam.json 
{
    "settings": {
        "http": {
            "log_route_selection": true
        }
    },

    "listeners": {
        "*:8080": {
            "pass": "routes/main"
        }
    },

    "routes" :{
        "main": [
            {
                "match": {
                    "headers": [
                        {
                            "user-agent": "curl/*",
                            "bypass": "!bypass"
                        }
                    ]
                },
                "action": {
                    "return": 400
                }
            },
            {
                "action": {
                    "share": [
                        "/home/alx/srv/www$uri",
                        "no-really-i-dont-exist"
                    ],
                    "fallback": {
                        "pass": "routes/err"
                    }
                }
            }
        ],
        "err": [
        	{
	            "action": {
    	            "return": 204
    	        }
            }
        ]
    }
}
```

(Re)configure:

```sh
alx@debian:~/etc/unitd$ sudo unit ctl http PUT /config <liam.json 
{
}
```

And now run some experiments:

```sh
alx@debian:~/etc/unitd$ curl -v localhost:8080/index.html
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET /index.html HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.88.1
> Accept: */*
> 
2023/03/14 14:10:10 [notice] 22275#22307 *7 http request line: "GET /index.html HTTP/1.1"
2023/03/14 14:10:10 [notice] 22275#22307 *7 selected: "routes/main/1/action"
< HTTP/1.1 200 OK
< Last-Modified: Thu, 26 May 2022 16:58:45 GMT
< ETag: "628f95a5-4"
< Content-Type: text/html
< Server: Unit/1.30.0
< Date: Tue, 14 Mar 2023 13:10:10 GMT
< Content-Length: 4
< 
idx
* Connection #0 to host localhost left intact
```
```sh
alx@debian:~/etc/unitd$ curl -v -H "bypass: Mr. Smith" localhost:8080/index.html
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET /index.html HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.88.1
> Accept: */*
> bypass: Mr. Smith
> 
< HTTP/1.1 400 Bad Request
< Content-Type: text/html
< Server: Unit/1.30.0
< Date: Tue, 14 Mar 2023 13:13:01 GMT
< 2023/03/14 14:13:01 [notice] 22275#22307 *14 http request line: "GET /index.html HTTP/1.1"
2023/03/14 14:13:01 [notice] 22275#22307 *14 selected: "routes/main/0/action"
Content-Length: 54
< 
<!DOCTYPE html><title>Error 400</title><p>Error 400.
* Connection #0 to host localhost left intact
```
```sh
alx@debian:~/etc/unitd$ curl -v localhost:8080/index.htm
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET /index.htm HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.88.1
> Accept: */*
> 
< 2023/03/14 14:13:22 [notice] 22275#22307 *15 http request line: "GET /index.htm HTTP/1.1"
2023/03/14 14:13:22 [notice] 22275#22307 *15 selected: "routes/main/1/action"
2023/03/14 14:13:22 [notice] 22275#22307 *15 selected: "fallback" action
2023/03/14 14:13:22 [notice] 22275#22307 *15 selected: "routes/err/0/action"
HTTP/1.1 204 No Content
< Server: Unit/1.30.0
< Date: Tue, 14 Mar 2023 13:13:22 GMT
< Content-Length: 0
< 
* Connection #0 to host localhost left intact
```
```sh
alx@debian:~/etc/unitd$ curl -v -X POST localhost:8080/index.htm
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> POST /index.htm HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.88.1
> Accept: */*
> 
< HTTP/1.1 204 No Content
< Server: Unit/1.30.0
< Date: Tue, 14 Mar 2023 13:14:41 GMT
< Content-Length: 0
< 
* Connection #0 to host localhost left intact
2023/03/14 14:14:41 [notice] 22275#22307 *16 http request line: "POST /index.htm HTTP/1.1"
2023/03/14 14:14:41 [notice] 22275#22307 *16 selected: "routes/main/1/action"
2023/03/14 14:14:41 [notice] 22275#22307 *16 select'd: "fallback" action
2023/03/14 14:14:41 [notice] 22275#22307 *16 selected: "routes/err/0/action"
```

All looks good.

```sh
alx@debian:~/etc/unitd$ fg
sudo tail -f /opt/local/unit/router/non/var/log/unit/unit.log	(wd: ~/src/nginx/unit/debug)
^C
alx@debian:~/etc/unitd$
```

@andrey-zelenkov @artemkonev @hongzhidao  This might interest you

PR #891, opened 2023-05-22
Man

alejandro-colomar:
- Docs: added libunit manual pages.                                              
- Add UTF-8 renderings of the libunit manual pages                               
                                                                        
This PR adds UTF-8 formatted renderings of the pages.  The intention is          
not to add those to Unit, but I have them in the PR to make it easy to           
comment on the rendered code.  Also, future versions of the PR will              
show the interdiff against old renderings, which is nice.  Prior to              
applying the patch, the renderings will be removed.                              

To see them, you can just run `less -R $file`, or `cat $file`, as you prefer.


alejandro-colomar:
[libunit_manual.pdf](https://github.com/nginx/unit/files/11553829/libunit_manual.pdf)

docs/man/man3/nxt_unit_done.3
```diff
@@ -0,0 +1,54 @@
...
+.TP
+.I ctx
+Context object to be destroyed.
+It should have been created with
+.MR nxt_unit_init 3
+or
+.MR nxt_unit_ctx_alloc 3 .
+.SH Description
+.MR nxt_unit_done 3
+destroys a context object in a Unit application.
```
andreivasiliu, line 25: Might be worth noting that this must happen before the parent context is destroyed (at least as far as I understand from looking at the code).

docs/man/man3/nxt_unit_request_read.3
```diff
@@ -0,0 +1,142 @@
...
+.I size
+bytes.
+.SH Description
+.MR nxt_unit_request_read 3
+copies the client request message body into a buffer.
+.PP
+This function does not read
+the header fields nor the request line,
+which are available in
+.IR req\->request .
```
andreivasiliu, line 35: This function may also read less than the Content-Length if a data_handler callback is defined, in which case the full body may or may not be available in request_handler.

If it is available, then it can be read, and data_handler() will not be called.

If it is not available, then data_handler() will be called, at which point it will be available.

docs/man/man3/nxt_unit_response_add_content.3
```diff
@@ -0,0 +1,91 @@
...
+.MR nxt_unit_response_add_content 3
+adds content to an HTTP response.
+.PP
+The response should have been initialized previously with
+.MR nxt_unit_response_init 3 .
+.PP
+The response should have enough remaining buffer size available.
+If there's not enough,
+more buffer can be allocated with
+.MR nxt_unit_response_buff_alloc 3 .
```
andreivasiliu, line 37: Typo in `buff`. There are a few more cases below.

docs/man/man3/nxt_unit_done.3
alejandro-colomar: Yep, worth mentioning.

docs/man/man3/nxt_unit_request_read.3
alejandro-colomar: Thanks!  Do you know if there's any way to know how much is available before actually attempting to read?  Or do you just read and then know if it's available?

docs/man/man3/nxt_unit_response_add_content.3
alejandro-colomar: Thanks!

docs/man/man3/nxt_unit_request_read.3
andreivasiliu: There isn't as far as I can see, although I'd like to know that myself as well.

PR #892, opened 2023-05-23
Fix bug in C example app

alejandro-colomar:
- Tests: fixed incorrect pointer assignment.
- Libunit: simplified nxt_unit_buf_read().

### DIFF
--- a/src/test/nxt_unit_app_test.c
+++ b/src/test/nxt_unit_app_test.c
@@ -257,8 +257,8 @@ greeting_app_request_handler(nxt_unit_request_info_t *req)
     if (r->content_length > 0) {
         p = copy(p, BODY, nxt_length(BODY));
 
-        res = nxt_unit_request_read(req, buf->free, buf->end - buf->free);
-        buf->free += res;
+        res = nxt_unit_request_read(req, p, buf->end - p);
+        p += res;
 
     }
 


src/nxt_unit.c
```diff
@@ -3224,23 +3224,20 @@ nxt_unit_buf_read(nxt_unit_buf_t **b, uint64_t *len, void *dst, size_t size)
...
+        copy = nxt_min(rest, buf->end - buf->free);
 
         p = nxt_cpymem(p, buf->free, copy);
 
         buf->free += copy;
         rest -= copy;
 
-        if (rest == 0) {
-            if (buf->end == buf->free) {
-                buf = nxt_unit_buf_next(buf);
```
hongzhidao, line 3237: If we need to touch the code, we should first ensure it has an issue.
And `buf` is not used after the block, right?
```
diff -r 3d1aecfc7ceb src/nxt_unit.c
--- a/src/nxt_unit.c    Wed May 24 14:01:28 2023 +0800
+++ b/src/nxt_unit.c    Wed May 24 14:33:18 2023 +0800
@@ -3233,10 +3233,6 @@
         rest -= copy;

         if (rest == 0) {
-            if (buf->end == buf->free) {
-                buf = nxt_unit_buf_next(buf);
-            }
-
             break;
         }
```
Do you know why?

hongzhidao: Actually, your code is equal to this:
```
        if (buf->end == buf->free) {
            buf = nxt_unit_buf_next(buf);
        }

        if (rest == 0) {
            break;
        }
```
==
```
        if (rest == 0) {
            break;
        }

         if (buf->end == buf->free) {
            buf = nxt_unit_buf_next(buf);
        }
```
==
```
        if (rest == 0) {
            break;
        }

         if (rest != 0) {
            buf = nxt_unit_buf_next(buf);
        }
```
==
```
        if (rest == 0) {
            break;
        }

         buf = nxt_unit_buf_next(buf);
```

But it's better to understand why `buf` is not used after the block before touching it.

src/nxt_unit.c
```diff
@@ -3224,23 +3224,20 @@ nxt_unit_buf_read(nxt_unit_buf_t **b, uint64_t *len, void *dst, size_t size)
     while (buf != NULL) {
         last_buf = buf;
 
-        copy = buf->end - buf->free;
-        copy = nxt_min(rest, copy);
+        copy = nxt_min(rest, buf->end - buf->free);
```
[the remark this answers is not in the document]
hongzhidao: 1. `buf` can be used for both request and response, the size of `end - free` means free size. Take a look at this.
```
#define nxt_buf_mem_used_size(bm)                                             \
    ((bm)->free - (bm)->pos)

#define nxt_buf_mem_free_size(bm)                                             \
    ((bm)->end - (bm)->free)
```
The logic here means how many bytes can be written.
```
/* It should be the free size of the buf. */
copy = buf->end - buf->free;

/* It should be no larger than the `rest` size. */
copy = nxt_min(rest, copy);
```
It's clear. And I'm afraid this doesn't improve the code readability.
```
copy = nxt_min(rest, buf->end - buf->free);
```
Since we need to understand what `buf->end - buf->free` means. Actually, it means free space to copy.
```
copy = buf->end - buf->free;
```
I would say both of them are fine, if yes, I suggest not touching it.

src/nxt_unit.c
```diff
@@ -3224,19 +3224,14 @@ nxt_unit_buf_read(nxt_unit_buf_t **b, uint64_t *len, void *dst, size_t size)
...
 
         p = nxt_cpymem(p, buf->free, copy);
 
         buf->free += copy;
         rest -= copy;
 
         if (rest == 0) {
-            if (buf->end == buf->free) {
```
hongzhidao, line 3236: 1. Is there any issue here?
2. Why can it be removed?
Removing it without explaining is confusing.

alejandro-colomar: The issue is that the code is quite confusing, apart from being dead code.
That code does nothing at all.  Why is it there?  Well, I explained here: <https://github.com/nginx/unit/pull/892#discussion_r1203818009>.  TLDR; the author probably started writing an optimization, and then realized it wasn't really optimizing and didn't finish writing the code, and forgot to remove it.

It can be removed, precisely because it's dead code:
The code has no side effects, and the store is a dead end (no reads beyond that point).

What are the doubts about if it can be removed?

src/nxt_unit.c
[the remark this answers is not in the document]
alejandro-colomar: Look at the next line:

```diff
-        copy = buf->end - buf->free;
-        copy = nxt_min(rest, copy);
+        copy = nxt_min(rest, buf->end - buf->free);
 
         p = nxt_cpymem(p, buf->free, copy);
```

We're copying _from_ `buf` into `p`.  We're _not_ copying into `buf`.

src/nxt_unit.c
alejandro-colomar: That's invalid, because `buf` may be `NULL`, and we don't want to store `NULL` in `*b`.

hongzhidao: `buf` return from `nxt_unit_buf_next()` maybe not `NULL`, right?

alejandro-colomar: Correct, so you'd need to do this:

```diff
diff -r 60c37d0e1df8 src/nxt_unit.c
--- a/src/nxt_unit.c    Wed May 24 14:01:28 2023 +0800
+++ b/src/nxt_unit.c    Thu May 25 00:14:15 2023 +0800
@@ -3235,6 +3235,7 @@
         if (rest == 0) {
             if (buf->end == buf->free) {
                 buf = nxt_unit_buf_next(buf);
+                if (buf != NULL) {
+                    last_buf = buf;
+                }
             }

             break;
```

alejandro-colomar: It's simpler to just remove the code, since it's not necessary, and it's not a clear optimization.

hongzhidao: I'm not sure if there is something missed here, it's better to be sure if there is an issue here.
If it's a dead code, this patch would be more concise.
```
diff -r 60c37d0e1df8 src/nxt_unit.c
--- a/src/nxt_unit.c    Wed May 24 14:01:28 2023 +0800
+++ b/src/nxt_unit.c    Thu May 25 00:23:54 2023 +0800
@@ -3221,7 +3221,7 @@
     buf = *b;
     last_buf = buf;

-    while (buf != NULL) {
+    while (buf != NULL && rest != 0) {
         last_buf = buf;

         copy = buf->end - buf->free;
@@ -3232,14 +3232,6 @@
         buf->free += copy;
         rest -= copy;

-        if (rest == 0) {
-            if (buf->end == buf->free) {
-                buf = nxt_unit_buf_next(buf);
-            }
-
-            break;
-        }
-
         buf = nxt_unit_buf_next(buf);
     }
```

But I still doubt it might miss something. And without understanding the libunit well, I have no idea.
I would prefer not to touch them.

hongzhidao: >  we don't want to store NULL in *b.

It depends on the intention of the function `nxt_unit_buf_read()`, not sure if it's allowed to set NULL for the argument `**b`. If you totally understand it, it's fine for me to know the design.

alejandro-colomar: Your patch is also valid, but it calls `nxt_unit_buf_next()` even if `rest == 0`, so it performs unnecessary operations.  It is less lines of code (mainly because you merge the conditions into one line), but it's not really less code.

But it's valid.  And no, you're not missing anything there.  It's easy to check that buf is not used after that point, so it's not possible to affect libunit behavior.

hongzhidao: The logic between 1 and 3 is different, right?

alejandro-colomar: > It depends on the intention of the function `nxt_unit_buf_read()`, not sure if it's allowed to set NULL for the argument `**b`. If you totally understand it, it's fine for me to know the design.

This function reads as many buffers from the list of buffers in `mmap_buf`, and copies the contents into `dst`.  When `dst` is full, it stops reading buffers.  If puts in `*b` the last buffer that was being read, so that a next call to this function will continue reading those buffers.

hongzhidao: If 1 is correct, isn't there an issue here? But 3 means there is no issue at all.

alejandro-colomar: They are different algorithms, but both can be used for the same purpose.

hongzhidao: Could you confirm this?
> If 1 is correct, isn't there an issue here?

alejandro-colomar: > Could you confirm this?
> 
> > If 1 is correct, isn't there an issue here?

I confirm there's not an issue there.

hongzhidao:
```
> diff -r 60c37d0e1df8 src/nxt_unit.c
--- a/src/nxt_unit.c    Wed May 24 14:01:28 2023 +0800
+++ b/src/nxt_unit.c    Thu May 25 00:14:15 2023 +0800
@@ -3235,6 +3235,7 @@
         if (rest == 0) {
             if (buf->end == buf->free) {
                 buf = nxt_unit_buf_next(buf);
+                if (buf != NULL) {
+                    last_buf = buf;
+                }
             }

             break;
```
The `buf` maybe not `NULL`, and it makes the value to `**b` be different from the previous way. Why isn't it an issue?
The `**b` has a different value in the two ways.

alejandro-colomar: > ```
> > diff -r 60c37d0e1df8 src/nxt_unit.c
> --- a/src/nxt_unit.c    Wed May 24 14:01:28 2023 +0800
> +++ b/src/nxt_unit.c    Thu May 25 00:14:15 2023 +0800
> @@ -3235,6 +3235,7 @@
>          if (rest == 0) {
>              if (buf->end == buf->free) {
>                  buf = nxt_unit_buf_next(buf);
> +                if (buf != NULL) {
> +                    last_buf = buf;
> +                }
>              }
> 
>              break;
> ```
> 
> The `buf` maybe not `NULL`, and it makes the value to `**b` be different from the previous way. Why isn't it an issue? The `**b` has a different value in the two ways.

Because in 3, the value in `*b` will be the last buffer, which is empty, and the next call to this function will detect that and load the next buffer.  And in 1, this function will load the next buffer so that the next call will not need to load it.  In any case, the next call will have the right buffer.

hongzhidao: With 3.
```
while (buf != NULL && rest != 0) {
        last_buf = buf;

        copy = buf->end - buf->free;
        copy = nxt_min(rest, copy);

        p = nxt_cpymem(p, buf->free, copy);

        buf->free += copy;
        rest -= copy;

        buf = nxt_unit_buf_next(buf);
    }
```
The buf which is the last one maybe not `NULL`, it still has space to store.

alejandro-colomar: > With 3.
> 
> ```
> while (buf != NULL && rest != 0) {
>         last_buf = buf;
> 
>         copy = buf->end - buf->free;
>         copy = nxt_min(rest, copy);
> 
>         p = nxt_cpymem(p, buf->free, copy);
> 
>         buf->free += copy;
>         rest -= copy;
> 
>         buf = nxt_unit_buf_next(buf);
>     }
> ```
> 
> The buf which is the last one maybe not `NULL`, it still has space to store.

last_buf contains a non-NULL input buffer `(buf != NULL)`.  We copy from it into `dst` as much as possible (`nxt_cpymem()`).  If there's more space in `dst`, advance to the next buffer (`buf = nxt_unit_buf_next(buf)` & `last_buf = buf`.

Keep going while there's sape in `dst` and there are still buffers to read.

hongzhidao: Based on the patch.
0:
```
while (buf != NULL) {
        copy = buf->end - buf->free;
        copy = nxt_min(rest, copy);

        p = nxt_cpymem(p, buf->free, copy);

        buf->free += copy;
        rest -= copy;

        if (rest == 0) {
            if (buf->end == buf->free) {
                buf = nxt_unit_buf_next(buf);
                last_buf = buf;
            }

            break;
        }

        buf = nxt_unit_buf_next(buf);
        last_buf = buf;
    }
```
1.
```
while (buf != NULL) {
        copy = buf->end - buf->free;
        copy = nxt_min(rest, copy);

        p = nxt_cpymem(p, buf->free, copy);

        buf->free += copy;
        rest -= copy;

        if (rest == 0) {
            if (buf->end == buf->free) {
                buf = nxt_unit_buf_next(buf);
                last_buf = buf;
            }

            break;
        }

        if (buf->end == buf->free) { /* always true */
            buf = nxt_unit_buf_next(buf);
            last_buf = buf;
        }
    }
```
2.
```
while (buf != NULL) {
        copy = buf->end - buf->free;
        copy = nxt_min(rest, copy);

        p = nxt_cpymem(p, buf->free, copy);

        buf->free += copy;
        rest -= copy;

        if (buf->end == buf->free) {
            buf = nxt_unit_buf_next(buf);
            last_buf = buf;
        }

        if (rest == 0) {
            break;
        }
    }
```
3.
```
while (buf != NULL && rest != 0) {
        copy = buf->end - buf->free;
        copy = nxt_min(rest, copy);

        p = nxt_cpymem(p, buf->free, copy);

        buf->free += copy;
        rest -= copy;

        if (buf->end == buf->free) {
            buf = nxt_unit_buf_next(buf);
            last_buf = buf;
        }
    }
```
4.
```
while (buf != NULL && rest != 0) {
        copy = buf->end - buf->free;
        copy = nxt_min(rest, copy);

        p = nxt_cpymem(p, buf->free, copy);

        buf->free += copy;
        rest -= copy;

        if (buf->end == buf->free) {
            buf = nxt_unit_buf_next(buf);
            last_buf = buf;
        }

        if (buf->end > buf->free) {
            last_buf = buf;
        }
    }
```

That's the final version.
```
while (buf != NULL && rest != 0) {
        copy = buf->end - buf->free;
        copy = nxt_min(rest, copy);

        p = nxt_cpymem(p, buf->free, copy);

        buf->free += copy;
        rest -= copy;

        if (buf->end == buf->free) {
            buf = nxt_unit_buf_next(buf);
        }

        last_buf = buf;
    }
```

hongzhidao: It's good to discuss it. My thoughts are:
1. The function is called for every application request, it's very important.
2. We didn't fix any issues or improve performance with the change.
Based on the risk and benefits, I prefer not to touch it.

hongzhidao: Just to compare the above and 3.
```
while (buf != NULL && rest != 0) {
        copy = buf->end - buf->free;
        copy = nxt_min(rest, copy);

        p = nxt_cpymem(p, buf->free, copy);

        buf->free += copy;
        rest -= copy;

        if (buf->end == buf->free) {
            buf = nxt_unit_buf_next(buf);
        }

        last_buf = buf;
    }
```
```
while (buf != NULL && rest != 0) {
        last_buf = buf;

        copy = buf->end - buf->free;
        copy = nxt_min(rest, copy);

        p = nxt_cpymem(p, buf->free, copy);

        buf->free += copy;
        rest -= copy;

        buf = nxt_unit_buf_next(buf);
    }
```

The 3 is allowed to point to an empty buffer. As I mentioned, it means nothing. But since your intention is to make this function better, we need to take into account the design of the function if we want to touch it.
So there are a few versions that are fine, I would suggest to keep it unchanged.

alejandro-colomar: These 5 suggestions add dead code, and change behavior regarding to NULL.  I don't like them.

I didn't intend to change the meaning of the function; only remove dead (and confusing) code.  And certainly not add more dead code, which would add to the confusion.

So I'll drop the patch.

ac000 (approved):

hongzhidao (approved):

alejandro-colomar:
Merged.

Issue #864, opened 2023-04-27
Python ASGI - RuntimeError: The server does not support "state" in the lifespan scope. 


cgearing:
Hi there,

I've been using nginx-unit for a FastAPI application, and everything has been working fine up until I attempted to use the [Lifespan State](https://asgi.readthedocs.io/en/latest/specs/lifespan.html#lifespan-state) part of the ASGI spec to store a reference to an object that I would like to maintain a reference to through the application lifecycle (it talks to the Google API to get authentication tokens which can seemingly take a long time sometimes).

It looks like currently nginx-unit doesn't support the state part of the lifespan functionality.

Was this purposefully left out or would you be open to PRs implementing this functionality?

ac000:
Hi Chris,

> Was this purposefully left out 

Not as far as I know, but lets check with some others.

Cc: @alejandro-colomar 
Cc: @hongzhidao

hawiliali:
I tested the pull request on a starltte app with a state holding multiple objects between db and http async clients
it's working

ac000:
The fix for this has been merged.

PR #894, opened 2023-05-27
Python: Add ASGI lifespan_state support.

synodriver:
fix https://github.com/nginx/unit/issues/864

The test script
```
import contextlib
from fastapi import FastAPI, Request


@contextlib.asynccontextmanager
async def lifespan(app):
    yield {"queue": []}


app = FastAPI(lifespan=lifespan)


@app.get("/")
async def _(request: Request):
    # print(request.scope["state"]["queue"])
    request.state.queue.append(1)
    return {"queue": request.state.queue}
```

### DIFF
--- a/src/python/nxt_python_asgi.c
+++ b/src/python/nxt_python_asgi.c
@@ -449,7 +449,8 @@ nxt_py_asgi_remove_reader(nxt_unit_ctx_t *ctx, nxt_unit_port_t *port)
 static void
 nxt_py_asgi_request_handler(nxt_unit_request_info_t *req)
 {
-    PyObject                *scope, *res, *task, *receive, *send, *done, *asgi;
+    PyObject                *scope, *res, *task, *receive, *send, *done, *asgi, *state, *newstate;
+    PyObject                *lifespan;
     PyObject                *stage2;
     nxt_python_target_t     *target;
     nxt_py_asgi_ctx_data_t  *ctx_data;
@@ -493,14 +494,38 @@ nxt_py_asgi_request_handler(nxt_unit_request_info_t *req)
     }
 
     req->data = asgi;
+    ctx_data = req->ctx->data;
     target = &nxt_py_targets->target[req->request->app_target];
-
-    scope = nxt_py_asgi_create_http_scope(req, target);
-    if (nxt_slow_path(scope == NULL)) {
+    lifespan = ctx_data->target_lifespans[req->request->app_target];
+    state = PyObject_GetAttr(lifespan, nxt_py_state_str);
+    if (nxt_slow_path(state == NULL)) {
+        nxt_unit_req_alert(req, "Python failed to get 'state' attribute");
         nxt_unit_request_done(req, NXT_UNIT_ERROR);
 
         goto release_done;
     }
+
+    newstate = PyDict_Copy(state);
+    if (nxt_slow_path(newstate == NULL)) {
+        nxt_unit_req_alert(req, "Python failed to call state.copy()");
+        nxt_unit_request_done(req, NXT_UNIT_ERROR);
+        Py_DECREF(state);
+        goto release_done;
+    }
+    Py_DECREF(state);
+
+    scope = nxt_py_asgi_create_http_scope(req, target);
+    if (nxt_slow_path(scope == NULL)) {
+        nxt_unit_request_done(req, NXT_UNIT_ERROR);
+        Py_DECREF(newstate);
+        goto release_done;
+    }
+
+    if (nxt_slow_path(PyDict_SetItem(scope, nxt_py_state_str, newstate) == -1)) {
+        Py_DECREF(newstate);
+        goto release_scope;
+    }
+    Py_DECREF(newstate);
 
     if (!target->asgi_legacy) {
         nxt_unit_req_debug(req, "Python call ASGI 3.0 application");
@@ -555,7 +580,6 @@ nxt_py_asgi_request_handler(nxt_unit_request_info_t *req)
         goto release_scope;
     }
 
-    ctx_data = req->ctx->data;
 
     task = PyObject_CallFunctionObjArgs(ctx_data->loop_create_task, res, NULL);
     if (nxt_slow_path(task == NULL)) {
--- a/src/python/nxt_python_asgi_lifespan.c
+++ b/src/python/nxt_python_asgi_lifespan.c
@@ -12,6 +12,8 @@
 #include <python/nxt_python_asgi.h>
 #include <python/nxt_python_asgi_str.h>
 
+#include <structmember.h>
+
 
 typedef struct  {
     PyObject_HEAD
@@ -25,6 +27,7 @@ typedef struct  {
     PyObject                *startup_future;
     PyObject                *shutdown_future;
     PyObject                *receive_future;
+    PyObject                *state;
 } nxt_py_asgi_lifespan_t;
 
 static PyObject *nxt_py_asgi_lifespan_target_startup(
@@ -41,6 +44,7 @@ static PyObject *nxt_py_asgi_lifespan_send_shutdown(
     nxt_py_asgi_lifespan_t *lifespan, int v, PyObject *dict);
 static PyObject *nxt_py_asgi_lifespan_disable(nxt_py_asgi_lifespan_t *lifespan);
 static PyObject *nxt_py_asgi_lifespan_done(PyObject *self, PyObject *future);
+void nxt_py_asgi_lifespan_dealloc(PyObject *self);
 
 
 static PyMethodDef nxt_py_asgi_lifespan_methods[] = {
@@ -48,6 +52,12 @@ static PyMethodDef nxt_py_asgi_lifespan_methods[] = {
     { "send",      nxt_py_asgi_lifespan_send,    METH_O,      0 },
     { "_done",     nxt_py_asgi_lifespan_done,    METH_O,      0 },
     { NULL, NULL, 0, 0 }
+};
+
+static PyMemberDef nxt_py_asgi_lifespan_members[] = {
+    { "state", T_OBJECT_EX, offsetof(nxt_py_asgi_lifespan_t, state), READONLY,
+      PyDoc_STR("lifespan.state") },
+    { NULL, 0, 0, 0, NULL }
 };
 
 static PyAsyncMethods nxt_py_asgi_async_methods = {
@@ -59,13 +69,14 @@ static PyTypeObject nxt_py_asgi_lifespan_type = {
 
     .tp_name      = "unit._asgi_lifespan",
     .tp_basicsize = sizeof(nxt_py_asgi_lifespan_t),
-    .tp_dealloc   = nxt_py_asgi_dealloc,
+    .tp_dealloc   = nxt_py_asgi_lifespan_dealloc,
     .tp_as_async  = &nxt_py_asgi_async_methods,
     .tp_flags     = Py_TPFLAGS_DEFAULT,
     .tp_doc       = "unit ASGI Lifespan object",
     .tp_iter      = nxt_py_asgi_iter,
     .tp_iternext  = nxt_py_asgi_next,
     .tp_methods   = nxt_py_asgi_lifespan_methods,
+    .tp_members   = nxt_py_asgi_lifespan_members,
 };
 
 
@@ -163,9 +174,24 @@ nxt_py_asgi_lifespan_target_startup(nxt_py_asgi_ctx_data_t *ctx_data,
     lifespan->shutdown_called = 0;
     lifespan->shutdown_future = NULL;
     lifespan->receive_future = NULL;
+    lifespan->state = NULL;
 
     scope = nxt_py_asgi_new_scope(NULL, nxt_py_lifespan_str, nxt_py_2_0_str);
     if (nxt_slow_path(scope == NULL)) {
+        goto release_future;
+    }
+
+    lifespan->state = PyDict_New();
+    if (nxt_slow_path(lifespan->state == NULL)) {
+        nxt_unit_req_error(NULL,
+                           "Python failed to create 'state' dict");
+        goto release_future;
+    }
+
+    if (nxt_slow_path(PyDict_SetItem(scope, nxt_py_state_str, lifespan->state) == -1)) {
+        nxt_unit_req_error(NULL,
+                           "Python failed to set 'scope.state' item");
+        Py_CLEAR(lifespan->state);
         goto release_future;
     }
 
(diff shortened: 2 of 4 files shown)


src/python/nxt_python_asgi.c
```diff
@@ -493,14 +494,35 @@ nxt_py_asgi_request_handler(nxt_unit_request_info_t *req)
...
     target = &nxt_py_targets->target[req->request->app_target];
+    lifespan = ctx_data->target_lifespans[req->request->app_target];
+    state = PyObject_GetAttr(lifespan, nxt_py_state_str);
+    if (nxt_slow_path(state == NULL)) {
+        nxt_unit_req_alert(req, "Python failed to get 'state' attribute");
+        nxt_unit_request_done(req, NXT_UNIT_ERROR);
 
+        goto release_send;
+    }
+    newstate = PyDict_Copy(state);
```
ac000, line 507: While there are too many blank lines already in the code base for my liking, I think one is deserved here after the if () {} block.

src/python/nxt_python_asgi.c
```diff
@@ -493,14 +494,35 @@ nxt_py_asgi_request_handler(nxt_unit_request_info_t *req)
...
+    }
+    Py_DECREF(state);
     scope = nxt_py_asgi_create_http_scope(req, target);
     if (nxt_slow_path(scope == NULL)) {
         nxt_unit_request_done(req, NXT_UNIT_ERROR);
-
+        Py_DECREF(newstate);
         goto release_done;
     }
+    if (nxt_slow_path(PyDict_SetItem(scope, nxt_py_state_str, newstate)==-1)) {
```
ac000, line 521: Perhaps a blank line before the if ()

Spaces around the ```==-1``` please. I.e ```) == -1)```

src/python/nxt_python_asgi.c
```diff
@@ -493,14 +494,35 @@ nxt_py_asgi_request_handler(nxt_unit_request_info_t *req)
...
+    }
+    newstate = PyDict_Copy(state);
+    if (nxt_slow_path(newstate == NULL)) {
+        nxt_unit_req_alert(req, "Python failed to call state.copy()");
+        nxt_unit_request_done(req, NXT_UNIT_ERROR);
+        Py_DECREF(state);
+        goto release_send;
+    }
+    Py_DECREF(state);
     scope = nxt_py_asgi_create_http_scope(req, target);
```
ac000, line 515: Maybe a blank line between these two.

src/python/nxt_python_asgi_lifespan.c
```diff
@@ -11,6 +11,7 @@
 #include <nxt_main.h>
 #include <python/nxt_python_asgi.h>
 #include <python/nxt_python_asgi_str.h>
+#include <structmember.h>
```
ac000, line 14: Hmm, what is this? Oh, it's actually from Python...

Maybe a blank line before it to show it's separate from the internal Unit includes.

src/python/nxt_python_asgi_lifespan.c
```diff
@@ -50,6 +52,11 @@ static PyMethodDef nxt_py_asgi_lifespan_methods[] = {
     { NULL, NULL, 0, 0 }
 };
 
+static PyMemberDef nxt_py_asgi_lifespan_members[] = {
+        {"state", T_OBJECT_EX, offsetof(nxt_py_asgi_lifespan_t, state), READONLY, PyDoc_STR("lifespan.state")},
+        {NULL, 0, 0, 0, NULL}
+};
```
ac000, line 58: Not overly keen on how that looks.

Maybe it would be better to specify the struct members and lay it out to match the structure. By which I mean something like [this](https://elixir.bootlin.com/linux/latest/source/drivers/dma/sun4i-dma.c#L1295).

src/python/nxt_python_asgi_lifespan.c
```diff
@@ -163,12 +171,26 @@ nxt_py_asgi_lifespan_target_startup(nxt_py_asgi_ctx_data_t *ctx_data,
...
         goto release_future;
     }
 
+    lifespan->state = PyDict_New();
+    if (nxt_slow_path(lifespan->state == NULL)) {
+        nxt_unit_req_error(NULL,
+                           "Python failed to create 'state' dict");
+        goto release_future;
+    }
+    if (nxt_slow_path(PyDict_SetItem(scope, nxt_py_state_str, lifespan->state) == -1)) {
```
ac000, line 187: Maybe a blank here...

ac000 (changes requested):
Thanks the pull request!

Looks generally fine, just some minor comments below.

Also, could you make the commit subject something like

```Python: Add ASGI lifespan_state support.```

Yes, for some reason we need to put a ```.``` on the subject line.

It would also be helpful if you could fill out the commit message with a little detail on what this actually adds and why it's needed.

Cheers,
Andrew

File: src/python/nxt_python_asgi_lifespan.c (hunk 5 above)
Comment (synodriver) [reply src/python/nxt_python_asgi_lifespan.c:58]: but ```nxt_py_asgi_lifespan_methods``` use the same way, only change here makes it looks wired

Comment (ac000) [reply src/python/nxt_python_asgi_lifespan.c:59]: The two below are designated initialised, though they aren't arrays... OK, go with your version, but could you format it as

```c
static PyMemberDef nxt_py_asgi_lifespan_members[] = {                           
    { "state", T_OBJECT_EX, offsetof(nxt_py_asgi_lifespan_t, state), READONLY,
      PyDoc_STR("lifespan.state") },                                            
    { NULL, 0, 0, 0, NULL }                                                     
};
```

(spaces around {}, 4 char indent, PyDoc_STR() aligmnent...)

Thanks.

ac000:
Good stuff, nearly there!

Could you squash the two commits together? Don't worry if you're unsure how, I can do it at merge time.

Heh, I actually meant could you make ```Python: Add ASGI lifespan_state support.``` the subject line of the git commit itself... otherwise it'll get blocked...

Could you also provide a few sentences in the git commit message just giving an overview of what this lifespan thing is and why it's needed. It''s not something that I'm personally familiar with.

Once you're done, could you rebase this against current master?

If you're unsure howto modify existing commits (git commit --amend and git rebase [-i] are your friends) then just drop the text here and I'll add it to the commit message.

Cheers,
Andrew

synodriver:
Ok, I'll have a try.

ac000:
Great thanks.

One last _optional_ thing. If you want to have a read (it's not long) of the [Developer Certificate of Origin](https://developercertificate.org/) and provide your ```Signed-off-by: synodriver <diguohuangjiajinweijun@gmail.com> ``` feel free.

No need to amend and repush your commit again, just stick it as a comment here and I'll add it to the commit.

synodriver:
Signed-off-by: synodriver <diguohuangjiajinweijun@gmail.com>

synodriver:
Sorry for the late reply. Is this ok?

ac000:
Perfect, thanks!

Feel free to do the same with your other PR.

ac000:
Now, I've tested the above application and another one that doesn't use this stuff (i think)

```python
from fastapi import FastAPI

app = FastAPI()

@app.get("/")
async def root():
    return {"message": "Hello, World!"}
```

and they both work fine.

However unfortunately this is failing on our tests. From the repository root...

```shell
$ ./pytest test/test_asgi_application.py
```

Though from what I remember it was a bit of a faff getting it going. But this is what it's failing on

```python
async def application(scope, receive, send):
    assert scope['type'] == 'http'

    body = b''
    while True:
        m = await receive()
        body += m.get('body', b'')
        if not m.get('more_body', False):
            break

    headers = scope.get('headers', [])

    def get_header(n):
        res = []
        for h in headers:
            if h[0] == n:
                res.append(h[1])
        return b', '.join(res)

    await send(
        {
            'type': 'http.response.start',
            'status': 200,
            'headers': [
                (b'content-type', get_header(b'content-type')),
                (b'content-length', str(len(body)).encode()),
                (b'request-method', scope['method'].encode()),
                (b'request-uri', scope['path'].encode()),
                (b'http-host', get_header(b'host')),
                (b'http-version', scope['http_version'].encode()),
                (b'asgi-version', scope['asgi']['version'].encode()),
                (b'asgi-spec-version', scope['asgi']['spec_version'].encode()),
                (b'scheme', scope['scheme'].encode()),
                (b'custom-header', get_header(b'custom-header')),
            ],
        }
    )

    await send({'type': 'http.response.body', 'body': body})
```

Specifically the

```python
assert scope['type'] == 'http'
```

line. With

```
2023/05/30 17:07:09 [info] 0#45749 [unit] ASGI Lifespan processing exception
Traceback (most recent call last):
  File "/home/andrew/src/unit/test/python/variables/asgi.py", line 2, in application
    assert scope['type'] == 'http'
AssertionError
2023/05/30 17:07:09 [notice] 45641#45641 app process 45642 exited with code 0
2023/05/30 17:07:09 [notice] 45637#45637 process 45641 exited with code 0
2023/05/30 17:07:24 [alert] 45749#45749 [unit] #20: Python failed to get 'state' attribute
2023/05/30 17:07:24 [alert] 45749#45749 [unit] Python failed to call 'loop.call_soon'
AttributeError: 'unit._asgi_lifespan' object has no attribute 'state'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/lib64/python3.11/asyncio/base_events.py", line 765, in call_soon
    handle = self._call_soon(callback, args, context)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/asyncio/base_events.py", line 781, in _call_soon
    handle = events.Handle(callback, args, self, context)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SystemError: <class 'asyncio.events.Handle'> returned a result with an exception set
```

So I suppose the question is. Is this a valid failure? or do our tests need updating for this new feature?

synodriver:
Oh, it seems that this asgi app does not support lifespan protocol which leads to this error. The ```AssertionError``` is fine and server should continue without lifespan support. I'll dig deeper to see why it can not get state from lifespan object.

ac000:
On Tue, 30 May 2023 10:46:27 -0700
synodriver ***@***.***> wrote:

> Oh, it seems that this asgi app does not support lifespan protocol which leads to this error. The ```AssertionError``` is fine and server should continue without lifespan support. I'll dig deeper to see why it can not get state from lifespan object.

Great, thanks!

ac000:
Thanks. Yeah, that looks better.

Only one small problem left, it fails to build on older Python i.e < 3.7

```shell
gcc -c -pipe -fPIC -fvisibility=hidden -O -W -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wmissing-prototypes -Werror -g   -DNXT_PYTHON_MOUNTS_H=\"nxt_python3_mounts.h\" \
           -I src -I build/include -I/usr/include/python3.6m -I/usr/include/python3.6m \
-MMD -MF build/src/python/nxt_python_wsgi-python3.dep -MT build/src/python/nxt_python_wsgi-python3.o \
-o build/src/python/nxt_python_wsgi-python3.o src/python/nxt_python_wsgi.c
src/python/nxt_python_asgi_lifespan.c:58:7: error: initialization discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
   58 |     { "state", T_OBJECT_EX, offsetof(nxt_py_asgi_lifespan_t, state), READONLY,
      |       ^~~~~~~
In file included from /usr/include/python3.6m/Python.h:54,
                 from src/python/nxt_python.h:10,
                 from src/python/nxt_python_asgi_lifespan.c:7:
src/python/nxt_python_asgi_lifespan.c:59:17: error: initialization discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
   59 |       PyDoc_STR("lifespan.state") },
      |                 ^~~~~~~~~~~~~~~~
/usr/include/python3.6m/pymacro.h:73:24: note: in definition of macro ‘PyDoc_ST’
   73 | #define PyDoc_STR(str) str
      |                        ^~~
cc1: all warnings being treated as errors
```

This is due to the ```PyMemberDef``` structure before Python 3.7 using ```char *``` for the ```name``` and ```doc``` members.

Leave this with me, I'll see what the best way to handle this is.

synodriver:
Actually these python versions are EOL now, and many third packages have dropped support for them. So maybe unit can do the same thing？

ac000:
On Tue, 30 May 2023 21:46:10 -0700
synodriver ***@***.***> wrote:

> Actually these python versions are EOL now, and many third packages have dropped support for them. So maybe unit can do the same thing？

Heh, unfortunately we can't just drop support for older versions, we
still technically support 2.7!

In this case it's simple enough to retain support for <3.7.

Just need to see what looks the best, whether we use some macro magic
or #if python version checks etc...

I'm also curious to see how these things were set in <3.7 by other
applications, did they just cast to (char *)? Or is there actually some
Python macro for setting these things?... need to investigate.

synodriver:
Cython use their own macros to support different python versions, and cast to char *, for example, [this cython generated binding](https://github.com/synodriver/python-bz3/blob/f20aa9f/bz3/backends/cython/_bz3.c#L27745-L27748), however, [```PyDoc_STR```](https://docs.python.org/3/c-api/intro.html?highlight=pydoc_str#c.PyDoc_STR) is recommended for newer python versions.

ac000 (approved):

ac000:
This was a combination of two issues

1) That Python <3.7 used char * for members of the PyMemberDef structure.

2) We compile with -Wwrite-strings which gives string constants the type const char * and we got the warning when trying to assign them to a non-const pointer.

I've fixed that the following way and folded it into your patch

```diff
diff --git a/src/python/nxt_python_asgi_lifespan.c b/src/python/nxt_python_asgi_lifespan.c
index 4daa1332..9fb361e1 100644
--- a/src/python/nxt_python_asgi_lifespan.c
+++ b/src/python/nxt_python_asgi_lifespan.c
@@ -52,12 +52,26 @@ static PyMethodDef nxt_py_asgi_lifespan_methods[] = {
     { "_done",     nxt_py_asgi_lifespan_done,    METH_O,      0 },
     { NULL, NULL, 0, 0 }
 };
 
 static PyMemberDef nxt_py_asgi_lifespan_members[] = {
-    { "state", T_OBJECT_EX, offsetof(nxt_py_asgi_lifespan_t, state), READONLY,
-      PyDoc_STR("lifespan.state") },
+    {
+#if PY_VERSION_HEX >= NXT_PYTHON_VER(3, 7)
+        .name   = "state",
+#else
+        .name   = (char *)"state",
+#endif
+        .type   = T_OBJECT_EX,
+        .offset = offsetof(nxt_py_asgi_lifespan_t, state),
+        .flags  = READONLY,
+#if PY_VERSION_HEX >= NXT_PYTHON_VER(3, 7)
+        .doc    = PyDoc_STR("lifespan.state")
+#else
+        .doc    = (char *)PyDoc_STR("lifespan.state")
+#endif
+    },
+
     { NULL, 0, 0, 0, NULL }
 };
 
 static PyAsyncMethods nxt_py_asgi_async_methods = {
     .am_await = nxt_py_asgi_await,
```

It's passed our internal testing and has been committed and pushed out.

Thanks!

lcrilly:
Thanks @synodriver ! 🚀

PR #900, opened 2023-06-06
Tools: setup-unit: Add restart command, + minor improvements

alejandro-colomar:
- Tools: setup-unit: using $0 is simpler.
- Tools: setup-unit: forcing full lines from ps(1).
- Tools: setup-unit: added restart command.
- Tools: setup-unit: added flags for cleaning the log file and state dir.


alejandro-colomar:
To be clear, unitd should not be restarted normally, but when developing, after recompilation, it's useful to restart unitd with the latest build.

PR #917, opened 2023-08-09
Add WebAssembly language module support

ac000:
This adds support to Unit for running WebAssembly (Wasm) modules.

It should be noted that this is being classed as a **Technology Preview**.

Wasm provides a secure sandboxed environment for running code. Code runs in a virtual machine with limited access to outside resources and any attempt to access memory not allocated to the module will result in a trap and the immediate termination of execution of the module.

While Wasm originated on the web browser side of things, influenced by asm.js, there are server side runtimes available and with the combination of WebAssembly System Interface (WASI) you can have a POSIX like environment to develop in.

This initial Wasm support makes use of [Wasmtime](https://wasmtime.dev/), however it has been designed with the ability to support different runtimes in mind (initially at build time, but perhaps eventually also at runtime).

There are actually sort of two parts of this (different sides of the same coin if you will).

This pull request deals solely with the Unit language module (runtime) side of things, what runs Wasm modules.

Then there is the Wasm modules side of things. There is a parallel project, [libunit-wasm](https://github.com/nginx/unit-wasm), this is done as a separate project as it is likely to be a rapidly evolving entity, to help with developing Wasm modules for Unit in C and Rust.

This pull request consists of the following commits

- Index initialise the nxt_app_msg_prefix array.

A preparatory patch that makes a structure array initialisation much clearer.

- Wasm: Register a new WebAssembly language module type.

This registers a new module type of 'NXT_APP_WASM'

- Wasm: Add core configuration data structure.

This adds the core Wasm configuration structure which is required by subsequent commits.

- Wasm: Add the core of initial WebAssembly language module support.

This adds the core of WebAssembly language module support under a new src/wasm directory.

- Wasm: Wire the Wasm language module up to the build system.

Allow the Wasm language module to be configured and built.

- Wasm: Wire up Wasm language module support to the config system.

Allow the Wasm language module to be configured via the Unit configuration.

- Wasm: Add support for directory access.

Allow to specify directories that can be made available for access by the Wasm modules.

- [WASM DEBUG] NOT TO BE COMMITTED.

**Ignore this patch**. It's still useful while we work on the Rust unit-wasm support.

As always, see individual commit messages for more details...

Each successive commit has been build tested.

PR #922, opened 2023-08-18
Fix WASM module caveats (post 1.31)

alejandro-colomar:
This is a set of fixes for the caveats noted in the review in <https://github.com/nginx/unit/pull/917>, which we agreed to fix after the release.

I prepared the PR already, but please ignore it until we release 1.31.

### DIFF
--- a/auto/modules/go
+++ b/auto/modules/go
@@ -115,14 +115,16 @@ ${NXT_GO}-install: ${NXT_GO}-install-src ${NXT_GO}-install-env
     GOPATH=\$(DESTDIR)\$(GOPATH) ${NXT_GO} build ${NXT_GO_PKG}
 
 ${NXT_GO}-install-src:
-	install -d \$(DESTDIR)\$(NXT_GO_DST)/src/${NXT_GO_PKG}
+	test -d \$(DESTDIR)\$(NXT_GO_DST)/src/${NXT_GO_PKG} \
+		|| install -d \$(DESTDIR)\$(NXT_GO_DST)/src/${NXT_GO_PKG}
 	install -p -m644 ./go/* \$(DESTDIR)\$(NXT_GO_DST)/src/${NXT_GO_PKG}/
 
 ${NXT_GO}-install-env: \$(DESTDIR)\$(NXT_GO_DST)/src/${NXT_GO_PKG}/env.go \
     ${NXT_VERSION_H} ${NXT_BUILD_DIR}/lib/${NXT_LIB_UNIT_STATIC}
 
 \$(DESTDIR)\$(NXT_GO_DST)/src/${NXT_GO_PKG}/env.go:
-	install -d \$(DESTDIR)\$(NXT_GO_DST)/src/${NXT_GO_PKG}
+	test -d \$(DESTDIR)\$(NXT_GO_DST)/src/${NXT_GO_PKG} \
+		|| install -d \$(DESTDIR)\$(NXT_GO_DST)/src/${NXT_GO_PKG}
 	$echo "package unit" > \$@
 	$echo "/*" >> \$@
 	$echo "#cgo CFLAGS: ${CFLAGS} ${NXT_CC_OPT}" >> \$@
--- a/auto/modules/java
+++ b/auto/modules/java
@@ -535,10 +535,12 @@ ${NXT_JAVA_MODULE}-install: \\
     $NXT_BUILD_DIR/$NXT_UNIT_JAR \\
     $NXT_BUILD_DIR/$NXT_WS_API_JAR \\
     java-shared-install
-	install -d \$(DESTDIR)$NXT_MODULESDIR
+	test -d \$(DESTDIR)$NXT_MODULESDIR \
+		|| install -d \$(DESTDIR)$NXT_MODULESDIR
 	install -p $NXT_BUILD_DIR/lib/unit/modules/${NXT_JAVA_MODULE}.unit.so \\
 		\$(DESTDIR)$NXT_MODULESDIR/
-	install -d \$(DESTDIR)$NXT_JARS
+	test -d \$(DESTDIR)$NXT_JARS \
+		|| install -d \$(DESTDIR)$NXT_JARS
 	install -p -m 0644 $NXT_BUILD_DIR/$NXT_UNIT_JAR \$(DESTDIR)$NXT_JARS/
 	install -p -m 0644 $NXT_BUILD_DIR/$NXT_WS_API_JAR \$(DESTDIR)$NXT_JARS/
 
@@ -629,7 +631,8 @@ if ! grep ^java-shared-install: $NXT_MAKEFILE 2>&1 > /dev/null; then
 .PHONY: java-shared-uninstall
 
 java-shared-install: $NXT_JAVA_INSTALL_JARS
-	install -d \$(DESTDIR)$NXT_JARS
+	test -d \$(DESTDIR)$NXT_JARS \
+		|| install -d \$(DESTDIR)$NXT_JARS
 	install -p -m 0644 $NXT_JAVA_INSTALL_JARS \$(DESTDIR)$NXT_JARS/
 
 java-shared-uninstall:
--- a/auto/modules/perl
+++ b/auto/modules/perl
@@ -194,7 +194,8 @@ $NXT_BUILD_DIR/lib/unit/modules/${NXT_PERL_MODULE}.unit.so:	$nxt_objs
 install: ${NXT_PERL_MODULE}-install
 
 ${NXT_PERL_MODULE}-install: ${NXT_PERL_MODULE} install-check
-	install -d \$(DESTDIR)$NXT_MODULESDIR
+	test -d \$(DESTDIR)$NXT_MODULESDIR \
+		|| install -d \$(DESTDIR)$NXT_MODULESDIR
 	install -p $NXT_BUILD_DIR/lib/unit/modules/${NXT_PERL_MODULE}.unit.so \\
 		\$(DESTDIR)$NXT_MODULESDIR/
 
--- a/auto/modules/php
+++ b/auto/modules/php
@@ -275,7 +275,8 @@ $NXT_BUILD_DIR/lib/unit/modules/${NXT_PHP_MODULE}.unit.so:	$nxt_objs
 install: ${NXT_PHP_MODULE}-install
 
 ${NXT_PHP_MODULE}-install: ${NXT_PHP_MODULE} install-check
-	install -d \$(DESTDIR)$NXT_MODULESDIR
+	test -d \$(DESTDIR)$NXT_MODULESDIR \
+		|| install -d \$(DESTDIR)$NXT_MODULESDIR
 	install -p $NXT_BUILD_DIR/lib/unit/modules/${NXT_PHP_MODULE}.unit.so \\
 		\$(DESTDIR)$NXT_MODULESDIR/
 
--- a/auto/modules/python
+++ b/auto/modules/python
@@ -223,7 +223,8 @@ $NXT_BUILD_DIR/lib/unit/modules/${NXT_PYTHON_MODULE}.unit.so:	$nxt_objs
 install: ${NXT_PYTHON_MODULE}-install
 
 ${NXT_PYTHON_MODULE}-install: ${NXT_PYTHON_MODULE} install-check
-	install -d \$(DESTDIR)$NXT_MODULESDIR
+	test -d \$(DESTDIR)$NXT_MODULESDIR \
+		|| install -d \$(DESTDIR)$NXT_MODULESDIR
 	install -p $NXT_BUILD_DIR/lib/unit/modules/${NXT_PYTHON_MODULE}.unit.so \\
 		\$(DESTDIR)$NXT_MODULESDIR/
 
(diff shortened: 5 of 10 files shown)


ac000 (commented):
> Fixes: https://github.com/nginx/unit/commit/6a211e2b7468a2524715b7f00a24aa460481bea7 ("Wasm: Add the core of initial WebAssembly language module support.")

I don't think that's appropriate as it's not broken...

Not a big fan of this change in general.

ac000 (commented):
> Fixes: https://github.com/nginx/unit/commit/6a211e2b7468a2524715b7f00a24aa460481bea7 ("Wasm: Add the core of initial WebAssembly language module support.")

Again, not broken. If it was actually broken, it wouldn't have been released like that.

Obviously in this case sizeof() or offsetof() will do the same thing.

Also with your code example, I added a new section that more closely mimics what I'm actually doing

```C
        s = xmalloc(1024);      /* I have some large chunk of memory */
        strcpy((char *)s + sizeof(struct s), "Hello, addr+offset!");
        p = (char *)s + sizeof(struct s);
        puts(p);
        free(s);
```

```
$ ./off 
sizeof: 8
offsetof: 5

With sizeof():
lo, sizeof!

With offsetof(3):
Hello, offsetof(3)!

With addr+offset:
Hello, addr+offset!
```

This is also easily fixed by adding some padding to the structure...

Before

```C
struct s {
        int                        i;                    /*     0     4 */
        char                       c;                    /*     4     1 */
        char                       fam[];                /*     5     0 */

        /* size: 8, cachelines: 1, members: 3 */
        /* padding: 3 */
        /* last cacheline: 8 bytes */
};
```

After __pad member added

```C
struct s {
        int                        i;                    /*     0     4 */
        char                       c;                    /*     4     1 */
        char                       __pad[3];             /*     5     3 */
        char                       fam[];                /*     8     0 */

        /* size: 8, cachelines: 1, members: 4 */
        /* last cacheline: 8 bytes */
};
```
I also changed all the mallocs to just allocate 1024 otherwise we get

```
$ gcc -Wall -Wextra -O2 -g -o off off.c
off.c: In function ‘main’:
off.c:39:9: warning: ‘__builtin_memcpy’ forming offset [22, 27] is out of the bounds [0, 22] [-Warray-bounds=]
   39 |         strcpy(s->fam, "Hello, offsetof(3)!");
      |
```

and a segfault from the malloc 

```
malloc(): corrupted top size
Aborted (core dumped)
```

And valgrind is generally very unhappy with that above strcpy.

```
$ ./off
sizeof: 8
offsetof: 8

With sizeof():
Hello, sizeof!

With offsetof(3):
Hello, offsetof(3)!

With addr+offset:
Hello, addr+offset!
```

I'm also unsure if the offsetoff() makes things a little less clear, particularly in the second case where I'm just wanting to start writing where the fam starts, but I don't access it as ->fam.

alejandro-colomar:
> > Fixes: [6a211e2](https://github.com/nginx/unit/commit/6a211e2b7468a2524715b7f00a24aa460481bea7) ("Wasm: Add the core of initial WebAssembly language module support.")
> 
> Again, not broken. If it was actually broken, it wouldn't have been released like that.

Yeah, I just used `Fixes:` because there's not standard `Improves:` or `Hardens:` that I know of.  Don't mean it was broken as in "it produces wrong behavior", but rather as in "it's not the most robust code that could be written".  Can agree to remove that tag.

The only really broken commits are the ones that did `install -d` without `test -d`.

> 
> Obviously in this case sizeof() or offsetof() will do the same thing.

Agree.  I also specified this in the commit message, to be precise.

[I'll reorder your reply below, to organize my answer.]


> This is also easily fixed by adding some padding to the structure...
> 
> Before
> 
> ```c
> struct s {
>         int                        i;                    /*     0     4 */
>         char                       c;                    /*     4     1 */
>         char                       fam[];                /*     5     0 */
> 
>         /* size: 8, cachelines: 1, members: 3 */
>         /* padding: 3 */
>         /* last cacheline: 8 bytes */
> };
> ```
> 
> After __pad member added
> 
> ```c
> struct s {
>         int                        i;                    /*     0     4 */
>         char                       c;                    /*     4     1 */
>         char                       __pad[3];             /*     5     3 */
>         char                       fam[];                /*     8     0 */
> 
>         /* size: 8, cachelines: 1, members: 4 */
>         /* last cacheline: 8 bytes */
> };
> ```

This is as brittle as it is now.  The size of the padding may change when new fields are added.  Maybe tomorrow the padding is 1, and we need to update this field.  It's doable, but I'd prefer not having to do it.

If we're going that way, we could also use `_Alignas()` to move the `fam[]`; it's a bit sad that `struct s` is incomplete at that point and we can't just use `_Alignas(struct s)`.  Maybe the explicit padding combined with a pragma enabling `-Werror=padding` would be good --since we would ensure that there's no more padding than the one we wrote--, but at that point, we're complicating it more than offsetof() ever would, I think.  And we would be wasting bytes needlessly.

> 
> I also changed all the mallocs to just allocate 1024 otherwise we get
> 
> ```
> $ gcc -Wall -Wextra -O2 -g -o off off.c
> off.c: In function ‘main’:
> off.c:39:9: warning: ‘__builtin_memcpy’ forming offset [22, 27] is out of the bounds [0, 22] [-Warray-bounds=]
>    39 |         strcpy(s->fam, "Hello, offsetof(3)!");
>       |
> ```
> 
> and a segfault from the malloc
> 
> ```
> malloc(): corrupted top size
> Aborted (core dumped)
> ```

Can't reproduce on Debian Sid with GCC (Debian) 13.1.0-9.  If you didn't accidentally change the program I wrote, I don't see any UB in it, so there might be a GCC bug.  Please send me your program to check, and GCC version.

```c
$ cat off.c 
#include <err.h>
#include <stddef.h>
#include <stdalign.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>


struct s {
	int   i;
	char  c;
	//char  ___pad[3];
	char  fam[];
};


static inline void *xmalloc(size_t size);


int
main(void)
{
	char      *p;
	struct s  *s;

	printf("sizeof(s): %zu\n", sizeof(struct s));
	printf("offsetof(s, fam): %zu\n", offsetof(struct s, fam));
	printf("alignof(s): %zu\n", alignof(struct s));
	printf("alignof(fam): %zu\n", alignof(s->fam));

	puts("\nWith sizeof():");

	s = xmalloc(sizeof(struct s) + sizeof("Hello, sizeof!"));
	strcpy(s->fam, "Hello, sizeof!");
	p = (char *) s + sizeof(struct s);
	puts(p);
	free(s);

	puts("\nWith offsetof(3):");

	s = xmalloc(offsetof(struct s, fam) + sizeof("Hello, offsetof!"));
	strcpy(s->fam, "Hello, offsetof!");
	p = (char *) s + offsetof(struct s, fam);
	puts(p);
	free(s);

	puts("\nWith addr+offset:");

	s = xmalloc(sizeof(struct s) + sizeof("Hello, addr+offset!"));
	strcpy((char *)s + sizeof(struct s), "Hello, addr+offset!");
	p = (char *)s + sizeof(struct s);
	puts(p);
	free(s);

	exit(EXIT_SUCCESS);
}


static inline void *
xmalloc(size_t size)
{
	void  *p;

	p = malloc(size);
	if (p == NULL)
		err(EXIT_FAILURE, "malloc");
	return p;
}
```

```sh
$ clang -Weverything -g -O3 -D_FORTIFY_SOURCE=3 -o off off.c -fsanitize=address -fsanitize=pointer-compare -Wno-c++98-compat -Wno-gnu-alignof-expression
off.c:9:8: warning: padding size of 'struct s' with 3 bytes to alignment boundary [-Wpadded]
struct s {
       ^
1 warning generated.
```

```sh
$ gcc-13 -Wall -Wextra -g -O3 -D_FORTIFY_SOURCE=3 -fanalyzer -o off off.c -fsanitize=address -fsanitize=pointer-compare
```

```sh
$ ./off 
sizeof(s): 8
offsetof(s, fam): 5
alignof(s): 4
alignof(fam): 1

With sizeof():
lo, sizeof!

With offsetof(3):
Hello, offsetof!

With addr+offset:
Hello, addr+offset!
```

You may notice that I wrote your new section but with the appropriate size for malloc(3).  It just works.


> 
> And valgrind is generally very unhappy with that above strcpy.

Can't reproduce either.  Debian Sid.  Valgrind 3.19.0.

```sh
$ gcc-13 -Wall -Wextra -g -O3 -D_FORTIFY_SOURCE=3 -fanalyzer -o off off.c
$ valgrind ./off 
==10807== Memcheck, a memory error detector
==10807== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==10807== Using Valgrind-3.19.0 and LibVEX; rerun with -h for copyright info
==10807== Command: ./off
==10807== 
sizeof(s): 8
offsetof(s, fam): 5
alignof(s): 4
alignof(fam): 1

With sizeof():
lo, sizeof!

With offsetof(3):
Hello, offsetof!

With addr+offset:
Hello, addr+offset!
==10807== 
==10807== HEAP SUMMARY:
==10807==     in use at exit: 0 bytes in 0 blocks
==10807==   total heap usage: 4 allocs, 4 frees, 1,097 bytes allocated
==10807== 
==10807== All heap blocks were freed -- no leaks are possible
==10807== 
==10807== For lists of detected and suppressed errors, rerun with: -s
==10807== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

And I see no UB where valgrind should complain either.  The sizes are chosen so that strcpy(3) is fine.  `_FORTIFY_SOURCE` didn't complain either.  I'm running glibc (Debian) 2.37-6.  The kernel is 6.3.0-1-amd64, in case that's relevant.

> 
> Also with your code example, I added a new section that more closely mimics what I'm actually doing
> 
> ```c
>         s = xmalloc(1024);      /* I have some large chunk of memory */
>         strcpy((char *)s + sizeof(struct s), "Hello, addr+offset!");
>         p = (char *)s + sizeof(struct s);
>         puts(p);
>         free(s);
> ```

Yes, this doesn't show problems.  The reason is that you're ignoring the first entries of the FAM.  Let's write some ASCIIart.

```
/* _:  nothing; just to separate nibbles visually
 * p:  padding
 * c:  `char c`
 * i:  `int i`
 * f:  `char fam[]`
 */
```

With sizeof:

```
iiii_cfff_ffff_ffff_ffff
      Hel lo,  size of!0
puts:     ^~~~
```

With offsetof:

```
iiii_cfff_ffff_ffff_ffff_ff
      Hel lo,  offs etof !0
puts: ^~~~
```

With addr+offset:

```
iiii_cfff_ffff_ffff_ffff_ffff_ffff
          Hell o, a ddr+ offs et!0
puts:     ^~~~
```

With explicit padding, or overalignment, you can fix the sizeof issues:

With sizeof:

```
iiii_cppp_ffff_ffff_ffff_fff
          Hell o, s izeo f!0
puts:     ^~~~
```

With offsetof:

```
iiii_cppp_ffff_ffff_ffff_ffff_f
          Hell o, o ffse tof! 0
puts:     ^~~~
```

With addr+offset:

```
iiii_cppp_ffff_ffff_ffff_ffff_ffff
          Hell o, a ddr+ offs et!0
puts:     ^~~~
```

You waste the bytes, but at least, sizeof() agrees with the position of the FAM.

> 
> I'm also unsure if the offsetoff() makes things a little less clear

It **is** the pedantically correct way of doing the calculation.  All other ways are wrong; just more or less wrong depending on the context, and the consequences may be none if the padding happens to be friends with you, benign if you just overallocate but don't misuse it, or pretty bad if you misuse it.

By patching the kernel to implement `struct_size()` in terms of `offsetof(3)`, I found a bug.  It was quite unrelated, but thanks to the restrictive conditions of offsetof(3) --sizeof() pretty much accepts anything--, it was caught.

<https://lore.kernel.org/lkml/202308161913.91369D4A@keescook/T/#mb5d5055f84b11bc8c113d39dc0c7fd77a718b096>

> , particularly in the second case where I'm just wanting to start writing where the fam starts, but I don't access it as ->fam.

But sizeof() doesn't tell you that.  sizeof() tells you "where the structure trailing padding ends", which may or may not coincide with "where the FAM starts".  The only reliable way to know where the FAM starts is with either `s->fam`, or `offsetof(s, fam)` (which internally uses `s->fam`).

ac000:
> But sizeof() doesn't tell you that.  sizeof() tells you "where the structure trailing padding ends", which may or may not coincide with "where the FAM starts".  The only 

I'm going to say this one last time.

I know exactly how the structure in question is laid out, for very
specific reasons, thus I know _there_ _is_ _no_ _padding_...

If we want to make our own struct_size(), sure go ahead and if the
kernel folks are happy to use offsetof() in it, then make it use that.

alejandro-colomar:
> > But sizeof() doesn't tell you that. sizeof() tells you "where the structure trailing padding ends", which may or may not coincide with "where the FAM starts". The only


> I'm going to say this one last time.  I know exactly how the structure in question is laid out, for very specific reasons, thus I know _there_ _is_ _no_ _padding_...

Sorry, I should be more precise with my words.  It tells you the right number in the Unit case.  But conceptually it tells a different thing.

> If we want to make our own struct_size(),

I think we'll agree on that.  (I guess I didn't propose it because adding generic APIs here is buhuu.)

> sure go ahead and if the kernel folks are happy to use offsetof() in it, then make it use that.

I'll do that in two patches.  In the first one, in which we'll agree, I'll move this calculation to nxt_struct_size().  In the second, I'll replace sizeof() with offsetof().

PR #925, opened 2023-08-21
Fix memory clobbering in libunit

alejandro-colomar:


hongzhidao:
Hi,
I can't approve this patch based on the below thoughts.
1. It will introduce two ways of handling FAM. And it's not good to introduce the new one and replace the existing ones in the future. If we find the replacing is not good when replacing them, we'll have to handle the new macro again. Anyway there should be only one style all the time.
2. About the changing of `nxt_length`, please refer to Max D's suggestion.
```
Either way, suggested change looks bad to me. It obfuscates
what's going on in the macro, making it harder to understand it
and to use it correctly. While it might save some newbie
developers from making silly mistakes, it will also make the code
harder to understand for a lot more developers.

Note well that using ternary operators in macro arguments is a bad
coding practice regardless of whether it works with the particular
macro or not, and should be avoided.
```

alejandro-colomar:
> Hi, I can't approve this patch based on the below thoughts.
> 
>     1. It will introduce two ways of handling FAM. And it's not good to introduce the new one and replace the existing ones in the future. If we find the replacing is not good when replacing them, we'll have to handle the new macro again. Anyway there should be only one way all the time.

I've added patches that replace all the uses, as you suggest.  Replacing them is fine.

> 
>     2. About the changing of `nxt_length`, please refer to Max D's suggestion.
> 
> 
> ```
> Either way, suggested change looks bad to me. It obfuscates
> what's going on in the macro, making it harder to understand it
> and to use it correctly. While it might save some newbie
> developers from making silly mistakes, it will also make the code
> harder to understand for a lot more developers.
> 
> Note well that using ternary operators in macro arguments is a bad
> coding practice regardless of whether it works with the particular
> macro or not, and should be avoided.
> ```

While I think Maxim is wrong there, in that experienced programmers do the same mistakes, let's try to improve the readability of the macro while keeping it safe.  I've rewritten it in terms of sizeof().  I think it's more readable now.

hongzhidao:
> replace all the uses, as you suggest. 

I didn't suggest replacing the uses, I just say there should be only one way to handle FAM in Unit.

Take this as an example.
```
@@ -1625,8 +1623,8 @@ nxt_conf_json_parse_object(nxt_mp_t *mp, nxt_conf_value_t *value, u_char *start,
        }
    }

    object = nxt_mp_get(mp, sizeof(nxt_conf_object_t)
                            + count * sizeof(nxt_conf_object_member_t));
    size = nxt_sizeof_struct(nxt_conf_object_t, members, count);
    object = nxt_mp_get(mp, size);
    if (nxt_slow_path(object == NULL)) {
        goto error;
    }
```
The current way is clear and any C developer can understand it without learning new macros like `nxt_sizeof_struct`.
Refactoring should make it better than before, otherwise, it's not ok to do it.

Issue #936, opened 2023-09-03
Building with Clang 16 fails with sign-compare


andypost:
Trying to build for Alpinelinux using Clang `--cc=clang` but getting error and the only workaround is to use `--cc-opt="$CFLAGS -Eno-sign-compare"`

```
checking for C compiler: clang
 + using Clang C compiler
 + Alpine clang version 16.0.6
building an "echo" program
checking for endianness ... little endian
checking for int size ... 4
checking for long size ... 8
checking for long long size ... 8
checking for void * size ... 8
checking for size_t size ... 8
checking for off_t size ... 8
checking for time_t size ... 8
....
In file included from src/nxt_socket_msg.c:7:
src/nxt_socket_msg.h:138:17: error: comparison of integers of different signs: 'unsigned long' and 'long' [-Werror,-Wsign-compare]
         cmsg = CMSG_NXTHDR(&msg, cmsg))
                ^~~~~~~~~~~~~~~~~~~~~~~
/usr/include/sys/socket.h:358:44: note: expanded from macro 'CMSG_NXTHDR'
        __CMSG_LEN(cmsg) + sizeof(struct cmsghdr) >= __MHDR_END(mhdr) - (unsigned char *)(cmsg) \
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from src/nxt_socket_msg.c:7:
src/nxt_socket_msg.h:177:17: error: comparison of integers of different signs: 'unsigned long' and 'long' [-Werror,-Wsign-compare]
         cmsg = CMSG_NXTHDR(&msg, cmsg))
                ^~~~~~~~~~~~~~~~~~~~~~~
/usr/include/sys/socket.h:358:44: note: expanded from macro 'CMSG_NXTHDR'
        __CMSG_LEN(cmsg) + sizeof(struct cmsghdr) >= __MHDR_END(mhdr) - (unsigned char *)(cmsg) \
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated.
make: *** [build/Makefile:276: build/src/nxt_socket_msg.o] Error 1
make: *** Waiting for unfinished jobs....
[...]

PR #941, opened 2023-09-06
Fix build on musl libc with clang.

ac000:
As reported by @andypost on GitHub, if you try to build Unit on a system that uses musl libc (such as Alpine Linux) with clang then you get the following

clang -c -pipe -fPIC -fvisibility=hidden -O -W -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -fstrict-aliasing -Wstrict-overflow=5 -Wmissing-prototypes -Werror -g   -I src -I build/include   \
                      \
                     \
-o build/src/nxt_socketpair.o \
-MMD -MF build/src/nxt_socketpair.dep -MT build/src/nxt_socketpair.o \
src/nxt_socketpair.c
In file included from src/nxt_socketpair.c:8:
src/nxt_socket_msg.h:138:17: error: comparison of integers of different signs: 'unsigned long' and 'long' [-Werror,-Wsign-compare]
         cmsg = CMSG_NXTHDR(&msg, cmsg))
                ^~~~~~~~~~~~~~~~~~~~~~~
/usr/include/sys/socket.h:358:44: note: expanded from macro 'CMSG_NXTHDR'
        __CMSG_LEN(cmsg) + sizeof(struct cmsghdr) >= __MHDR_END(mhdr) - (unsigned char *)(cmsg) \
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from src/nxt_socketpair.c:8:
src/nxt_socket_msg.h:177:17: error: comparison of integers of different signs: 'unsigned long' and 'long' [-Werror,-Wsign-compare]
         cmsg = CMSG_NXTHDR(&msg, cmsg))
                ^~~~~~~~~~~~~~~~~~~~~~~
/usr/include/sys/socket.h:358:44: note: expanded from macro 'CMSG_NXTHDR'
        __CMSG_LEN(cmsg) + sizeof(struct cmsghdr) >= __MHDR_END(mhdr) - (unsigned char *)(cmsg) \
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated.
make: *** [build/Makefile:261: build/src/nxt_socketpair.o] Error 1

GCC works fine, it seems to have some smarts so that it doesn't give warnings on system header files.

This seems to be a long standing issue with musl libc (bad casting in the CMSG_NXTHDR macro) and the workaround employed by several projects is to disable the -Wsign-compare clang warning for the code in question.

So, that's what we do. We wrap the CMSG_NXTHDR macro in a function, so we can use the pre-processor in it to selectively disable the warning.

### DIFF
--- a/src/nxt_socket_msg.h
+++ b/src/nxt_socket_msg.h
@@ -69,6 +69,20 @@ NXT_EXPORT ssize_t nxt_recvmsg(nxt_socket_t s,
     nxt_iobuf_t *iob, nxt_uint_t niob, nxt_recv_oob_t *oob);
 
 
+nxt_inline struct cmsghdr *
+NXT_CMSG_NXTHDR(struct msghdr *msgh, struct cmsghdr *cmsg)
+{
+#if !defined(__GLIBC__) && defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wsign-compare"
+#endif
+    return CMSG_NXTHDR(msgh, cmsg);
+#if !defined(__GLIBC__) && defined(__clang__)
+#pragma clang diagnostic pop
+#endif
+}
+
+
 nxt_inline void
 nxt_socket_msg_oob_init(nxt_send_oob_t *oob, int *fds)
 {
@@ -135,7 +149,7 @@ nxt_socket_msg_oob_get_fds(nxt_recv_oob_t *oob, nxt_fd_t *fd)
 
     for (cmsg = CMSG_FIRSTHDR(&msg);
          cmsg != NULL;
-         cmsg = CMSG_NXTHDR(&msg, cmsg))
+         cmsg = NXT_CMSG_NXTHDR(&msg, cmsg))
     {
         size = cmsg->cmsg_len - CMSG_LEN(0);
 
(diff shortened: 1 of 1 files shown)


ac000:
Added missing 'Closes' tag.

andypost:
Thank you, posted review to https://github.com/nginx/unit/issues/936#issuecomment-1708938752 that fixed build on 4 arches

but it still fail on arm(7/hf/64) and riscv64 because of atomics - there's pipelines for clang 15 and 16 https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/51157/pipelines

ac000:
> Thank you, posted review to [#936 (comment)](https://github.com/nginx/unit/issues/936#issuecomment-1708938752) that fixed build on 4 arches
> 
> but it still fail on arm(7/hf/64) and riscv64 because of atomics - there's pipelines for clang 15 and 16 https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/51157/pipelines

To avoid any confusion. This is _not_ related to this patch and is being tracked in a separate [issue](https://github.com/nginx/unit/issues/942).

hongzhidao:
LGTM.
The only question is if we need to add a BB test or not. @thresheek @ac000 @lcrilly.

ac000:
Do we want a BB with say, Alpine Linux + clang?

Apart from the extra resources and possible extra time to complete a run, I don't think having more clang coverage would be bad thing in general. AFAICT the only builders that currently use clang are the FreeBSD ones and the code analyser.

hongzhidao:
> Do we want a BB with say, Alpine Linux + clang?

Yep, that's what I mean.

PR #963, opened 2023-09-30
Fixes an error in memory copying when increasing the size of nxt_vector_t

nikolay-govorov:
Thanks for the work you are doing. Unit is a really great tool.

I found that when system increase the size if `nxt_ vector_t`, the function copy more bytes into the new buffer from the old buffer than there are (the new size, not the old). This usually doesn't result in an apparent error since there is something in the allocator buffer, but occasionally it can end in a segfault.

The address-sanitizer will also not tell you this because of its own allocator. You could potentially [markup your allocator](https://github.com/google/sanitizers/wiki/AddressSanitizerManualPoisoning) to see such errors in tests.

### DIFF
--- a/src/nxt_vector.c
+++ b/src/nxt_vector.c
@@ -75,12 +75,13 @@ void *
 nxt_vector_add(nxt_vector_t *vector, const nxt_mem_proto_t *proto, void *pool)
 {
     void      *item, *start, *old;
-    size_t    size;
+    size_t    old_size, size;
     uint32_t  n;
 
     n = vector->avalaible;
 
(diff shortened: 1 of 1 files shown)


hongzhidao:
Hi,
Thanks for reporting it, but this file is not used.
You could take a look at `nxt_array.h/c` instead which does the same thing.
We don't maintain these unused files and we might consider deleting them in the future.

nikolay-govorov:
Oh yes indeed, `nxt_array` works great and did not inherit this problem. Thank you.

Issue #967, opened 2023-10-04
[Homebrew/Golang] Default paths for Apple Silicon


lcrilly:
Currently, the Homebrew experience for Unit with Go applications is set up for x86 and fails on arm64 machines. Typically the problems experiences are:

```
$ go build …
…
# unit.nginx.org/go
In file included from ../../../go/pkg/mod/unit.nginx.org/go@v0.0.0-20231004115303-cfd5bf6e4324/port.go:9:
./nxt_cgo_lib.h:14:10: fatal error: 'nxt_unit.h' file not found
#include <nxt_unit.h>
         ^~~~~~~~~~~~
1 error generated.
```

Or
```
# command-line-arguments
/opt/homebrew/Cellar/go/1.21.1/libexec/pkg/tool/darwin_arm64/link: running cc failed: exit status 1
ld: library 'unit' not found
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

The solution is to provide CGO_ environment variables
```
$ CGO_LDFLAGS="-L/opt/homebrew/include -L/opt/homebrew/lib" CGO_CFLAGS="-I/opt/homebrew/include" go build …
```

Let's fix this for a better user experience! :)

PR #1017, opened 2023-12-05
Go: Use Homebrew include paths

danielledeleo:
This patch adds include paths for `libunit` installed via Homebrew only when built on Darwin (macOS). I don't know if Linux Homebrew uses the same paths, so this shouldn't get in the way in that edge case.

Small note, the syntax used on line 1 is the preferred syntax as of go1.18 and the old style of `// +build` should be `go fix`d across the repository in the future.

Fixes nginx/unit#967

### DIFF
--- a/go/ldflags-darwin.go
+++ b/go/ldflags-darwin.go
@@ -1,0 +1,14 @@
+//go:build darwin
+
+/*
+ * Copyright (C) Danielle De Leo
+ * Copyright (C) NGINX, Inc.
+ */
+
+package unit
+
+/*
+#cgo LDFLAGS: -L/opt/homebrew/lib
+#cgo CFLAGS: -I/opt/homebrew/include
+*/
+import "C"


ac000:
Don't forget to update the issue link in the commit message... (you can force push here)

go/ldflags-darwin.go
```diff
@@ -0,0 +1,14 @@
...
+
+/*
+ * Copyright (C) Danielle De Leo
+ * Copyright (C) NGINX, Inc.
+ */
+
+package unit
+
+/*
+#cgo LDFLAGS: -L/opt/homebrew/include -L/opt/homebrew/lib
```
ac000, line 11: Hmm, is `-L/opt/homebrew/include` really needed here? I hope there's no libraries in there!

danielledeleo: Good point :) Thanks Andrew!

thresheek:
Would it be possible for a user to redefine those flags to make it possible to build with custom Unit installs even if you have Unit via Homebrew installed?

danielledeleo:
> Would it be possible for a user to redefine those flags to make it possible to build with custom Unit installs even if you have Unit via Homebrew installed?

Yes. I can confirm that setting the `CGO_` environment variables overrides those set in the .go file. With Unit built from source and installed with `make libunit-install` these flags work. It is pretty opaque though.

```
CGO_LDFLAGS="-L/usr/local/lib" CGO_CFLAGS="-I/usr/local/include" go build
```

ac000 (approved):
LGTM

danielledeleo:
Actually... this isn't ideal. If artifacts from a source build are present in the default includepath, the Homebrew libraries will not be used unless specified as per #967. 

So, if no source builds are present in the default includepath, a Homebrew install will work with a simple `go build`, otherwise it will have to be overridden. 

If this behaviour is acceptable, I'm okay to merge this.

ac000:
> Actually... this isn't ideal. If artifacts from a source build are present in the default includepath, the Homebrew libraries will not be used unless specified as per #967.

Assuming I'm understanding correctly (and it's entirely possible I'm not), might that not be desirable? If you want to try with a newer version of Unit built from source...

danielledeleo:
> might that not be desirable? If you want to try with a newer version of Unit built from source...

Yeah that would be the behaviour!

ac000:
On Tue, 05 Dec 2023 10:57:31 -0800
Dani De Leo ***@***.***> wrote:

> > might that not be desirable? If you want to try with a newer version of Unit built from source...  
> 
> Yeah that would be the behaviour! 

Sounds reasonable to me...

PR #1020, opened 2023-12-07
Add additional replace rules for node:* modules to custom unit node loaders.

### LINKED ISSUE #1013: Can Unit run NodeJS Nuxt applications?
We have infrastructure with many PHP applications. A part of these applications we're running in Nginx Unit.
We have external developers who develop Nuxt application.
We want to unify our infrastructure and use Unit everywhere.
But when we trying to build Nuxt application and run it in Unit (as documented [here, in Unit documentation](https://unit.nginx.org/configuration/#configuration-nodejs-loader)), `curl` stucks on Unit port, but on Node port, 3000, `curl` works normally.

javorszky:
Closes: <https://github.com/nginx/unit/issues/1013>

In that particular issue the compiled nuxt files end up importing the http module as `node:http` rather than `http` only. This bypasses unit's custom loader implementation which only check for the `http` or `unit-http` modules, and their websocket counterparts.

This changeset adds replace sources for both the `node:http` and `node:websocket` import signatures.

### DIFF
--- a/src/nodejs/unit-http/loader.js
+++ b/src/nodejs/unit-http/loader.js
@@ -11,10 +11,12 @@ if (module.parent && module.parent.id === "internal/preload") {
         Module.prototype.require = function (id) {
             switch(id) {
                 case "http":
+                case "node:http":
                 case "unit-http":
                     return http
 
                 case "websocket":
+                case "node:websocket":
                 case "unit-http/websocket":
                     return websocket
             }
--- a/src/nodejs/unit-http/loader.mjs
+++ b/src/nodejs/unit-http/loader.mjs
@@ -2,6 +2,7 @@
 export async function resolve(specifier, context, defaultResolver) {
     switch (specifier) {
         case "websocket":
+        case "node:websocket":
             return {
                 url: new URL("./websocket.js", import.meta.url).href,
                 format: "commonjs",
@@ -9,6 +10,7 @@ export async function resolve(specifier, context, defaultResolver) {
             }
 
         case "http":
+        case "node:http":
             return {
                 url: new URL("./http.js", import.meta.url).href,
                 format: "commonjs",


javorszky:
see https://nodejs.org/api/https.html

and https://nodejs.org/api/http.html

javorszky:
@ac000 I've updated the commit message with description from the PR, and changed the link on both the PR and the commit message to include the full URI of the github issue.

This is now ready to review / merge.

ac000 (changes requested):
While the changes themselves look simple enough. I do think it's a good idea to give recognition to people who report issues and test fixes etc (it also gives some further points of contact down the line).

So I would add a Reported-by tag at the bottom, at which point lets move the closes there also, and then you may as well go the full hog and add your Signed-off-by also. Ultimately I'd like for us to officially adopt the [DCO](https://developercertificate.org/) **(not a CLA)**.

So the bottom of the commit message (separated by a blank line) you'd end up with

```
Reported-by:  rommanio <https://github.com/rommanio>
Closes: <https://github.com/nginx/unit/issues/1013>
Signed-off-by: Your Name <email address>
```

I can't force you to do that (yet), but if you have a look through git-log(1) you'll see plenty of examples of this usage.

ac000:
GH will recognise full URL's, but doesn't recognise it being in <>'s.

ac000:
To quote https://man7.org/linux/man-pages/man7/uri.7.html

> When written, URIs should be placed inside double quotes (e.g.,
       "http://www.kernel.org"), enclosed in angle brackets (e.g.,
       \<http://lwn.net/>), or placed on a line by themselves.  A warning
       for those who use double-quotes: never move extraneous
       punctuation (such as the period ending a sentence or the comma in
       a list) inside a URI, since this will change the value of the
       URI.  Instead, use angle brackets instead, or switch to a quoting
       system that never includes extraneous characters inside quotation
       marks.  This latter system, called the 'new' or 'logical' quoting
       system by "Hart's Rules" and the "Oxford Dictionary for Writers
       and Editors", is preferred practice in Great Britain and in
       various European languages.  Older documents suggested inserting
       the prefix "URL:" just before the URI, but this form has never
       caught on.

ac000:
GH should probably learn to understand that format... lets see where to file such a request...

javorszky:
I don't think GH is going to entertain a request for this. Out of curiosity, is the angle bracket format used somewhere downstream that gets picked up automatically by some tool?

ac000:
No.

I used to write them without the brackets until Alex pointed me to the above and suggested we should.

I can live without them if it's a problem.

callahad:
Yeah, GitHub doesn't recognize them for auto-closing. Full URLs,`org/repo#issuenum`, `repo#issuenum`, and `#issuenum` all work. I'm fine with any of those formats. If we want to specify a preferred format, we should move that to a GitHub Discussion.

ac000:
Full URLs (without the brackets) so they are not tied to and are usable outside of GH.

Obviously email addresses should still be brackets and and also where we refer to GH users where we don't have an email address for them but use their GH profile link.

tippexs:
Any update on this PR? If we can complete the review I will add this to 1.32

andrey-zelenkov (approved):

alejandro-colomar:
I opened a bug-report ticket to github.  It seems those a private, so I can't link here.  :|

javorszky:
> Any update on this PR? If we can complete the review I will add this to 1.32 or do we need to push it further out to resolve the `https` issue posted in #1025

From my point this PR is done. I'm unsure what else is needed to get this merged.

ac000:
It's just the `Closes` tag that needs some adjustment...

And to avoid any ambiguity, it should look like this (for now, if and when GH fix the issue with the <>'s we can start putting them in again)

[You can also loose the trailing '.' from the subject line... so]

```
Add additional replace rules for node:* modules

In that particular issue the compiled nuxt files end up importing the
http module as node:http rather than http only. This bypasses unit's
custom loader implementation which only check for the http or unit-http
modules, and their websocket counterparts.

This changeset adds replace sources for both the node:http and
node:websocket import signatures.

Closes: https://github.com/nginx/unit/issues/1013
```

javorszky:
@ac000 rebased on latest master and commit message updated

callahad:
LGTM. Minimal change, already separately reviewed. Changes requested by @ac000 have been applied. Bypassing branch protections and merging.

PR #1022, opened 2023-12-11
HTTP: added TSTR validation flag to the rewrite option.

hongzhidao:
Take the configuration as an example:
```
{
    "rewrite": "`${foo + "
}
```
Before it was compiled in the router process with a simple error message:
```
failed to apply previous configuration
```
It should be checked in the controller process with the error message:
```
the previous configuration is invalid: "SyntaxError: Unexpected end of input in default:1" in the "rewrite" value.
```

### DIFF
--- a/src/nxt_conf_validation.c
+++ b/src/nxt_conf_validation.c
@@ -690,6 +690,7 @@ static nxt_conf_vldt_object_t  nxt_conf_vldt_action_common_members[] = {
     {
         .name       = nxt_string("rewrite"),
         .type       = NXT_CONF_VLDT_STRING,
+        .flags      = NXT_CONF_VLDT_TSTR,
     },
     {
         .name       = nxt_string("response_headers"),


callahad:
Yes, it's an unclosed template string (backtick delimited)

hongzhidao:
> The error here is that the rewrite rule is invalid something?

Yes, the option value is invalid.

> What's the value of something? Is it NJS?

It's a JS code that starts with `. Then it's an invalid JS code. We should report the compiled error.
```
# JS code
${foo + 
```
With error
```
SyntaxError: Unexpected end of input in default:1
```
Btw, we should do it in the controller process, since the related option name needs to be listed.
```
in the "rewrite" value.
```

Based on the above, the message below is necessary for users to read.
```
the previous configuration is invalid: "SyntaxError: Unexpected end of input in default:1" in the "rewrite" value.
```

> This information (once complete) should also go in the commit message.

Makes sense.

ac000:
So this new error _only_ applies to invalid JS?.

hongzhidao:
Correct, internally we call it template string (tstr), that’s why we add the flag here.

hongzhidao:
Sure, what about the other one?
Or I'll rewrite it like this. Is that ok?
```
This is to improve error messages for rewrite configuration.
Take the configuration as an example:
{
    "rewrite": "`${foo + "
}

Previously, when applying it the user would see this error message:

failed to apply previous configuration

After this change, the user will see this improved error message:

the previous configuration is invalid: "SyntaxError: Unexpected end of input in default:1" in the "rewrite" value.
```

ac000:
Oh, yes, that's good, just put a blank line before the config snippet.

(Personally I would also indent the config snippet by a couple of chars, but I'll leave that up to you).

ac000:
Heh, OK, miscommunication. This is really what I meant (in fact lets indent the error messages also...)

```
This is to improve error messages for rewrite configuration.
Take the configuration as an example:

  {
      "rewrite": "`${foo + "
  }

Previously, when applying it the user would see this error message:

  failed to apply previous configuration

After this change, the user will see this improved error message:

  the previous configuration is invalid: "SyntaxError: Unexpected end of input in default:1" in the "rewrite" value.
```

ac000:
The other issue I see is that it's using your @gmail address.

You can fix this by doing in your unit repository

```
$ git config user.name "Zhidao HONG"
$ git config user.email z.hong@f5.com
$ git commit --amend --no-edit --author="Zhidao HONG <z.hong@f5.com>"
```

That will set your .git/config user.{name,email} entries and update the HEAD commits 'Author' entry. The 'Commit' entry will update automatically.

You can check the result with

```
$ git show --pretty=fuller
```

ac000 (approved):
Thanks Zhidao, all looks good now (incl email address).

andrey-zelenkov (approved):

PR #1045, opened 2024-01-05
Fix typo in tools/README.md

callahad:
As suggested by @lcrilly

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -33,7 +33,7 @@ web page with NGINX Unit.
 
 | Options | |
 |---------|-|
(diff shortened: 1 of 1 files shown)


lcrilly:
Suggest the commit message is prefixed with `Tools:` and ends with a `.` (if we're still doing that)

ac000:
> and ends with a `.` (if we're still doing that)

We're not (well some folks might still be, but it goes against standard practice), Look at recent commits.

lcrilly (approved):
LGTM

PR #1050, opened 2024-01-08
Docker: Added Python 3.12

tclesius:
As suggested by @callahad.

### DIFF
--- a/pkg/docker/Dockerfile.python3.12
+++ b/pkg/docker/Dockerfile.python3.12
@@ -1,0 +1,89 @@
+FROM python:3.12-bullseye
+
+LABEL org.opencontainers.image.title="Unit (python3.12)"
+LABEL org.opencontainers.image.description="Official build of Unit for Docker."
+LABEL org.opencontainers.image.url="https://unit.nginx.org"
+LABEL org.opencontainers.image.source="https://github.com/nginx/unit"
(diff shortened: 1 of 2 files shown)


ac000:
I'll make a couple of observations

1) The versions seem to get bumped periodically.

It's not entirely clear what dictates that.

2) For some things we have multiple versions for others we don't...

Again, I'm not sure what dictates this...

thresheek:
The change is good.



> I'll make a couple of observations
> 
>     1. The versions seem to get bumped periodically.
> 
> 
> It's not entirely clear what dictates that.

I typically go through the list and do the bumps as a part of pre-release activity.

>     2. For some things we have multiple versions for others we don't...
> 
> 
> Again, I'm not sure what dictates this...

I've introduced support for multiple versions of one language a not to long time ago - so that probably explains it?

As a sidenote (and I don't think it's written anywhere), I think a good idea will be to keep the list synced to what Docker Library offers for a given release time, but probably not fully, e.g. two most recent versions of a given language?
I don't think it's beneficial to have e.g. 4 perl versions (5.38, 5.36, 5.34, 5.39).  It's something to discuss I guess.

callahad:
> It's something to discuss I guess.

Absolutely :) Let's keep this discussion scoped to Py3.12, and I'll add language versions to the queue of process discussions we need to have in the next few weeks.

thresheek (approved):

PR #1055, opened 2024-01-09
Go: Add missing +build and go:build comments

danielledeleo:
A RHEL 8 test was failing because it uses go1.16. The old style must be retained for backwards compat.

### DIFF
--- a/go/ldflags-lrt.go
+++ b/go/ldflags-lrt.go
@@ -1,3 +1,4 @@
+//go:build linux || netbsd
 // +build linux netbsd
 
 /*
--- a/go/ldflags-darwin.go
+++ b/go/ldflags-darwin.go
@@ -1,4 +1,5 @@
 //go:build darwin
+// +build darwin
 
 /*
  * Copyright (C) Danielle De Leo


arbourd (approved):

tarynmusgrave:
I know we are still working through these and looking through the PR history we haven't been sticking to this but the [Git Style Guide](https://github.com/nginx/unit/blob/master/CONTRIBUTING.md#git-style-guide) for this repo mentions using past tense language for example 'added' instead of 'add' for the subject line.

ac000:
Personally I dont use past tense as it generally reads strange to me (I did initially in this project, until it grated on my brain enough), I prefer to use the more common imperative form where you give orders to make changes.

So for me the above summary is exacly how I'd write it...

go/ldflags-lrt.go
```diff
@@ -1,3 +1,4 @@
+//go:build linux || netbsd
```
ac000, line 1: How about Free and Open BSD?

Just a side note, but on newer glibc (around 2.17 IIRC), explictly linking against rt is no longer required.

Has this been run on buildbot?

ac000:
If this issue was introduced by https://github.com/nginx/unit/commit/9a36de84c8534cdd8d22b33d90309519e03abf3b, then perhaps you could add a ```Fixes:``` tag (see 763396b8be07be41b1baf336952fd222cbeb8db7 as an example) at the footer of the commit message (separated by a blank line) such as

```
Fixes: 9a36de84c ("Go: Use Homebrew include paths")
```

You can also add my

```
```

(Feel free to add your ```Signed-off-by:``` also...)

ac000 (approved):

tarynmusgrave:
> Personally I dont use past tense as it generally reads strange to me (I did initially in this project, until it grated on my brain enough), I prefer to use the more common imperative form where you give orders to make changes.
> 
> So for me the above summary is exacly how I'd write it...

I don't have a strong opinion either way however it would be nice we were were consistent - in either direction. Should we be changing the Git Style Guide in that case? We don't want to make it more difficult for new contributors.

ac000:
> Should we be changing the Git Style Guide in that case?

I would say yes (no surprises there!). There are also some things which just don't translate into past tense...

On the other hand, we do get contributions from non-native English speakers so unless we want to be re-writing everything into perfect English prose we need to allow some leeway. As long as the information is there and it's understandable...

PR #1057, opened 2024-01-09
Updated security.txt

thresheek:
Refs: https://github.com/nginx/unit-docs/pull/78


danielledeleo (approved):

danielledeleo:
We'll fetch this file when deploying to Netlify instead of storing it twice.

Issue #1009, opened 2023-11-24
Docker: Unable to build images for PHP 8.3.0RC6 using packages Makefile


tippexs:
According to the [documentation ](https://unit.nginx.org/installation/#inst-lang-docker) it should be possible to build custom container images using official Docker base images using `make` in `pkg/docker`.

Issuing
```bash
make build-php8.3.0RC6 VERSIONS_php=8.3.0RC6 VARIANT_php=cli-bullseye CONFIGURE_php=php INSTALL_php=php-install RUN_php=ldconfig MODULE_PREBUILD_php=/bin/true
```

Will create a Dockerfile BUT it will not include the information to build the Unit language module. Here is a diff from the two Dockerfiles generated by `make`.

```bash
$> diff Dockerfile.php8.3.0RC6 Dockerfile.php8.2
1c1
< FROM php:8.3.0RC6-cli-bullseye
---
> FROM php:8.2-cli
3c3
< LABEL org.opencontainers.image.title="Unit (php8.3.0RC6)"
---
> LABEL org.opencontainers.image.title="Unit (php8.2)"
48c48
<     &&  \
---
>     && /bin/true \
50,51c50,51
<     && ./configure  \
<     && make -j $NCPU  \
---
>     && ./configure php \
>     && make -j $NCPU php-install \
54,55c54,55
<     && ./configure  \
<     && make -j $NCPU  \
---
>     && ./configure php \
>     && make -j $NCPU php-install \
63c63
<     &&  \
---
>     && ldconfig \
```

To test application code and Unit against early versions / Release-Candidates / Alpha-Versions of a language is super important and should NOT require hard extra work.

ac000:
It seems to break as soon as it hits anything that isn't [0-9.] in the version.

tippexs:
Yes with a high chance because of this https://github.com/nginx/unit/blob/3fdf8c63a2dd4b3d676754d2e605877a8689e4c9/pkg/docker/Makefile#L112

ac000:
On Fri, 24 Nov 2023 07:19:08 -0800
Timo Stark ***@***.***> wrote:

> I am flexible what exactly we should do here but not beeing able to building a Container Image that is based on a name different from `php:8.3` is a missing feature. Calling it a feature because the `Makefile` was never designed to work with such releases.

It seems we only need to find the right thing for the configure command
and it can only be one of a set number of items; php, python, perl
etc...

I don't see why we can't just take the contents of the
[...]

PR #1076, opened 2024-01-19
Docker: Add a new target for generic language versions

ac000:
You can use the Docker Makefile to create dockerfiles for specific language environments, e.g

  $ make build-python3.12 VERSIONS_python=3.12

However this breaks when trying to do something like

  $ make build-php8.3.0RC6 VERSIONS_php=8.3.0RC6 VARIANT_php=cli

This breaks due to the RC in the version, due to how we're trying to parse out the module name from the version string, we strip out anything that is a 0-9 or a . or -.

That works for 'python3.12' leaving 'python'. However for 'php8.3.0RC6' that leaves 'phpRC'.

After spending some time on this, the best (and simplest) solution I could come up with, that doesn't break existing usage, is to create a new make target for this.

So the above now becomes

  $ make gen-php8.3.0RC6 MODULE=php VERSIONS_php=8.3.0RC6 VARIANT_php=cli

The new target is gen-<language version> and you need to specify the language module.

### DIFF
--- a/pkg/docker/Makefile
+++ b/pkg/docker/Makefile
@@ -8,6 +8,8 @@ DEFAULT_VERSION := $(NXT_VERSION)
 VERSION ?= $(DEFAULT_VERSION)
 PATCHLEVEL ?= 1
 
+MODULE   =
+
 MODULES ?= go jsc node perl php python ruby wasm
 
 VARIANT ?= bullseye
@@ -105,7 +107,7 @@ export RUST_VERSION=1.71.0 \\\n \
 endef
 
 default:
-	@echo "valid targets: all build dockerfiles library clean"
+	@echo "valid targets: all build gen-<target> dockerfiles library clean"
 
 MODVERSIONS = $(foreach module, $(MODULES), $(foreach modversion, $(shell for v in $(VERSIONS_$(module)); do echo $$v; done | sort -r), $(module)$(modversion))) wasm minimal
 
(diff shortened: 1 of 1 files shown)


thresheek:
It does seem to work fine - however I'm reluctant to that change since it's a mess from users' POV: why would I choose to use `make gen-` targets instead of `make build-` ones?

ac000:
> It does seem to work fine - however I'm reluctant to that change since it's a mess from users' POV: why would I choose to use `make gen-` targets instead of `make build-` ones?

Because of the problem it addresses...

```
You can use the Docker Makefile to create dockerfiles for specific language environments, e.g

$ make build-python3.12 VERSIONS_python=3.12

However this breaks when trying to do something like

$ make build-php8.3.0RC6 VERSIONS_php=8.3.0RC6 VARIANT_php=cli

This breaks due to the RC in the version, due to how we're trying to parse out the module name from the version string, we strip out anything that is a 0-9 or a . or -.

That works for 'python3.12' leaving 'python'. However for 'php8.3.0RC6' that leaves 'phpRC'.

After spending some time on this, the best (and simplest) solution I could come up with, that doesn't break existing usage, is to create a new make target for this.

So the above now becomes

$ make gen-php8.3.0RC6 MODULE=php VERSIONS_php=8.3.0RC6 VARIANT_php=cli

The new target is gen- and you need to specify the language module.
```

thresheek:
Sure, but I mean it would be a mess to document and from user POV it's kinda weird to have two ways of doing the same thing.

(I don't have a good solution in hand though)

callahad:
Feels like we agree that the ergonomics are challenging, but we should come up with a way to address the issue that this is trying to fix.

@ac000 I'm inclined to close this PR pending us coming up with a "better" solution, since it's stalled for several months and the approach seems unlikely to get merged. Thoughts?

callahad:
Let's close this for now and take a fresh run at it the next time #1009 is really affecting someone. Empirically, we've left that issue open for nearly a year, and this PR for almost six months without anything catching fire, so I submit that we can kick this back into triage.

Issue #1043, opened 2024-01-03
nodejs unit-http misses `options` in createServer function


stefanhuber:
I tried to run [strapi ](https://strapi.io/) inside nginx unit. The startup of strapi doesn't work because of the following error:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "listener" argument must be of type function. Received an instance of Object
    at checkListener (node:events:270:3)
    at _addListener (node:events:550:3)
    at Server.addListener (node:events:609:10)
    at new Server (/usr/local/lib/node_modules/unit-http/http_server.js:426:14)
    at Object.createServer (/usr/local/lib/node_modules/unit-http/http.js:15:12)
    at Module.createHTTPServer (/var/www/node_modules/@strapi/strapi/dist/services/server/http-server.js:16:40)
    at Module.createServer (/var/www/node_modules/@strapi/strapi/dist/services/server/index.js:26:35)
    at new Strapi (/var/www/node_modules/@strapi/strapi/dist/Strapi.js:130:27)
    at initFn (/var/www/node_modules/@strapi/strapi/dist/Strapi.js:466:18)
    at Object.<anonymous> (/var/www/index.js:2:1) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

Strapi uses `options` with the `http.createServer` function as seen [here](https://github.com/strapi/strapi/blob/762abbf7f8a8cae3cda9b6affcea50be8abc3ce9/packages/core/strapi/src/services/server/http-server.ts#L25C31-L25C68) in the source code. The node signature of the `createServer` function looks like this: `http.createServer([options][, requestListener])`. Looking at the `http` module inside the `unit-http` package the `createServer` function is missing the [options argument](https://github.com/nginx/unit/blob/49aee6760a26b14c06180e4f25088cb18e2037a7/src/nodejs/unit-http/http.js#L14).

Can this be solved somehow? How could strapi be started otherwise within nginx unit?

Nginx Unit has the following config.json:
```json
{
    "listeners": {
        "*:1337": {
            "pass": "routes/main"
        }
    },
    "routes": {
        "main": [
            {
                "action": {
                    "pass": "applications/app"
                }
            }
        ]
    },
    "applications": {
        "app": {
            "type": "external",
            "working_directory": "/var/www/",
            "executable": "/usr/bin/env",
            "arguments": [
                "node",
                "--loader",
                "unit-http/loader.mjs",
                "--require",
                "unit-http/loader",
                "index.js"
            ]
[...]

PR #1091, opened 2024-01-25
fix: Take options as well as requestListener

javorszky:
Closes https://github.com/nginx/unit/issues/1043

Unit-http have not kept up with the signature of nodejs's http package development. Nodejs allows an optional `options` object to be passed to the `createServer` function, we didn't. This resulted in function signature errors when user code that did make use of the options arg tried to call unit's replaced function.

This change changes the signature to be more in line with how nodejs does it discarding it and printing a message to stderr so it shows up in unit logs.

### DIFF
--- a/docs/changes.xml
+++ b/docs/changes.xml
@@ -53,6 +53,12 @@ NGINX Unit updated to 1.32.0.
 </para>
 </change>
 
+<change type="bugfix">
+<para>
+http.createServer() now accepts "options" argument introduced in Node.js v9.6.0, v8.12.0.
+</para>
+</change>
+
 <change type="feature">
 <para>
 conditional access logging.
--- a/src/nodejs/unit-http/http.js
+++ b/src/nodejs/unit-http/http.js
@@ -11,8 +11,8 @@ const {
     ServerResponse,
 } = require('./http_server');
 
-function createServer (requestHandler) {
-    return new Server(requestHandler);
+function createServer (options, requestHandler) {
+    return new Server(options, requestHandler);
 }
 
 const http = require("http")
--- a/src/nodejs/unit-http/http_server.js
+++ b/src/nodejs/unit-http/http_server.js
@@ -5,6 +5,7 @@
 
 'use strict';
 
+const { stderr } = require('process');
 const EventEmitter = require('events');
 const http = require('http');
 const util = require('util');
@@ -413,7 +414,14 @@ ServerRequest.prototype._read = function _read(n) {
 };
 
 
-function Server(requestListener) {
+function Server(options, requestListener) {
+    if (typeof options === 'function') {
+        requestListener = options;
+        options = {};
+    } else {
+        stderr.write("http.Server constructor was called with unsupported options, using default settings\n");
+    }
+
     EventEmitter.call(this);
 
     this.unit = new unit_lib.Unit();
--- a/test/node/options/app.js
+++ b/test/node/options/app.js
@@ -1,0 +1,4 @@
+require('http').createServer({}, function (req, res) {
+    res.writeHead(200, {'Content-Length': 12, 'Content-Type': 'text/plain'})
+       .end('Hello World\n');
+}).listen(8080);
--- a/test/test_node_application.py
+++ b/test/test_node_application.py
@@ -21,6 +21,12 @@ def test_node_application_basic():
 
     assert_basic_application()
 
+def test_node_application_options(wait_for_record):
+    client.load('options')
+
+    assert_basic_application()
+    assert wait_for_record(r'constructor was called with unsupported') is not None
+
 
 def test_node_application_loader_unit_http():
     client.load('loader/unit_http')


javorszky:
I tested with the following nodejs application file at `/Users/g.javorszky/Projects/Unit/issue-1043/flob/index.js`:

```js
const http = require("http");

const host = 'localhost';
const port = 8000;

const requestListener = function (req, res) {
    res.writeHead(200);
    res.end("My first server!");
};

const options = {
    keepAlive: true
};

const server = http.createServer(options, requestListener);
server.listen(port, host, () => {
    console.log(`Server is running on http://${host}:${port}`);
});
```

and the following unit configuration file:

```json
{
    "listeners": {
        "*:5555": {
            "pass": "routes/main"
        }
    },
    "routes": {
        "main": [
            {
                "action": {
                    "pass": "applications/app"
                }
            }
        ]
    },
    "applications": {
        "app": {
            "type": "external",
            "working_directory": "/Users/g.javorszky/Projects/Unit/issue-1043/flob/",
            "executable": "/usr/bin/env",
            "arguments": [
                "/Users/g.javorszky/.nvm/versions/node/v20.9.0/bin/node",
                "--loader",
                "/Users/g.javorszky/.nvm/versions/node/v20.9.0/lib/node_modules/unit-http/loader.mjs",
                "--require",
                "/Users/g.javorszky/.nvm/versions/node/v20.9.0/lib/node_modules/unit-http/loader",
                "index.js"
            ]
        }
    }
}
```

I had to use absolute paths for the node executable and the two loader files.

src/nodejs/unit-http/http.js
```diff
@@ -5,13 +5,21 @@
...
     ServerResponse,
 } = require('./http_server');
 
-function createServer (requestHandler) {
+function createServer (options, requestHandler) {
+    if (typeof options === 'function') {
+        requestHandler = options;
+        options = {};
+    } else {
+        stderr.write('createServer was called with an options object. Unit ignores this for the time being.');
```
callahad, line 20: I wonder if we should move this error handling into the `Server` constructor, and always pass `options` to it, even if empty? That'd potentially catch more attempts to customize the server.

callahad: Here's the Server constructor in Node: https://github.com/nodejs/node/blob/v21.6.1/lib/_http_server.js#L520

src/nodejs/unit-http/http.js
```diff
@@ -14,7 +14,7 @@ const {
 
 function createServer (options, requestHandler) {
     if (typeof options === 'function') {
-        requestListener = options;
+        requestHandler = options;
```
ac000, line 17: This should be squashed into the first commit. No point having commits that fix your own commits... ;)

ac000 (changes requested):

src/nodejs/unit-http/http.js
ac000: That does make review harder. I have to track changes across multiple commits that shouldn't actually exist.

I'm reviewing these things as if that's what will actually be merged. So once I'm finished reviewing I know exactly what state it's being merged as.

That's certainly how it's worked in this project and is what I would consider as best practice.

I see three logical changes in the four commits, so I would expect three commits.

javorszky: I think what @ac000 asks is entirely reasonable. I'll fixup that typo fix commit into the original one, but I do tend to work with atomic commits where related things stay together.

javorszky: Typos squashed into the relevant commits

src/nodejs/unit-http/http.js
javorszky: Option handling moved into the Server constructor in [52550b1](https://github.com/nginx/unit/pull/1091/commits/52550b18897802b5820bec809902513195afe4af)

javorszky:
@andrey-zelenkov I added the docs to changes in https://github.com/nginx/unit/pull/1091/commits/cdf9bfe32d17925f56d50b82e19ff7a45b094b71 as a response to your comment on the issue here: https://github.com/nginx/unit/issues/1013#issuecomment-1912149102

Everyone, let me know if the wording is appropriate, or I should tweak it.

src/nodejs/unit-http/http.js
```diff
@@ -13,14 +13,7 @@ const {
...
-        options = {};
-    } else {
-        stderr.write('createServer was called with an options object. Unit ignores this for the time being.');
-    }
-
-    return new Server(requestHandler);
+    return new Server(options, requestHandler);
 }
 
 const http = require("http")
```
ac000: Sorry, but I'm finding these commits very confusing...

It seems like this should just mostly replace the first commit, except that you want the options as part of the return and setting stderr.

From my point of view it looks like

- Setting stderr should be in its own commit.
- [Take options as well as requestListener](https://github.com/nginx/unit/pull/1091/commits/920be1168ddaf073240009cbf596b3f7d9647b31) should simply add the options parameter, and shouldn't it also return it? as you do in the latter patch...
- [Move options handling to unit's Server constructor](https://github.com/nginx/unit/pull/1091/commits/52550b18897802b5820bec809902513195afe4af) should then just do the meat of what the first patch did...

docs/changes.xml
```diff
@@ -26,6 +26,15 @@ NGINX Unit updated to 1.32.0.
 </para>
 </change>
 
+<change type="bugfix">
+<para>
+Unit's Node.js replacement http package brought the signature of http.createServer
+function and http.Server constructors in line with modern Node.js to include an option
+object to allow modern applications to function. Unit discards the passed in option
+object and prints to log that it has done so.
+</para>
```
andrey-zelenkov, line 35: Changes are just a description for users who briefly scanning release notes to find something related to them. So short note with keywords should be enough. Consider following:

`http.createServer() now accepts "options" argument introduced in Node.js v9.6.0, v8.12.0.`

javorszky: This has now been changed to this version, thank you!

test/node/options/app.js
```diff
@@ -0,0 +1,4 @@
+require('http').createServer({keepAlive: true}, function (req, res) {
```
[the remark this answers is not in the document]
javorszky: Added the test case into the python files next to the others, and removed the actual option from the object. It's now just an empty object being passed.

src/nodejs/unit-http/http.js
javorszky: I squashed all of these into a single commit as it's a single unit of work that relates to that. I don't see the need to separate the individual parts of this into their own commits per the feedback. My thinking is:

- commit for the actual functionality
- commit for the test
- commit for the docs

src/nodejs/unit-http/http_server.js
```diff
@@ -413,7 +413,14 @@ ServerRequest.prototype._read = function _read(n) {
 };
 
 
-function Server(requestListener) {
+function Server(options, requestListener) {
+    if (typeof options === 'function') {
+        requestListener = options;
+        options = {};
+    } else {
+        stderr.write('createServer was called with an options object. Unit ignores this for the time being.');
```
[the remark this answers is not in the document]
javorszky: Hm, is unrecognized (also, what's the style guide on using flavours of English? unrecognized / unrecognised and flavor / flavour?) the right word? Or would unsupported be better? Should we signal the temporaryness of this? (I think yes)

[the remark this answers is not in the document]
callahad: Yes, "Unsupported" is better. For spelling, Unit appears to use American English in both code and prose.

I think the most important things to communicate are:

1. You passed options we don't support.
3. This is a warning, not a fatal error; we're just ignoring them.

I'd _love_ to have a place in our docs we could link to outlining what parameters we _do_ support, but we don't have that (yet!)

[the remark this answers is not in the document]
javorszky: Reworded, thank you

ac000:
For 

```
 b090121 Take options as well as requestListener
```

I would rather the ```Closes:```, with a **:**, was at the bottom the commit message, separated by a blank line, for consistency. But other than that.

```
```

ac000 (dismissed):

test/test_node_application.py
```diff
@@ -21,6 +21,11 @@ def test_node_application_basic():
 
     assert_basic_application()
 
+def test_node_application_options():
+    client.load('options')
+
+    assert_basic_application()
+
```
andrey-zelenkov, line 28: I would suggest to check (in couple tests) that we don't have new error message in `unit.log`. Currently it should fail:

```
diff --git a/test/test_node_application.py b/test/test_node_application.py
--- a/test/test_node_application.py
+++ b/test/test_node_application.py
@@ -10,33 +10,36 @@ prerequisites = {'modules': {'node': 'al
 client = ApplicationNode()
 
 
-def assert_basic_application():
+def assert_basic_application(search_in_file):
     resp = client.get()
     assert resp['headers']['Content-Type'] == 'text/plain', 'basic header'
     assert resp['body'] == 'Hello World\n', 'basic body'
 
+    assert search_in_file(r'constructor was called with unsupported') is None
 
-def test_node_application_basic():
+
+def test_node_application_basic(search_in_file):
     client.load('basic')
 
-    assert_basic_application()
+    assert_basic_application(search_in_file)
 
-def test_node_application_options():
+
+def test_node_application_options(search_in_file):
     client.load('options')
 
-    assert_basic_application()
+    assert_basic_application(search_in_file)
 
 
-def test_node_application_loader_unit_http():
+def test_node_application_loader_unit_http(search_in_file):
     client.load('loader/unit_http')
 
-    assert_basic_application()
+    assert_basic_application(search_in_file)
 
 
-def test_node_application_loader_transitive_dependency():
+def test_node_application_loader_transitive_dependency(search_in_file):
```

src/nodejs/unit-http/http_server.js
```diff
@@ -413,7 +414,14 @@ ServerRequest.prototype._read = function _read(n) {
 };
 
 
-function Server(requestListener) {
+function Server(options, requestListener) {
+    if (typeof options === 'function') {
+        requestListener = options;
+        options = {};
+    } else {
```
andrey-zelenkov, line 421: This logic needs a little adjustment since an error message will be logged even when the `options` are an object. I would suggest adding an additional check in tests to verify this (please see below).

As far as I understood, NodeJS is also happy with `null` here, so we can copy this behavior too (to reduce further rework here).

src/nodejs/unit-http/http_server.js
```diff
@@ -413,7 +414,14 @@ ServerRequest.prototype._read = function _read(n) {
...
+        stderr.write('http.Server constructor was called with unsupported options, using default settings');
```
andrey-zelenkov, line 422: New line needed at the end of the error message string. Now it looks like following:

```
2024/01/30 17:14:07.682 [debug] 34852#18000878 malloc(87): 137F05980
http.Server constructor was called with unsupported options, using default settings2024/01/30 17:14:07.796 [debug] 0#18000881 [unit] Unit::Unit()
2024/01/30 17:14:07.797 [debug] 34852#18000881 [unit]
```

andrey-zelenkov (changes requested):

src/nodejs/unit-http/http_server.js
javorszky: This check is a slight adjustment of what the nodejs source does: https://github.com/nodejs/node/pull/15752/files#diff-2464899ab38c61dd42654fdb9658c75989cb6883b9c9d76613872a12d3fe2780R273-R278

The idea there is to figure out whether people are calling this function:

```js
function Server(options, requestListener) {...}
```
with one, or two arguments:

```js
Server(listener); // one argument

Server(options, listener); // two arguments
```

The only thing we care about right now is that if the first argument is a function, then we assume the function was called without the options argument, and the first incoming arg as `option` is actually intended to be a request listener.

If the users called this function with two arguments (ie the first argument is not a function), then it doesn't matter what the options is (undefined, null, empty object, actual object...), we should log that we are aware there's something, we're just not doing anything about it at the moment.

I don't think this needs changing.

andrey-zelenkov: > If the users called this function with two arguments (ie the first argument is not a function), then it doesn't matter what the options is (undefined, null, empty object, actual object...)

yeah, in this case no changes in logic required. I just assumed that sinse `{}` and `null` are valid values we could not report error in this case. Please ignore this one.

src/nodejs/unit-http/http_server.js
javorszky: Added a newline

javorszky:
@andrey-zelenkov I added the test, and in that I'm checking that the log DOES have the log line added, and the test passes for me locally. I have not changed any of the other tests.

andrey-zelenkov (approved):

andrey-zelenkov:
Currently this patch breaks our nodejs module on RHEL8:

```
% uname -a
Linux ip-10-1-18-198.eu-central-1.compute.internal 4.18.0-477.27.1.el8_8.x86_64 #1 SMP Thu Aug 31 10:29:22 EDT 2023 x86_64 x86_64 x86_64 GNU/Linux
```

```
% cat /etc/os-release
NAME="Red Hat Enterprise Linux"
VERSION="8.8 (Ootpa)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="8.8"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Red Hat Enterprise Linux 8.8 (Ootpa)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:8::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 8"
REDHAT_BUGZILLA_PRODUCT_VERSION=8.8
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="8.8"
```

If you send request to nodejs app it will fail with 503:

```
>>>
GET / HTTP/1.1
Host: localhost
Connection: close


<<<
HTTP/1.1 503 Service Unavailable
Content-Type: text/html
Server: Unit/1.32.0
Date: Thu, 08 Feb 2024 16:53:49 GMT
Content-Length: 54
Connection: close

<!DOCTYPE html><title>Error 503</title><p>Error 503.
```

Error message from unit.log:


```internal/modules/cjs/loader.js:638
    throw err;
    ^

Error: Cannot find module 'node:process'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/tmp/unit-test-2wn2yn68/node/node_modules/unit-http/http_server.js:8:20)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
```

From my side, I can suggest just dropping error reporting (that uses `node:process`) for now and committing it later once RHEL8 is outdated. It will allow us to fix issues without breaking any existing builds at the same time.

andrey-zelenkov:
fyi, applying this diff fixes the problem:

```diff --git a/src/nodejs/unit-http/http_server.js b/src/nodejs/unit-http/http_server.js
--- a/src/nodejs/unit-http/http_server.js
+++ b/src/nodejs/unit-http/http_server.js
@@ -5,7 +5,6 @@
 
 'use strict';
 
-const { stderr } = require('node:process');
 const EventEmitter = require('events');
 const http = require('http');
 const util = require('util');
@@ -418,8 +417,6 @@ function Server(options, requestListener
     if (typeof options === 'function') {
         requestListener = options;
         options = {};
-    } else {
-        stderr.write("http.Server constructor was called with unsupported options, using default settings\n");
     }
 
     EventEmitter.call(this);
diff --git a/test/test_node_application.py b/test/test_node_application.py
--- a/test/test_node_application.py
+++ b/test/test_node_application.py
@@ -21,11 +21,10 @@ def test_node_application_basic():
 
     assert_basic_application()
 
-def test_node_application_options(wait_for_record):
+def test_node_application_options():
     client.load('options')
 
     assert_basic_application()
-    assert wait_for_record(r'constructor was called with unsupported') is not None
 
 
 def test_node_application_loader_unit_http():
```

callahad:
What version of Node is in use on RHEL 8?

callahad:
Whatever it is, I'd expect `require('process')` to work reliably everywhere.

andrey-zelenkov:
> What version of Node is in use on RHEL 8?

Sorry, forgot most important part:

```
./configure nodejs
...
configuring nodejs module
checking for node ... found
 + node version v10.24.0
checking for npm ... found
 + npm version 6.14.11
checking for node-gyp ... found
 + node-gyp version v9.4.0
```

callahad:
Awesome, thank you for providing that :) Using `require('process')` instead of `require('node:process')` will work on Node 10. Technically, the `require()` call isn't necessary on any version, as `process` is a Node.js global, but it doesn't hurt to call it.

Separately, we should consider this case when discussing supported platforms. Node 10 has been end-of-life for nearly three years, and even RedHat [won't provide paid support for it](https://access.redhat.com/articles/3376841#NODE_10) (or v12 or v14) on RHEL. For those customers, the oldest commercially supported version is Node 16 on RHEL 8.7 or greater.

andrey-zelenkov:
> Whatever it is, I'd expect require('process') to work reliably everywhere.

That works!

ac000:
For the first patch please put the `Closes:` (with a `:``) tag at the bottom of the commit message, I.e

```
Take options as well as requestListener

Unit-http have not kept up with the signature of nodejs's http package
development. Nodejs allows an optional `options` object to be passed to
the `createServer` function, we didn't. This resulted in function
signature errors when user code that did make use of the options arg
tried to call unit's replaced function.

This change changes the signature to be more in line with how nodejs
does it discarding it and printing a message to stdout.

Closes: https://github.com/nginx/unit/issues/1043
```

ac000 (approved):

callahad:
Heh, during all the discussion about `process.stderr` I somehow glossed over the fact that we could just call `console.warn`, and it'd arguably be slightly more idiomatic. Suggested that in #1135

PR #1094, opened 2024-01-26
Fix isolation uid/gidmaps on arm64

ac000:
This pull-request fixes an issue on arm64 when using user namespaces on Linux whereby the "container", "host" and "size" members of the uid/gidmaps could be corrupted. 

This would happen due to using nxt_int_t's to store these things, but on arm64 (and potentially others) these are 64bits. However there was an assumption that these would only be 32bits and due to setting these variables through a 32bit union alias, we would only set the four lower half bytes, leaving whatever was around in the upper half, giving some whacked out values.

It consists of the following five commits.

- Isolation: Add a new nxt_cred_t type

This adds a new type for holding uid/gids on Linux when using user namespaces.

- Configuration: Add a new CRED map type

This adds a new configuration map type that corresponds to the above.

- Isolation: Use an appropriate type for storing uid/gids

This is the meat and potatoes of the series. It switches over to using the above two new items.

The following two commits are some cleanup allowed by the previous.

- Configuration: Use the NXT_CONF_VLDT_REQUIRED flag for procmap

Each of the "container", "host" and "size" parameters are _required_.

- Configuration: Remove procmap validation code

This removes the custom validation of the above three items, it's sole purpose was to check that they were set and now we _know_ they will be.

### DIFF
--- a/src/nxt_clone.c
+++ b/src/nxt_clone.c
@@ -143,7 +143,7 @@ nxt_clone_credential_map_set(nxt_task_t *task, const char* mapfile, pid_t pid,
         end = mapinfo + len;
 
         for (i = 0; i < map->size; i++) {
-            p = nxt_sprintf(p, end, "%d %d %d", map->map[i].container,
+            p = nxt_sprintf(p, end, "%L %L %L", map->map[i].container,
                             map->map[i].host, map->map[i].size);
 
             if (nxt_slow_path(p == end)) {
@@ -332,7 +332,7 @@ nxt_clone_vldt_credential_gidmap(nxt_task_t *task,
 
         if (nxt_slow_path((nxt_gid_t) m.host != nxt_egid)) {
             nxt_log(task, NXT_LOG_ERR, "\"gidmap\" field has an entry for "
-                    "host gid %d but unprivileged unit can only map itself "
+                    "host gid %L but unprivileged unit can only map itself "
                     "(gid %d) into child namespaces.", m.host, nxt_egid);
 
             return NXT_ERROR;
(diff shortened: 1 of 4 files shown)


ac000:
Changes:

- Use an int64_t as the base type for nxt_cred_t rather than unsigned int, so we can make use of the existing NXT_CONF_MAP_INT64 type
- Drop the patch that added a NXT_CONF_MAP_CRED type as we are now just using the above mentioned NXT_CONF_MAP_INT64

hongzhidao (approved):

ac000:
Add Zhidao's Reviewed-by

```
$ git range-diff aa091295...25db3123
1:  4927f5bf ! 1:  c1e747a4 Isolation: Add a new nxt_cred_t type
    @@ Commit message
     
         This will be used in subsequent commits.
     
      ## src/nxt_clone.h ##
2:  09b9eed0 ! 2:  0a0de26b Isolation: Use an appropriate type for storing uid/gids
    @@ Commit message
         variables.
     
         Reported-by: Andrei Zeliankou <zelenkov@nginx.com>
     
      ## src/nxt_clone.c ##
3:  a1517e77 ! 3:  726c6885 Configuration: Use the NXT_CONF_VLDT_REQUIRED flag for procmap
    @@ Commit message
         These are for the uidmap & gidmap settings in the config.
     
         Suggested-by: Zhidao HONG <z.hong@f5.com>
     
4:  aa091295 ! 4:  25db3123 Configuration: Remove procmap validation code
    @@ Commit message
         validation, it's only purpose was to check if the three uidmap/gidmap
         settings had been provided.
     
```

ac000:
Rebased with master.

PR #1108, opened 2024-01-31
Config: Store abstract sockets with @ internally

alejandro-colomar:
We accept both \u0000socket-name and @socket-name as abstract Unix sockets.  The first one is passed to the kernel pristine, while the second is transformed '@'->'\0'.

The commit that added support for Unix sockets accepts both variants, but we internally stored it in the same way, using \u0000... for both.

We want to support abstract sockets transparently to the user, so that if the user configures unitd with '@', if we receive a query about the current configuration, the user should see the same exact thing that was configured.  So, this commit avoids the transformation in the internal state file, storing user input pristine, and we only transform the '@' for a string that will be used internally (not user-visible).

This commit (indirectly) fixes a small bug, where we created abstract sockets with a trailing '\0' in their name due to calling twice nxt_sockaddr_parse() on the same string.  By calling that function only once with each copy of the string, we have fixed that bug.

This bug was found thanks to some experiment about using 'const' for some strings.

And here's some history:

-  9041d276fc6a ("nxt_sockaddr_parse() introducted.")

   This commit introduced support for abstract Unix sockets, but they only worked as "servers", and not as "listeners".  We modified (corrupted?) the JSON config file, and stored a \u0000.

-  d8e0768a5bae ("Fixed support for abstract Unix sockets.")

   This commit (partially) fixed support for abstract Unix sockets, so they they worked also as listeners.  We still modified the JSON config file, and stored a \u0000.

   That next commit also documents some issue about a terminating '\0', but I don't remember the details about it.  I'm not sure it it was introduced here or before.

-  e2aec6686a4d ("Storing abstract sockets with @ internally.")

   This commit fixed the problem by which we were modifying the config file, but only for "listeners", not for "servers".  (It also fixes the issue about the terminating '\0'.)  We completely forgot about "servers".

To reproduce the problem, I used the following config:

```json
{
	"listeners": {
		"*:80": {
			"pass": "routes/u"
		},
		"unix:@abstract": {
			"pass": "routes/a"
		}
	},

	"routes": {
		"u": [{
			"action": {
				"pass": "upstreams/u"
			}
		}],
		"a": [{
			"action": {
				"return": 302,
				"location": "/i/am/not/at/home/"
			}
		}]
	},

	"upstreams": {
		"u": {
			"servers": {
				"unix:@abstract": {}
			}
		}
	}
}
```

And then check the state file:

```sh
$ sudo cat /opt/local/nginx/unit/master/var/lib/unit/conf.json \
| jq . \
| grep unix;
    "unix:@abstract": {
        "unix:\u0000abstract": {}
```

After this patch, the state file has a '@' as expected:
    
```sh
$ sudo cat /opt/local/nginx/unit/unix/var/lib/unit/conf.json \
| jq . \
| grep unix;
    "unix:@abstract": {
        "unix:@abstract": {}
```

Fixes: e2aec6686a4d ("Storing abstract sockets with @ internally.")
Fixes: 9041d276fc6a ("nxt_sockaddr_parse() introducted.")
Cc: @lcrilly

### DIFF
--- a/src/nxt_conf_validation.c
+++ b/src/nxt_conf_validation.c
@@ -2,6 +2,7 @@
 /*
  * Copyright (C) Valentin V. Bartenev
  * Copyright (C) NGINX, Inc.
+ * Copyright 2024, Alejandro Colomar <alx@kernel.org>
  */
 
 #include <nxt_main.h>
@@ -1936,10 +1937,13 @@ static nxt_int_t
 nxt_conf_vldt_proxy(nxt_conf_validation_t *vldt, nxt_conf_value_t *value,
     void *data)
 {
-    nxt_str_t       name;
+    nxt_str_t       name, *ret;
     nxt_sockaddr_t  *sa;
 
-    nxt_conf_get_string(value, &name);
+    ret = nxt_conf_get_string_dup(value, vldt->pool, &name);
+    if (nxt_slow_path(ret == NULL)) {
+        return NXT_ERROR;
+    }
 
     if (nxt_str_start(&name, "http://", 7)) {
         name.length -= 7;
@@ -2913,13 +2917,11 @@ nxt_conf_vldt_object_iterator(nxt_conf_validation_t *vldt,
 
     for ( ;; ) {
         member = nxt_conf_next_object_member(value, &name, &index);
-
         if (member == NULL) {
             return NXT_OK;
         }
 
         ret = validator(vldt, &name, member);
-
         if (ret != NXT_OK) {
             return ret;
         }
@@ -3268,16 +3270,19 @@ nxt_conf_vldt_server(nxt_conf_validation_t *vldt, nxt_str_t *name,
     nxt_conf_value_t *value)
 {
     nxt_int_t       ret;
+    nxt_str_t       str;
     nxt_sockaddr_t  *sa;
 
     ret = nxt_conf_vldt_type(vldt, name, value, NXT_CONF_VLDT_OBJECT);
-
     if (ret != NXT_OK) {
         return ret;
     }
 
-    sa = nxt_sockaddr_parse(vldt->pool, name);
+    if (nxt_slow_path(nxt_str_dup(vldt->pool, &str, name) == NULL)) {
+        return NXT_ERROR;
+    }
 
+    sa = nxt_sockaddr_parse(vldt->pool, &str);
     if (sa == NULL) {
         return nxt_conf_vldt_error(vldt, "The \"%V\" is not valid "
                                    "server address.", name);
--- a/src/nxt_conf.c
+++ b/src/nxt_conf.c
@@ -3,6 +3,7 @@
  * Copyright (C) Igor Sysoev
  * Copyright (C) Valentin V. Bartenev
  * Copyright (C) NGINX, Inc.
+ * Copyright 2024, Alejandro Colomar <alx@kernel.org>
  */
 
 #include <nxt_main.h>
@@ -171,6 +172,16 @@ nxt_conf_get_string(nxt_conf_value_t *value, nxt_str_t *str)
         str->length = value->u.string.length;
         str->start = value->u.string.start;
     }
+}
+
+
+nxt_str_t *
+nxt_conf_get_string_dup(nxt_conf_value_t *value, nxt_mp_t *mp, nxt_str_t *str)
+{
+    nxt_str_t  s;
+
+    nxt_conf_get_string(value, &s);
+    return nxt_str_dup(mp, str, &s);
 }
 
 
--- a/src/nxt_conf.h
+++ b/src/nxt_conf.h
@@ -3,6 +3,7 @@
  * Copyright (C) Igor Sysoev
  * Copyright (C) Valentin V. Bartenev
  * Copyright (C) NGINX, Inc.
+ * Copyright 2024, Alejandro Colomar <alx@kernel.org>
  */
 
 #ifndef _NXT_CONF_INCLUDED_
@@ -116,6 +117,8 @@ void nxt_conf_json_position(u_char *start, const u_char *pos, nxt_uint_t *line,
 nxt_int_t nxt_conf_validate(nxt_conf_validation_t *vldt);
 
 NXT_EXPORT void nxt_conf_get_string(nxt_conf_value_t *value, nxt_str_t *str);
+NXT_EXPORT nxt_str_t *nxt_conf_get_string_dup(nxt_conf_value_t *value,
+    nxt_mp_t *mp, nxt_str_t *str);
 NXT_EXPORT void nxt_conf_set_string(nxt_conf_value_t *value, nxt_str_t *str);
 NXT_EXPORT nxt_int_t nxt_conf_set_string_dup(nxt_conf_value_t *value,
     nxt_mp_t *mp, const nxt_str_t *str);
--- a/src/nxt_http_static.c
+++ b/src/nxt_http_static.c
@@ -117,9 +117,7 @@ nxt_http_static_init(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
         nxt_str_set(&conf->index, "index.html");
 
     } else {
-        nxt_conf_get_string(acf->index, &str);
-
-        ret = nxt_str_dup(mp, &conf->index, &str);
+        ret = nxt_conf_get_string_dup(acf->index, mp, &conf->index);
         if (nxt_slow_path(ret == NULL)) {
             return NXT_ERROR;
         }
--- a/src/nxt_router.c
+++ b/src/nxt_router.c
@@ -2275,7 +2275,7 @@ nxt_router_conf_process_static(nxt_task_t *task, nxt_router_conf_t *rtcf,
 {
     uint32_t          next, i;
     nxt_mp_t          *mp;
-    nxt_str_t         *type, exten, str;
+    nxt_str_t         *type, exten, str, *s;
     nxt_int_t         ret;
     nxt_uint_t        exts;
     nxt_conf_value_t  *mtypes_conf, *ext_conf, *value;
@@ -2311,9 +2311,8 @@ nxt_router_conf_process_static(nxt_task_t *task, nxt_router_conf_t *rtcf,
             }
 
             if (nxt_conf_type(ext_conf) == NXT_CONF_STRING) {
-                nxt_conf_get_string(ext_conf, &str);
-
-                if (nxt_slow_path(nxt_str_dup(mp, &exten, &str) == NULL)) {
+                s = nxt_conf_get_string_dup(ext_conf, mp, &exten);
+                if (nxt_slow_path(s == NULL)) {
                     return NXT_ERROR;
                 }
 
@@ -2331,9 +2330,8 @@ nxt_router_conf_process_static(nxt_task_t *task, nxt_router_conf_t *rtcf,
             for (i = 0; i < exts; i++) {
                 value = nxt_conf_get_array_element(ext_conf, i);
 
-                nxt_conf_get_string(value, &str);
-
-                if (nxt_slow_path(nxt_str_dup(mp, &exten, &str) == NULL)) {
+                s = nxt_conf_get_string_dup(value, mp, &exten);
+                if (nxt_slow_path(s == NULL)) {
                     return NXT_ERROR;
                 }
 
@@ -2425,14 +2423,11 @@ static nxt_int_t
 nxt_router_conf_forward_header(nxt_mp_t *mp, nxt_conf_value_t *conf,
     nxt_http_forward_header_t *fh)
 {
-    char       c;
-    size_t     i;
-    uint32_t   hash;
-    nxt_str_t  header;
+    char      c;
+    size_t    i;
+    uint32_t  hash;
 
-    nxt_conf_get_string(conf, &header);
-
-    fh->header = nxt_str_dup(mp, NULL, &header);
+    fh->header = nxt_conf_get_string_dup(conf, mp, NULL);
     if (nxt_slow_path(fh->header == NULL)) {
         return NXT_ERROR;
     }


ac000:
Hi Alex,

Just to be clear, this doesn't change what we send to the Kernel right? As I remember in the past when looking at this stuff, while it looked right, when you looked at the socket under ss(8) it wasn't quite right...

alejandro-colomar:
> Hi Alex,
> 
> Just to be clear, this doesn't change what we send to the Kernel right?

Correct.

> As I remember in the past when looking at this stuff, while it looked right, when you looked at the socket under ss(8) it wasn't quite right...

Yeah, that's the terminating '\0' that I mention in the commit.  I don't remember when/how that was introduced.  Could you please check yourself both before and after my patch?

alejandro-colomar:
Per my tests, it seems to be correct, both before and after.

alejandro-colomar:
We should have kept those commits separate, as they fix different stuff.  Now I can't remember what was the problem that we fixed, and can't see what was the fix.  :|

alejandro-colomar:
Ahhh, I finally got it.  As suspected this fixes two bugs.  I'll make sure to document this.

```
$ sudo strace -f -e 'bind,connect' /opt/local/nginx/unit/d8e0/sbin/unitd |& grep abstract
[pid 22406] bind(10, {sa_family=AF_UNIX, sun_path=@"abstract\0"}, 12) = 0
[pid 22410] connect(134, {sa_family=AF_UNIX, sun_path=@"abstract\0"}, 12) = 0
^C
$ sudo killall unitd
$ sudo strace -f -e 'bind,connect' /opt/local/nginx/unit/master/sbin/unitd |& grep abstract
[pid 22449] bind(10, {sa_family=AF_UNIX, sun_path=@"abstract"}, 11) = 0
[pid 22453] connect(134, {sa_family=AF_UNIX, sun_path=@"abstract\0"}, 12) = -1 ECONNREFUSED (Connection refused)
^C
$ sudo killall unitd
$ sudo strace -f -e 'bind,connect' /opt/local/nginx/unit/unix/sbin/unitd |& grep abstract
[pid 22488] bind(10, {sa_family=AF_UNIX, sun_path=@"abstract"}, 11) = 0
[pid 22492] connect(134, {sa_family=AF_UNIX, sun_path=@"abstract"}, 11) = 0
^C
```

So, master currently is sending a bogus `\0` to the kernel, and this patch fixes that (as a side effect of the other fix).

alejandro-colomar:
> Ahhh, I finally got it. As suspected this fixes two bugs. I'll make sure to document this.
> 
> ```
> $ sudo strace -f -e 'bind,connect' /opt/local/nginx/unit/d8e0/sbin/unitd |& grep abstract
> [pid 22406] bind(10, {sa_family=AF_UNIX, sun_path=@"abstract\0"}, 12) = 0
> [pid 22410] connect(134, {sa_family=AF_UNIX, sun_path=@"abstract\0"}, 12) = 0
> ^C
> $ sudo killall unitd
> $ sudo strace -f -e 'bind,connect' /opt/local/nginx/unit/master/sbin/unitd |& grep abstract
> [pid 22449] bind(10, {sa_family=AF_UNIX, sun_path=@"abstract"}, 11) = 0
> [pid 22453] connect(134, {sa_family=AF_UNIX, sun_path=@"abstract\0"}, 12) = -1 ECONNREFUSED (Connection refused)
> ^C
> $ sudo killall unitd
> $ sudo strace -f -e 'bind,connect' /opt/local/nginx/unit/unix/sbin/unitd |& grep abstract
> [pid 22488] bind(10, {sa_family=AF_UNIX, sun_path=@"abstract"}, 11) = 0
> [pid 22492] connect(134, {sa_family=AF_UNIX, sun_path=@"abstract"}, 11) = 0
> ^C
> ```
> 
> So, master currently is sending a bogus `\0` to the kernel, and this patch fixes that (as a side effect of the other fix). I'm not entirely sure of what was the cause of the bogus `\0`, though; I don't remember. If you find it, I can document it, and if not, I'll just document the experiment.

Bug found.

```
nxt_sockaddr_parse()
  nxt_sockaddr_parse_optport()
    nxt_sockaddr_unix_parse()
```

```c
$ grepc -tfd nxt_sockaddr_unix_parse . | grep -A10 @
    if (path[0] == '@') {
        path[0] = '\0';
        socklen--;
#if !(NXT_LINUX)
        nxt_thread_log_error(NXT_LOG_ERR,
                             "abstract unix domain sockets are not supported");
        return NULL;
#endif
    }

    sa = nxt_sockaddr_alloc(mp, socklen, addr->length);
```

So, the second time this function is called with the same name, which has been corrupted to contain a `'\0'` instead of an `'@'`. is considered as a file-backed Unix socket, instead of an abstract Unix socket, and so we don't execute that `socklen--`.  Since the `nxt_sockaddr_alloc()` call right after that does an internal `nxt_mp_zalloc()`, the spurious byte is a `'\0'`.  It seems that the kernel is ignoring that bogus byte, but we better remove it.

Now I'm happy about it.  I'll document that.

alejandro-colomar:
v1b changes:

-  Document second bug fix in commit message [@ac000 ]

```
$ git range-diff master gh/unix unix 
1:  5b26baf4 ! 1:  4ab438f5 Config: Store abstract sockets with @ internally
    @@ Commit message
         state file, storing user input pristine, and we only transform the '@'
         for a string that will be used internally (not user-visible).
     
    -    This commit (indirectly) fixes a small bug, where we created abstract
    +    This commit fixes another bug, where we try to connect to abstract
         sockets with a trailing '\0' in their name due to calling twice
         nxt_sockaddr_parse() on the same string.  By calling that function only
         once with each copy of the string, we have fixed that bug.
     
    +    The following code was responsible for this bug, which the second time
    +    it was called, considered these sockets as file-backed (not abstract)
    +    Unix socket, and so appended a '\0' to the socket name.
    +
    +        $ grepc -tfd nxt_sockaddr_unix_parse . | grep -A10 @
    +            if (path[0] == '@') {
    +                path[0] = '\0';
    +                socklen--;
    +        #if !(NXT_LINUX)
    +                nxt_thread_log_error(NXT_LOG_ERR,
    +                                     "abstract unix domain sockets are not supported");
    +                return NULL;
    +        #endif
    +            }
    +
    +            sa = nxt_sockaddr_alloc(mp, socklen, addr->length);
    +
         This bug was found thanks to some experiment about using 'const' for
         some strings.
     
    @@ Commit message
     
            This commit introduced support for abstract Unix sockets, but they
            only worked as "servers", and not as "listeners".  We modified
    -       (corrupted?) the JSON config file, and stored a \u0000.
    +       (corrupted?) the JSON config file, and stored a \u0000.  This also
    +       caused calling connect(2) with a bogus trailing null byte, which the
    +       kernel silently ignored.
     
         -  d8e0768a5bae ("Fixed support for abstract Unix sockets.")
     
            This commit (partially) fixed support for abstract Unix sockets, so
            they they worked also as listeners.  We still modified the JSON
    -       config file, and stored a \u0000.
    -
    -       That next commit also documents some issue about a terminating '\0',
    -       but I don't remember the details about it.  I'm not sure it it was
    -       introduced here or before.
    +       config file, and stored a \u0000.  This caused calling connect(2)
    +       (and now bind(2) too) with a bogus trailing null byte.
     
         -  e2aec6686a4d ("Storing abstract sockets with @ internally.")
     
    @@ Commit message
                 "unix:@abstract": {
                     "unix:@abstract": {}
     
    -    Fixes: e2aec6686a4d ("Storing abstract sockets with @ internally.")
    +    Regarding the trailing null byte, here are some tests:
    +
    +        $ sudo strace -f -e 'bind,connect' /opt/local/nginx/unit/d8e0/sbin/unitd \
    +        |& grep abstract;
    +        [pid 22406] bind(10, {sa_family=AF_UNIX, sun_path=@"abstract\0"}, 12) = 0
    +        [pid 22410] connect(134, {sa_family=AF_UNIX, sun_path=@"abstract\0"}, 12) = 0
    +        ^C
    +        $ sudo killall unitd
    +        $ sudo strace -f -e 'bind,connect' /opt/local/nginx/unit/master/sbin/unitd \
    +        |& grep abstract;
    +        [pid 22449] bind(10, {sa_family=AF_UNIX, sun_path=@"abstract"}, 11) = 0
    +        [pid 22453] connect(134, {sa_family=AF_UNIX, sun_path=@"abstract\0"}, 12) = -1 ECONNREFUSED (Connection refused)
    +        ^C
    +        $ sudo killall unitd
    +        $ sudo strace -f -e 'bind,connect' /opt/local/nginx/unit/unix/sbin/unitd \
    +        |& grep abstract;
    +        [pid 22488] bind(10, {sa_family=AF_UNIX, sun_path=@"abstract"}, 11) = 0
    +        [pid 22492] connect(134, {sa_family=AF_UNIX, sun_path=@"abstract"}, 11) = 0
    +        ^C
    +
         Fixes: 9041d276fc6a ("nxt_sockaddr_parse() introducted.")
    +    Fixes: d8e0768a5bae ("Fixed support for abstract Unix sockets.")
    +    Fixes: e2aec6686a4d ("Storing abstract sockets with @ internally.")
    +    Cc: Andrew Clayton <a.clayton@nginx.com>
         Cc: Liam Crilly <liam.crilly@nginx.com>
     
```

alejandro-colomar:
v1c changes:

-  Say corrupted.

```
$ git range-diff master gh/unix unix 
1:  4ab438f5 ! 1:  c9705581 Config: Store abstract sockets with @ internally
    @@ Commit message
         -  9041d276fc6a ("nxt_sockaddr_parse() introducted.")
     
            This commit introduced support for abstract Unix sockets, but they
    -       only worked as "servers", and not as "listeners".  We modified
    -       (corrupted?) the JSON config file, and stored a \u0000.  This also
    -       caused calling connect(2) with a bogus trailing null byte, which the
    -       kernel silently ignored.
    +       only worked as "servers", and not as "listeners".  We corrupted the
    +       JSON config file, and stored a \u0000.  This also caused calling
    +       connect(2) with a bogus trailing null byte, which the kernel silently
    +       ignored.
     
         -  d8e0768a5bae ("Fixed support for abstract Unix sockets.")
     
            This commit (partially) fixed support for abstract Unix sockets, so
    -       they they worked also as listeners.  We still modified the JSON
    +       they they worked also as listeners.  We still corrupted the JSON
            config file, and stored a \u0000.  This caused calling connect(2)
            (and now bind(2) too) with a bogus trailing null byte.
     
         -  e2aec6686a4d ("Storing abstract sockets with @ internally.")
     
    -       This commit fixed the problem by which we were modifying the config
    +       This commit fixed the problem by which we were corrupting the config
            file, but only for "listeners", not for "servers".  (It also fixes
            the issue about the terminating '\0'.)  We completely forgot about
            "servers".
```

alejandro-colomar:
v1d changes:

-  connect(2) was actually failing due to that bogus `'\0'`.  In some of my tests it seemed to work just because the listener was equally bogus.  Fix wording of commit message.

```
$ git range-diff master gh/unix unix 
1:  c9705581 ! 1:  560aa7b9 Config: Store abstract sockets with @ internally
    @@ Commit message
            This commit introduced support for abstract Unix sockets, but they
            only worked as "servers", and not as "listeners".  We corrupted the
            JSON config file, and stored a \u0000.  This also caused calling
    -       connect(2) with a bogus trailing null byte, which the kernel silently
    -       ignored.
    +       connect(2) with a bogus trailing null byte, which tried to connect to
    +       a different abstract socket.
     
         -  d8e0768a5bae ("Fixed support for abstract Unix sockets.")
     
```

alejandro-colomar:
v1e changes:

-  Rebase to master

```
$ git range-diff master..gh/unix ngx/master..unix 
1:  560aa7b9 = 1:  aa013acf Config: Store abstract sockets with @ internally
```

alejandro-colomar:
v1f changes:

-  Subject line: `Configuration: ...`

```
$ git range-diff master..gh/unix ngx/master..unix 
1:  aa013acf ! 1:  536c60ed Config: Store abstract sockets with @ internally
    @@ Metadata
     
      ## Commit message ##
    -    Config: Store abstract sockets with @ internally
    +    Configuration: Store abstract sockets with @ internally
     
         We accept both \u0000socket-name and @socket-name as abstract Unix
         sockets.  The first one is passed to the kernel pristine, while the
```

alejandro-colomar:
I'll use a different approach: I'll make the string copy inside nxt_sockaddr_parse(), so that it fixes any other instances of the same bug all in one place.  I'll thus remove the old copy we made for the listeners.

ac000:
On Thu, 01 Feb 2024 04:00:57 -0800

> I'll use a different approach: I'll make the string copy inside nxt_sockaddr_parse(), so that it fixes any other instances of the same bug all in one place.  I'll thus remove the old copy we made for the listeners.
> 

Sounds good.

src/nxt_sockaddr.c
```diff
@@ -572,8 +573,8 @@ static nxt_sockaddr_t *
 nxt_sockaddr_unix_parse(nxt_mp_t *mp, nxt_str_t *addr)
 {
 #if (NXT_HAVE_UNIX_DOMAIN)
-    size_t          length, socklen;
-    u_char          *path;
+    size_t          socklen;
+    nxt_str_t       *path;
```
ac000, line 577: Hi Alex,

What's the motivation for switching path to a nxt_str_t?

I'm not seeing an obvious reason and not doing so would keep the patch simpler...

Hmm, maybe I'm missing something, but that seems to be the only change here... or maybe I should look at this thing properly in git!

alejandro-colomar: You're right.  I did a different change which needed the nxt_str_t, and then I realized I could simplify a lot of it, but didn't see that I could drop even more changes.  :)

alejandro-colomar:
v3 changes:

-  Reduce the diff [@ac000 ]

```
$ git range-diff master gh/unix unix 
1:  f6333c7c ! 1:  ceda6da0 Configuration: Store abstract sockets with @ internally
    @@ src/nxt_sockaddr.c
       */
      
      #include <nxt_main.h>
    -@@ src/nxt_sockaddr.c: static nxt_sockaddr_t *
    - nxt_sockaddr_unix_parse(nxt_mp_t *mp, nxt_str_t *addr)
    - {
    +@@ src/nxt_sockaddr.c: nxt_sockaddr_unix_parse(nxt_mp_t *mp, nxt_str_t *addr)
      #if (NXT_HAVE_UNIX_DOMAIN)
    --    size_t          length, socklen;
    --    u_char          *path;
    -+    size_t          socklen;
    -+    nxt_str_t       *path;
    +     size_t          length, socklen;
    +     u_char          *path;
    ++    nxt_str_t       *str;
          nxt_sockaddr_t  *sa;
      
          /*
    @@ src/nxt_sockaddr.c: nxt_sockaddr_unix_parse(nxt_mp_t *mp, nxt_str_t *addr)
          const size_t max_len = sizeof(struct sockaddr_un)
                                 - offsetof(struct sockaddr_un, sun_path) - 1;
      
    --    /* Cutting "unix:". */
    --    length = addr->length - 5;
    --    path = addr->start + 5;
    -+    path = nxt_str_dup(mp, NULL, addr);
    -+    if (nxt_slow_path(path == NULL)) {
    ++    str = nxt_str_dup(mp, NULL, addr);
    ++    if (nxt_slow_path(str == NULL)) {
     +        return NULL;
     +    }
    - 
    --    if (length > max_len) {
    -+    /* Cutting "unix:". */
    -+    path->length -= 5;
    -+    path->start += 5;
     +
    -+    if (path->length > max_len) {
    +     /* Cutting "unix:". */
    +-    length = addr->length - 5;
    +-    path = addr->start + 5;
    ++    length = str->length - 5;
    ++    path = str->start += 5;
    + 
    +     if (length > max_len) {
              nxt_thread_log_error(NXT_LOG_ERR,
    -                              "unix domain socket \"%V\" name is too long",
    -                              addr);
    -         return NULL;
    -     }
    - 
    --    socklen = offsetof(struct sockaddr_un, sun_path) + length + 1;
    -+    socklen = offsetof(struct sockaddr_un, sun_path) + path->length + 1;
    - 
    -     /*
    -      * Linux unix(7):
    -@@ src/nxt_sockaddr.c: nxt_sockaddr_unix_parse(nxt_mp_t *mp, nxt_str_t *addr)
    -      *   are covered by the specified length of the address structure.
    -      *   (Null bytes in the name have no special significance.)
    -      */
    --    if (path[0] == '@') {
    --        path[0] = '\0';
    -+    if (path->start[0] == '@') {
    -+        path->start[0] = '\0';
    -         socklen--;
    - #if !(NXT_LINUX)
    -         nxt_thread_log_error(NXT_LOG_ERR,
    -@@ src/nxt_sockaddr.c: nxt_sockaddr_unix_parse(nxt_mp_t *mp, nxt_str_t *addr)
    - 
    -     if (nxt_fast_path(sa != NULL)) {
    -         sa->u.sockaddr_un.sun_family = AF_UNIX;
    --        nxt_memcpy(sa->u.sockaddr_un.sun_path, path, length);
    -+        nxt_memcpy(sa->u.sockaddr_un.sun_path, path->start, path->length);
    -     }
    - 
    -     return sa;
```

alejandro-colomar:
v3b changes:

-  tfix

```
$ git range-diff master gh/unix unix 
1:  ceda6da0 ! 1:  f25320d0 Configuration: Store abstract sockets with @ internally
    @@ src/nxt_sockaddr.c: nxt_sockaddr_unix_parse(nxt_mp_t *mp, nxt_str_t *addr)
     -    length = addr->length - 5;
     -    path = addr->start + 5;
     +    length = str->length - 5;
    -+    path = str->start += 5;
    ++    path = str->start + 5;
      
          if (length > max_len) {
              nxt_thread_log_error(NXT_LOG_ERR,
```

src/nxt_sockaddr.c
```diff
@@ -588,9 +590,14 @@ nxt_sockaddr_unix_parse(nxt_mp_t *mp, nxt_str_t *addr)
...
+    str = nxt_str_dup(mp, NULL, addr);
+    if (nxt_slow_path(str == NULL)) {
+        return NULL;
+    }
+
     /* Cutting "unix:". */
-    length = addr->length - 5;
-    path = addr->start + 5;
+    length = str->length - 5;
+    path = str->start + 5;
```
ac000, line 600: Heh, I think I must be being dense!

But why can't we just use addr directly?

alejandro-colomar:
```c
$ grepc -tfd nxt_conf_vldt_server .
./src/nxt_conf_validation.c:static nxt_int_t
nxt_conf_vldt_server(nxt_conf_validation_t *vldt, const nxt_rstr_t *name,
    nxt_conf_value_t *value)
{
    ...

    sa = nxt_sockaddr_parse(vldt->pool, name);

    ...
}
```

`name` is a config string, read-only in my tree.  Showing where it comes from would be complex, but let's say it is mapped to the config (if you edit it, you'll see).

Then `nxt_sockaddr_parse()` will pass that `name` to what `nxt_sockaddr_unix_parse()` knows as `addr`.

ac000: I.e why do we need to strdup it?

alejandro-colomar:
> We accept both \u0000socket-name and @socket-name as abstract Unix sockets. The first one is passed to the kernel pristine, while the second is transformed '@'->'\0'.

Maybe I should reword that.  We don't really support abstract Unix sockets with `\u0000`.  It just happens to work, because we don't explicitly reject it, but it certainly behaves wrong, by appending a `'\0'`

src/nxt_sockaddr.c
alejandro-colomar: Try not to dup it, and inspect `$prefix/var/lib/unit/conf.json` :)

alejandro-colomar:
```
        $ sudo strace -f -e 'bind,connect' /opt/local/nginx/unit/d8e0/sbin/unitd \
        |& grep abstract;
        [pid 22406] bind(10, {sa_family=AF_UNIX, sun_path=@"abstract\0"}, 12) = 0
        [pid 22410] connect(134, {sa_family=AF_UNIX, sun_path=@"abstract\0"}, 12) = 0
        ^C
        $ sudo killall unitd
        $ sudo strace -f -e 'bind,connect' /opt/local/nginx/unit/master/sbin/unitd \
        |& grep abstract;
        [pid 22449] bind(10, {sa_family=AF_UNIX, sun_path=@"abstract"}, 11) = 0
        [pid 22453] connect(134, {sa_family=AF_UNIX, sun_path=@"abstract\0"}, 12) = -1 ECONNREFUSED (Connection refused)
        ^C
        $ sudo killall unitd
        $ sudo strace -f -e 'bind,connect' /opt/local/nginx/unit/unix/sbin/unitd \
        |& grep abstract;
        [pid 22488] bind(10, {sa_family=AF_UNIX, sun_path=@"abstract"}, 11) = 0
        [pid 22492] connect(134, {sa_family=AF_UNIX, sun_path=@"abstract"}, 11) = 0
        ^C
```

alejandro-colomar: See how `master` differs from `unix` (the branch where I've fixed that).

alejandro-colomar: And the JSON:

```
        $ sudo cat /opt/local/nginx/unit/master/var/lib/unit/conf.json \
        | jq . \
        | grep unix;
            "unix:@abstract": {
                "unix:\u0000abstract": {}
 
        $ sudo cat /opt/local/nginx/unit/unix/var/lib/unit/conf.json \
        | jq . \
        | grep unix;
            "unix:@abstract": {
                "unix:@abstract": {}
```

alejandro-colomar: Or try writing `"foo"` to that string, and see the config file after that.  You'll see that the config file is modified.

alejandro-colomar:
v3c changes:

-  Commit message wfix (we don't really support `\u0000`; that's just an implementation detail).

```
$ git range-diff master gh/unix unix 
1:  f25320d0 ! 1:  3a324ede Configuration: Store abstract sockets with @ internally
    @@ Metadata
      ## Commit message ##
         Configuration: Store abstract sockets with @ internally
     
    -    We accept both \u0000socket-name and @socket-name as abstract Unix
    -    sockets.  The first one is passed to the kernel pristine, while the
    -    second is transformed '@'->'\0'.
    -
    -    The commit that added support for Unix sockets accepts both variants,
    -    but we internally stored it in the same way, using \u0000... for both.
    +    The commit that added support for Unix sockets accept abstract sockets
    +    using '@' in the config, but we stored it internally using '\0'.
     
         We want to support abstract sockets transparently to the user, so that
         if the user configures unitd with '@', if we receive a query about the
```

alejandro-colomar:
v3d changes:

-  grfix

```
$ git range-diff master gh/unix unix 
1:  3a324ede ! 1:  91e2a8c1 Configuration: Store abstract sockets with @ internally
    @@ Metadata
      ## Commit message ##
         Configuration: Store abstract sockets with @ internally
     
    -    The commit that added support for Unix sockets accept abstract sockets
    +    The commit that added support for Unix sockets accepts abstract sockets
         using '@' in the config, but we stored it internally using '\0'.
     
         We want to support abstract sockets transparently to the user, so that
```

src/nxt_sockaddr.c
ac000: I didn't see where we're editing it... I need to look at this thing in git...

ac000:
Hi Alex,

Here's nxt_sockaddr_unix_parse() from current master (with comments and #if's removed)

```c
static nxt_sockaddr_t *nxt_sockaddr_unix_parse(nxt_mp_t *mp, nxt_str_t *addr)
{                                                                                                                                  
    size_t          length, socklen;                                            
    u_char          *path;                                                      
    nxt_sockaddr_t  *sa;                                                        
                                                                                
    const size_t max_len = sizeof(struct sockaddr_un)                           
                           - offsetof(struct sockaddr_un, sun_path) - 1;        
                                                                                
    /* Cutting "unix:". */                                                      
    length = addr->length - 5;                                                  
    path = addr->start + 5;                                                     
                                                                                
    if (length > max_len) {                                                     
        nxt_thread_log_error(NXT_LOG_ERR,                                       
                             "unix domain socket \"%V\" name is too long",      
                             addr);                                             
        return NULL;                                                            
    }                                                                           
                                                                                
    socklen = offsetof(struct sockaddr_un, sun_path) + length + 1;              
                                                                                
    if (path[0] == '@') {                                                       
        path[0] = '\0';                                                         
        socklen--;                                                                                                                                    
    }                                                                           
                                                                                
    sa = nxt_sockaddr_alloc(mp, socklen, addr->length);                         
                                                                                
    if (nxt_fast_path(sa != NULL)) {                                            
        sa->u.sockaddr_un.sun_family = AF_UNIX;                                 
        nxt_memcpy(sa->u.sockaddr_un.sun_path, path, length);                   
    }                                                                           
                                                                                
    return sa;
}                                                                               
```

Now apologies, but I'm really struggling to see how strdup'ing addr makes any difference here.

```c
length = str->length - 5;
path = str->start + 5;
```

is the same as what's currently there and `str` is not used anywhere else, I also don't see addr being changed in here and by the time we exit the function we've created our new sockaddr structure.

ac000:
Oh, dang it!

```
path[0] = '\0';
```

OK, I see now...

src/nxt_sockaddr.c
ac000: Yes, I see it now :(

alejandro-colomar: No problem!  It took me a long time to realize it myself  :-)

alejandro-colomar:
Hmmmm, there are reasons we may want to de the copy in `nxt_conf_validation.c` instead of in `nxt_sockaddr_unix_parse()`, even if it means more code:

So, I'm not entirely sure of this, because this code isn't easy to track, but here's how I think it works:

-  We read the config from the control socket, and store it in memory as a series of strings.  I think those strings use a short-lived memory pool, which will only live until the reconfiguration has finished.

-  We call the validation funtions, with those strings.  We better not modify them, as they're the only copy we hold of the configuration.

-  Once we made sure that the configuration is valid, we write those strings we have in memory into the state file (`$prefix/var/lib/unit/conf.json`).

-  We create a long-lived memory pool that will live until the next reconfiguration, and read the new configuration into this memory, as strings.

-  And now we do the actual reconfiguration.  If we messed the strings before this, that's bad.  But if we modify the strings in this step, it's not a problem, because those strings won't be read again.  Doing a copy here would be unnecessary, and would waste memory for a long time.

Any thoughts?  I think I'll roll back to v1.

ac000:
OK, stripping this back to the basics.

```json
{
	"listeners": {
		"unix:@abstract": {
			"pass": "routes/a"
		}
	},
```

That is fine right? Is stored in the state file correctly and the right thing is sent to the kernel...

```json
	"upstreams": {
		"u": {
			"servers": {
				"unix:@abstract": {}
			}
		}
	}
```

This is broken? stored wrongly in the state file and wrong thing sent to the kernel...

Apologies if you've pointed this out already, but what is it that causes these two to be handled differently?

alejandro-colomar:
> OK, stripping this back to the basics.
> 
> ```json
> {
> 	"listeners": {
> 		"unix:@abstract": {
> 			"pass": "routes/a"
> 		}
> 	},
> ```
> 
> That is fine right? Is stored in the state file correctly and the right thing is sent to the kernel...

Yes, that's fine; that was fixed in e2aec6686a4d.

> 
> ```json
> 	"upstreams": {
> 		"u": {
> 			"servers": {
> 				"unix:@abstract": {}
> 			}
> 		}
> 	}
> ```
> 
> This is broken? stored wrongly in the state file and wrong thing sent to the kernel...

Yes, this is broken.  Wrong state file, and wrong thing sent to kernel.

> 
> Apologies if you've pointed this out already, but what is it that causes these two to be handled differently?

Apologies if I didn't explain myself clearly enough :)  Feel free to suggest tweaks to the commit message.

Below is the cause:

```c
$ grepc -rntu 'nxt_sockaddr_parse' src/nxt_conf_validation.c 


src/nxt_conf_validation.c:1681:
static nxt_int_t
nxt_conf_vldt_listener(nxt_conf_validation_t *vldt, nxt_str_t *name,
    nxt_conf_value_t *value)
{
    nxt_int_t       ret;
    nxt_str_t       str;
    nxt_sockaddr_t  *sa;

    if (nxt_slow_path(nxt_str_dup(vldt->pool, &str, name) == NULL)) {
        return NXT_ERROR;
    }

    sa = nxt_sockaddr_parse(vldt->pool, &str);
    if (nxt_slow_path(sa == NULL)) {
        return nxt_conf_vldt_error(vldt,
                                   "The listener address \"%V\" is invalid.",
                                   name);
    }

    ret = nxt_conf_vldt_type(vldt, name, value, NXT_CONF_VLDT_OBJECT);
    if (ret != NXT_OK) {
        return ret;
    }

    return nxt_conf_vldt_object(vldt, value, nxt_conf_vldt_listener_members);
}


src/nxt_conf_validation.c:1935:
static nxt_int_t
nxt_conf_vldt_proxy(nxt_conf_validation_t *vldt, nxt_conf_value_t *value,
    void *data)
{
    nxt_str_t       name;
    nxt_sockaddr_t  *sa;

    nxt_conf_get_string(value, &name);

    if (nxt_str_start(&name, "http://", 7)) {
        name.length -= 7;
        name.start += 7;

        sa = nxt_sockaddr_parse(vldt->pool, &name);
        if (sa != NULL) {
            return NXT_OK;
        }
    }

    return nxt_conf_vldt_error(vldt, "The \"proxy\" address is invalid \"%V\"",
                               &name);
}


src/nxt_conf_validation.c:3266:
static nxt_int_t
nxt_conf_vldt_server(nxt_conf_validation_t *vldt, nxt_str_t *name,
    nxt_conf_value_t *value)
{
    nxt_int_t       ret;
    nxt_sockaddr_t  *sa;

    ret = nxt_conf_vldt_type(vldt, name, value, NXT_CONF_VLDT_OBJECT);

    if (ret != NXT_OK) {
        return ret;
    }

    sa = nxt_sockaddr_parse(vldt->pool, name);

    if (sa == NULL) {
        return nxt_conf_vldt_error(vldt, "The \"%V\" is not valid "
                                   "server address.", name);
    }

    return nxt_conf_vldt_object(vldt, value,
                                nxt_conf_vldt_upstream_server_members);
}
```

`name` holds the string which will be written to the state file.

-  In `nxt_conf_vldt_proxy()` we explicitly reject unix sockets, so we can forget.  (One may wonder why not allow Unix sockets there, though...)

In `nxt_conf_vldt_listener()`, we dup it, and then call `nxt_sockaddr_parse()` with that dup, so `name` is unmodified.

In `nxt_conf_vldt_server()`, we pass `name` to `nxt_sockaddr_parse()`, so we corrupt it.

alejandro-colomar:
>     * In `nxt_conf_vldt_proxy()` we explicitly reject Unix sockets, so we can forget.  (One may wonder why not allow Unix sockets there, though...)

Heh, I didn't think of `http://unix:`.  That's actually supported.

```json
$ cat unix.json 
{
	"listeners": {
		"*:80": {
			"pass": "routes/u"
		},
		"*:81": {
			"pass": "routes/p"
		},
		"unix:@abstract": {
			"pass": "routes/a"
		}
	},

	"routes": {
		"u": [{
			"action": {
				"pass": "upstreams/u"
			}
		}],
		"p": [{
			"action": {
				"proxy": "http://unix:@abstract"
			}
		}],
		"a": [{
			"action": {
				"return": 302,
				"location": "/i/am/at/home/"
			}
		}]
	},

	"upstreams": {
		"u": {
			"servers": {
				"unix:@abstract": {}
			}
		}
	}
}
```

```sh
$ curl -v localhost:81
* Host localhost:81 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:81...
* connect to ::1 port 81 from ::1 port 58282 failed: Connection refused
*   Trying 127.0.0.1:81...
* Connected to localhost (127.0.0.1) port 81
> GET / HTTP/1.1
> Host: localhost:81
> User-Agent: curl/8.5.0
> Accept: */*
> 
< HTTP/1.1 302 Found
< Location: /i/am/at/home/
< Date: Fri, 02 Feb 2024 14:15:57 GMT
< Content-Length: 0
< Server: Unit/1.32.0
< 
* Connection #0 to host localhost left intact
```

ac000:
Heh, this is a total head fsck...

Correct me if I'm wrong.

```nxt_conf_vldt_listener()``` handles the `"listeners": {}` case and that is fine.

One or both of the other two functions handle the ```"upstreams": {}``` case and that is broken.

It's broken because in all three of those functions we call ```nxt_sockaddr_parse()```, and with the other two functions we aslo end up calling (although not directly) ```nxt_conf_vldt_listener()``` which means in that case we call ```nxt_sockaddr_parse()``` a second time.

So the question is, why do we call ``nxt_conf_vldt_listener()``` from the other functions code paths?...

**Comment by alejandro-colomar** [comment]
> Heh, this is a total head fsck...
> 
> Correct me if I'm wrong.
> 
> `nxt_conf_vldt_listener()` handles the `"listeners": {}` case and that is fine.

Yes.

> 
> One or both of the other two functions handle the `"upstreams": {}` case and that is broken.

Yes.

`nxt_conf_vldt_proxy()` validates `"action"` `"proxy"`.

and

`nxt_conf_vldt_server()` validates `"upstreams"` `"servers"`.

Both are broken in master.  They are both fixed by the current version of this patch.  v1 would only fix `"upstreams"` `"servers"`.

See how they're both broken in master:

```json
$ sudo /opt/local/nginx/unit/master/sbin/unitd 
2024/02/02 16:09:42 [info] 13010#13010 unit 1.32.0 started
$ tee /dev/tty <unix.json | sudo unit ctl http PUT /config
{
	"listeners": {
		"*:80": {
			"pass": "routes/u"
		},
		"*:81": {
			"pass": "routes/p"
		},
		"unix:@abstract": {
			"pass": "routes/a"
		}
	},

	"routes": {
		"u": [{
			"action": {
				"pass": "upstreams/u"
			}
		}],
		"p": [{
			"action": {
				"proxy": "http://unix:@abstract"
			}
		}],
		"a": [{
			"action": {
				"return": 302,
				"location": "/i/am/at/home/"
			}
		}]
	},

	"upstreams": {
		"u": {
			"servers": {
				"unix:@abstract": {}
			}
		}
	}
}
{
}
```
```
$ sudo unit ctl http GET /config </dev/null |& grep unix
		"unix:@abstract": {
					"proxy": "http://unix:\u0000abstract"
				"unix:\u0000abstract": {}
```

> 
> It's broken because in all three of those functions we call `nxt_sockaddr_parse()`, and with the other two functions we aslo end up calling (although not directly) `nxt_conf_vldt_listener()` which means in that case we call `nxt_sockaddr_parse()` a second time.

No; this is not correct.

In all three cases the situation is similar:

-  We receive a new config in the control socket.
-  We call `nxt_conf_vldt_***()` to validate the config.
   -  `nxt_conf_vldt_listener()` to validate the `"listeners"`.
   -  `nxt_conf_vldt_proxy()` to validate `"action"` `"proxy"`.
   -  `nxt_conf_vldt_server()` to validate `"upstreams"` `"servers"`.
   Each of those calls do call `nxt_sockaddr_parse()` with the corresponding string, exactly once.  The difference being that `***_listener()` makes a copy of the string before modifying it, but the others don't make a dup, and so modify the `name`.
   But mind that these functions only called `nxt_sockaddr_parse()` to see if it reports any error; the sockaddr that it allocates is discarded:

   ```c
   $ grepc -tfd nxt_conf_vldt_listener . | grep -- '\<sa\>'
       nxt_sockaddr_t  *sa;
       sa = nxt_sockaddr_parse(vldt->pool, &str);
       if (nxt_slow_path(sa == NULL)) {
   ```
   After all this returns, we can consider the new config to be valid.

-  After validating the configuration, we do the actual reconfiguration.  Here's where the sockets are `bind(2)`ed.  To do that, we need to call `nxt_sockaddr_parse()` again (remember we discarded `sa`).  This second call is responsible for one of the bugs: passing the kernel the terminating `'\0'`.

-  After validating the entire config and successfully reconfiguring, we write the new configuration to the state file.
   Here's where the second bug is exposed: since the `vldt` calls corrupsted the `name` string, we write it corrupted to the state file.

> 
> So the question is, why do we call ``nxt_conf_vldt_listener()``` from the other functions code paths?...

No, that doesn't happen.

ac000:
Thanks! That makes things much clearer!

So hopefully the below statements are correct...

In the listeners case we take a copy of the socket name and can munge it all we want. In the other two cases we're operating on the original name and fubar it...

So we really just need to strdup the socket name in the server and proxy cases.

alejandro-colomar:
This might be useful:

```c
$ grepc -htu 'nxt_sockaddr_parse\b' . | grep -o -e '^nxt_[^(]*(' -e '^[{}]$' -e '.*nxt_sockaddr_parse(.*'
nxt_runtime_conf_init(
{
    sa = nxt_sockaddr_parse(rt->mem_pool, &control);
}
nxt_http_proxy_init(
{
        sa = nxt_sockaddr_parse(mp, &name);
}
nxt_upstream_round_robin_create(
{
        sa = nxt_sockaddr_parse(mp, &name);
}
nxt_router_socket_conf(
{
    sa = nxt_sockaddr_parse(tmcf->mem_pool, name);
}
nxt_conf_vldt_listener(
{
    sa = nxt_sockaddr_parse(vldt->pool, &str);
}
nxt_conf_vldt_proxy(
{
        sa = nxt_sockaddr_parse(vldt->pool, &name);
}
nxt_conf_vldt_server(
{
    sa = nxt_sockaddr_parse(vldt->pool, name);
}
```

```c
$ grepc -htu 'nxt_runtime_conf_init\b' . | grep -o -e '^nxt_[^(]*(' -e '^[{}]$' -e '.*nxt_runtime_conf_init(.*'
nxt_runtime_create(
{
    if (nxt_runtime_conf_init(task, rt) != NXT_OK) {
}

$ grepc -htu 'nxt_runtime_create\b' . | grep -o -e '^\(nxt_\|main\)[^(]*(' -e '^[{}]$' -e '.*nxt_runtime_create(.*'
main(
{
    ret = nxt_runtime_create(&nxt_main_task);
}
```
```c
$ grepc -htu 'nxt_http_proxy_init\b' . | grep -o -e '^nxt_[^(]*(' -e '^[{}]$' -e '.*nxt_http_proxy_init(.*'
nxt_http_action_init(
{
        return nxt_http_proxy_init(mp, action, &acf);
}

$ grepc -htu 'nxt_http_action_init\b' . | grep -o -e '^\(nxt_\|main\)[^(]*(' -e '^[{}]$' -e '.*nxt_http_action_init(.*'
nxt_http_route_match_create(
{
    ret = nxt_http_action_init(task, tmcf, action_conf, &match->action);
}
nxt_http_static_init(
{
        return nxt_http_action_init(task, tmcf, acf->fallback,
}

$ grepc -htu 'nxt_http_route_match_create\b' . | grep -o -e '^\(nxt_\|main\)[^(]*(' -e '^[{}]$' -e '.*nxt_http_route_match_create(.*'
nxt_http_route_create(
{
        match = nxt_http_route_match_create(task, tmcf, value);
}

$ grepc -htu 'nxt_http_route_create\b' . | grep -o -e '^\(nxt_\|main\)[^(]*(' -e '^[{}]$' -e '.*nxt_http_route_create(.*'
nxt_http_routes_create(
{
            route = nxt_http_route_create(task, tmcf, route_conf);
        route = nxt_http_route_create(task, tmcf, routes_conf);
}

$ grepc -htu 'nxt_http_routes_create\b' . | grep -o -e '^\(nxt_\|main\)[^(]*(' -e '^[{}]$' -e '.*nxt_http_routes_create(.*'
nxt_router_conf_create(
{
        routes = nxt_http_routes_create(task, tmcf, conf);
}

$ grepc -htu 'nxt_router_conf_create\b' . | grep -o -e '^\(nxt_\|main\)[^(]*(' -e '^[{}]$' -e '.*nxt_router_conf_create(.*'
nxt_router_conf_data_handler(
{
    ret = nxt_router_conf_create(task, tmcf, p, nxt_pointer_to(p, size));
}
```
```c
$ grepc -htu 'nxt_upstream_round_robin_create\b' . | grep -o -e '^nxt_[^(]*(' -e '^[{}]$' -e '.*nxt_upstream_round_robin_create(.*'
nxt_upstreams_create(
{
        ret = nxt_upstream_round_robin_create(task, tmcf, upcf,
}

$ grepc -htu 'nxt_upstreams_create\b' . | grep -o -e '^\(nxt_\|main\)[^(]*(' -e '^[{}]$' -e '.*nxt_upstreams_create(.*'
nxt_router_conf_create(
{
    ret = nxt_upstreams_create(task, tmcf, root);
}
```
```c
$ grepc -htu 'nxt_router_socket_conf\b' . | grep -o -e '^nxt_[^(]*(' -e '^[{}]$' -e '.*nxt_router_socket_conf(.*'
nxt_router_conf_create(
{
            skcf = nxt_router_socket_conf(task, tmcf, &name);
}
```
The `vldt` functions aren't as easy, as they are used via function pointers:
```c
$ grep -rn '=.nxt_conf_vldt_\(listener\|proxy\|server\)\>'
src/nxt_conf_validation.c:280:        .u.object   = nxt_conf_vldt_listener,
src/nxt_conf_validation.c:781:        .validator  = nxt_conf_vldt_proxy,
src/nxt_conf_validation.c:1350:        .u.object   = nxt_conf_vldt_server,
```
Those are stored in a
```c
$ grepc nxt_conf_vldt_object_t .
./src/nxt_conf_validation.c:typedef struct nxt_conf_vldt_object_s  nxt_conf_vldt_object_t;
$ grepc nxt_conf_vldt_object_s .
./src/nxt_conf_validation.c:struct nxt_conf_vldt_object_s {
    nxt_str_t                     name;
    nxt_conf_vldt_type_t          type:32;
    nxt_conf_vldt_flags_t         flags:32;
    nxt_conf_vldt_handler_t       validator;

    union {
        nxt_conf_vldt_object_t    *members;
        nxt_conf_vldt_object_t    *next;
        nxt_conf_vldt_member_t    object;
        nxt_conf_vldt_element_t   array;
        const char                *string;
    } u;
};
```
Here's where `.u.object` seems to be called:
```c
$ grepc -rtuf nxt_conf_vldt_member_t .
./src/nxt_conf_validation.c:static nxt_int_t
nxt_conf_vldt_object_iterator(nxt_conf_validation_t *vldt,
    nxt_conf_value_t *value, void *data)
{
    uint32_t                index;
    nxt_int_t               ret;
    nxt_str_t               name;
    nxt_conf_value_t        *member;
    nxt_conf_vldt_member_t  validator;

    validator = (nxt_conf_vldt_member_t) data;
    index = 0;

    for ( ;; ) {
        member = nxt_conf_next_object_member(value, &name, &index);

        if (member == NULL) {
            return NXT_OK;
        }

        ret = validator(vldt, &name, member);

        if (ret != NXT_OK) {
            return ret;
        }
    }
}
```
`.validator` is obscure enough that I couldn't find it.

alejandro-colomar:
> Thanks! That makes things much clearer!
> 
> So hopefully the below statements are correct...
> 
> In the listeners case we take a copy of the socket name and can munge it all we want. In the other two cases we're operating on the original name and fubar it...
> 
> So we really just need to strdup the socket name in the server and proxy cases.

Yup!  That's entirely correct.  :)

alejandro-colomar:
And let's investigate the memory pools with which `nxt_sockaddr_parse()` is being called:

```c
$ grepc -tfd nxt_runtime_conf_init . | grep -e '\*rt' -e '^nxt_' -e '^[{}]' -e nxt_sockaddr_parse
nxt_runtime_conf_init(nxt_task_t *task, nxt_runtime_t *rt)
{
    sa = nxt_sockaddr_parse(rt->mem_pool, &control);
}
```

```c
$ grepc -tfd nxt_http_proxy_init . | grep -e '\*rt' -e '^nxt_' -e '^[{}]' -e nxt_sockaddr_parse
nxt_http_proxy_init(nxt_mp_t *mp, nxt_http_action_t *action,
{
        sa = nxt_sockaddr_parse(mp, &name);
}
alx@debian:~/src/nginx/unit/master$ grepc -tfd nxt_http_action_init . | grep -e '\*rt' -e '^nxt_' -e '^[{}]' -e nxt_http_proxy_init -e 'mp =' -e 'rtcf ='
nxt_http_action_init(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
{
    nxt_router_conf_t       *rtcf;
    rtcf = tmcf->router_conf;
    mp = rtcf->mem_pool;
        return nxt_http_proxy_init(mp, action, &acf);
}
```

```c
$ grepc -tfd nxt_upstream_round_robin_create . | grep -e '\*rt' -e '^nxt_' -e '^[{}]' -e nxt_sockaddr_parse -e 'mp ='
nxt_upstream_round_robin_create(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
{
    mp = tmcf->router_conf->mem_pool;
        sa = nxt_sockaddr_parse(mp, &name);
}
```

The above seem to be using a long-lived memory pools.

The below seem to be using short-lived memory pools:

```c
$ grepc -tfd nxt_router_socket_conf . | grep -e '\*rt' -e '^nxt_' -e '^[{}]' -e nxt_sockaddr_parse
nxt_router_socket_conf(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
{
    sa = nxt_sockaddr_parse(tmcf->mem_pool, name);
}
```

```c
nxt_conf_vldt_listener(
{
    sa = nxt_sockaddr_parse(vldt->pool, &str);
}
nxt_conf_vldt_proxy(
{
        sa = nxt_sockaddr_parse(vldt->pool, &name);
}
nxt_conf_vldt_server(
{
    sa = nxt_sockaddr_parse(vldt->pool, name);
}
```

(but this is pure guess.)

ac000:
Hmm, yeah, you might be right...

alejandro-colomar:
v4 changes:

-  Dup the strings in validation code, to avoid wasting memory from long-lived memory pools.
-  Simplify some code with a new function: `nxt_conf_get_string_dup()`

```
$ git range-diff master gh/unix unix 
1:  91e2a8c1 < -:  -------- Configuration: Store abstract sockets with @ internally
-:  -------- > 1:  a4ffc04f Configuration: Add nxt_conf_get_string_dup()
-:  -------- > 2:  ccdfd085 Simplify, by calling nxt_conf_get_string_dup()
-:  -------- > 3:  673a7328 Configuration: Store abstract Unix sockets with '@' internally
```

alejandro-colomar:
v4b changes:

-  Copyright

```
$ git range-diff master gh/unix unix 
1:  a4ffc04f = 1:  a4ffc04f Configuration: Add nxt_conf_get_string_dup()
2:  ccdfd085 = 2:  ccdfd085 Simplify, by calling nxt_conf_get_string_dup()
3:  673a7328 ! 3:  f5fe3076 Configuration: Store abstract Unix sockets with '@' internally
    @@ Commit message
     
    +@@
    + /*
    +  * Copyright (C) Valentin V. Bartenev
    +  * Copyright (C) NGINX, Inc.
    ++ * Copyright 2024, Alejandro Colomar <alx@kernel.org>
    +  */
    + 
    + #include <nxt_main.h>
     @@ src/nxt_conf_validation.c: nxt_conf_vldt_listener(nxt_conf_validation_t *vldt, nxt_str_t *name,
          nxt_conf_value_t *value)
      {
```

src/nxt_conf_validation.c
```diff
@@ -1683,14 +1684,9 @@ nxt_conf_vldt_listener(nxt_conf_validation_t *vldt, nxt_str_t *name,
...
     nxt_int_t       ret;
-    nxt_str_t       str;
     nxt_sockaddr_t  *sa;
 
-    if (nxt_slow_path(nxt_str_dup(vldt->pool, &str, name) == NULL)) {
-        return NXT_ERROR;
-    }
-
-    sa = nxt_sockaddr_parse(vldt->pool, &str);
+    sa = nxt_sockaddr_parse(vldt->pool, name);
```
ac000, line 1689: I take it we don't need to strdup it here as this is called from nxt_conf_vldt_object_iterator() where it's already duped?

src/nxt_conf_validation.c
```diff
@@ -1936,10 +1932,13 @@ static nxt_int_t
...
 {
-    nxt_str_t       name;
+    nxt_str_t       name, *ret;
     nxt_sockaddr_t  *sa;
 
-    nxt_conf_get_string(value, &name);
+    ret = nxt_conf_get_string_dup(value, vldt->pool, &name);
+    if (nxt_slow_path(ret == NULL)) {
+        return NXT_ERROR;
+    }
```
ac000, line 1942: I take it this doesn't get called from nxt_conf_vldt_object_iterator()?

src/nxt_conf_validation.c
```diff
@@ -2913,13 +2912,15 @@ nxt_conf_vldt_object_iterator(nxt_conf_validation_t *vldt,
...
         if (member == NULL) {
             return NXT_OK;
         }
 
-        ret = validator(vldt, &name, member);
+        if (nxt_slow_path(nxt_str_dup(vldt->pool, &str, &name) == NULL)) {
+            return NXT_ERROR;
+        }
 
+        ret = validator(vldt, &str, member);
```
ac000, line 2923: I suppose this will be duping all object members...

src/nxt_conf_validation.c
alejandro-colomar: Right.

(rant)
Why can't github show it as a hunk so that I know in which function this is happening?
This is
```diff
@@ -1683,14 +1684,9 @@ nxt_conf_vldt_listener(nxt_conf_validation_t *vldt, nxt_str_t *nam
e,
```
in git
(/rant)

src/nxt_conf_validation.c
alejandro-colomar: Right.

Both `"listeners"` and `"servers"` have the Unix socket in the object name, which results in a validator stored in `.u.object`, buit `"proxy"` has the Unix socket in the string, and so it results in a validator stored in `.validator`.

(recall from yesterday:

```c
$ grep -rn '=.nxt_conf_vldt_\(listener\|proxy\|server\)\>'
src/nxt_conf_validation.c:280:        .u.object   = nxt_conf_vldt_listener,
src/nxt_conf_validation.c:781:        .validator  = nxt_conf_vldt_proxy,
src/nxt_conf_validation.c:1350:        .u.object   = nxt_conf_vldt_server,
```
)

`.u.object` is a `nxt_conf_vldt_member_t`, which I found to be called here:

```c
$ grepc -tuf nxt_conf_vldt_member_t . | grep -e ^nxt_ -e '^[{}]' -e validator -e name
nxt_conf_vldt_object_iterator(nxt_conf_validation_t *vldt,
{
    nxt_str_t               name, str;
    nxt_conf_vldt_member_t  validator;
    validator = (nxt_conf_vldt_member_t) data;
        member = nxt_conf_next_object_member(value, &name, &index);
        if (nxt_slow_path(nxt_str_dup(vldt->pool, &str, &name) == NULL)) {
        ret = validator(vldt, &str, member);
}
```

`.validator` I found it to be called here:

```c
$ grepc -tuf '>validator\(' . | grep -e ^nxt_ -e '^[{}]' -e '>validator('
nxt_conf_vldt_object(nxt_conf_validation_t *vldt, nxt_conf_value_t *value,
{
                ret = vals->validator(vldt, member, vals->u.members);
}
```

`.validator()` isn't being passed the name, which is why it needs to get it inside the call.

src/nxt_conf_validation.c
alejandro-colomar: Hunk:

```diff
@@ -2913,13 +2912,15 @@ nxt_conf_vldt_object_iterator(nxt_conf_validation_t *vldt,
```

src/nxt_conf_validation.c
alejandro-colomar: Context:

```diff
@@ -1936,10 +1932,13 @@ static nxt_int_t
 nxt_conf_vldt_proxy(nxt_conf_validation_t *vldt, nxt_conf_value_t *value,
```

src/nxt_conf_validation.c
alejandro-colomar: I encourage you to test memory usage before and after this patch, to make sure it's as we expect.

ac000: My feeling (based on the past 18 months) is this is going to be a controversial change (tbh I'm not even completely fond of it).

For the best chances of this being accepted I would not do this change,, but rather just do the strdups in the required places to fix this issue... yes, may be a _little_ extra code, but shouldn't be too bad with your new function...

alejandro-colomar:
v5 changes:

-  Reduce the amount of dup'd strings, at the expense of slightly more code.  [@ac000 ]

```
$ git range-diff master gh/unix unix 
1:  a4ffc04f = 1:  a4ffc04f Configuration: Add nxt_conf_get_string_dup()
2:  ccdfd085 = 2:  ccdfd085 Simplify, by calling nxt_conf_get_string_dup()
3:  f5fe3076 ! 3:  5269386a Configuration: Store abstract Unix sockets with '@' internally
    @@ Commit message
         state file, storing user input pristine, and we only transform the '@'
         in temporary strings.
     
    -    To reduce the places where we dup strings, dup the name in
    -    nxt_conf_vldt_object(), so that every object validator gets a dup'd
    -    name.  While this may seem an unnecessary waste of memory, that memory
    -    pool will live only for a short time, during reconfiguration, so that's
    -    fine.
    -
         This commit fixes another bug, where we try to connect to abstract
         sockets with a trailing '\0' in their name due to calling twice
         nxt_sockaddr_parse() on the same string.  By calling that function only
    @@ src/nxt_conf_validation.c
       */
      
      #include <nxt_main.h>
    -@@ src/nxt_conf_validation.c: nxt_conf_vldt_listener(nxt_conf_validation_t *vldt, nxt_str_t *name,
    -     nxt_conf_value_t *value)
    - {
    -     nxt_int_t       ret;
    --    nxt_str_t       str;
    -     nxt_sockaddr_t  *sa;
    - 
    --    if (nxt_slow_path(nxt_str_dup(vldt->pool, &str, name) == NULL)) {
    --        return NXT_ERROR;
    --    }
    --
    --    sa = nxt_sockaddr_parse(vldt->pool, &str);
    -+    sa = nxt_sockaddr_parse(vldt->pool, name);
    -     if (nxt_slow_path(sa == NULL)) {
    -         return nxt_conf_vldt_error(vldt,
    -                                    "The listener address \"%V\" is invalid.",
     @@ src/nxt_conf_validation.c: static nxt_int_t
      nxt_conf_vldt_proxy(nxt_conf_validation_t *vldt, nxt_conf_value_t *value,
          void *data)
    @@ src/nxt_conf_validation.c: nxt_conf_vldt_object_iterator(nxt_conf_validation_t *
                  return NXT_OK;
              }
      
    --        ret = validator(vldt, &name, member);
    -+        if (nxt_slow_path(nxt_str_dup(vldt->pool, &str, &name) == NULL)) {
    -+            return NXT_ERROR;
    -+        }
    - 
    -+        ret = validator(vldt, &str, member);
    +         ret = validator(vldt, &name, member);
    +-
              if (ret != NXT_OK) {
                  return ret;
              }
    +@@ src/nxt_conf_validation.c: nxt_conf_vldt_server(nxt_conf_validation_t *vldt, nxt_str_t *name,
    +     nxt_conf_value_t *value)
    + {
    +     nxt_int_t       ret;
    ++    nxt_str_t       str;
    +     nxt_sockaddr_t  *sa;
    + 
    +     ret = nxt_conf_vldt_type(vldt, name, value, NXT_CONF_VLDT_OBJECT);
    +-
    +     if (ret != NXT_OK) {
    +         return ret;
    +     }
    + 
    +-    sa = nxt_sockaddr_parse(vldt->pool, name);
    ++    if (nxt_slow_path(nxt_str_dup(vldt->pool, &str, name) == NULL)) {
    ++        return NXT_ERROR;
    ++    }
    + 
    ++    sa = nxt_sockaddr_parse(vldt->pool, &str);
    +     if (sa == NULL) {
    +         return nxt_conf_vldt_error(vldt, "The \"%V\" is not valid "
    +                                    "server address.", name);
```

hongzhidao:
Hi,
Thanks for the catch.
It looks like the issue is `addr` in `nxt_sockaddr_unix_parse()` should be constant but was modified if it's an abstract socket.
It means it makes sense to only touch abstract socket. Take a look at the commit. https://github.com/nginx/unit/commit/e2aec6686a4d2cf1ddc017b50dbb39baefd2d425
It dups the option `name` for the listeners, I feel we only need to touch the abstract socket instead, and abstract socket configuration is not as common as other sockets.

Here's a draft fixing.
```
diff -r aa207ced9712 src/nxt_sockaddr.c
--- a/src/nxt_sockaddr.c        Thu Aug 19 18:17:12 2021 +0300
+++ b/src/nxt_sockaddr.c        Mon Feb 05 13:04:31 2024 +0800
@@ -693,7 +693,18 @@
      *   (Null bytes in the name have no special significance.)
      */
     if (path[0] == '@') {
-        path[0] = '\0';
+        u_char  *tmp;
+
+        tmp = nxt_mp_nget(mp, length);
+        if (nxt_slow_path(tmp == NULL)) {
+            return NULL;
+        }
+
+        nxt_memcpy(tmp, path, length);
+        tmp[0] = '\0';
+
+        path = tmp;
+
         socklen--;
     }
```
Welcome to suggest.

ac000:
While this is certainly a simpler fix, I'm not overly fond of it as, as Alex says, we really need to do this duplication in the validation case.

So having mulled it over for a few minutes, I think I _do_ prefer Alex's latest patch.

hongzhidao:
> Unfortunately, that would be a performance regression

Could you be more specific?

hongzhidao:
Sorry, I can't understand where the performance regression is from your comment.
Now for the configuration options, only the abstract socket isn't handled well. I guess the reason is that `nxt_sockaddr.c` was written before the design of support JSON format configuration.
Unit is designed with a high-performance, low-memory oss, I'd suggest keeping it.
Duplicating all the object names looks unnecessary.

ac000:
> Sorry, I can't understand where the performance regression is from your comment. Now for the configuration options, only the

I think the point is that with the change in nxt_sockaddr_parse(), you'd be duplicating the abstract names for both the validation case and for the non-validation case.

This patch just extends what currently happens in the listener validation to the other couple of places where abstract sockets may be used.

> Duplicating all the object names looks unnecessary.

Not _all_ the objects, just the socket name objects.

If you're concerned about the duplicating of non-abstract socket name objects, I guess we could add a simple guard around the duplication to only do it with abstract sockets in the config validation.

ac000:
I guess ultimately it would have been better to _not_ have a _parsing_ function that _mutates_ its input...

src/nxt_conf_validation.c
hongzhidao: > My feeling (based on the past 18 months) is this is going to be a controversial change (tbh I'm not even completely fond of it).
> 
> For the best chances of this being accepted I would not do this change,, but rather just do the strdups in the required places to fix this issue... yes, may be a _little_ extra code, but shouldn't be too bad with your new function...

I see. My mistake that duplicating will be applied. Thanks for correcting me.
This new way makes sense.

ac000:
Now that Zhidao seems OK with the current patches, can I perhaps _suggest_ a slight tweaking of the last commit subject? i.e

```
Configuration: Store abstract Unix sockets with '@' internally
```

For the subject, could I possibly suggest something like

```
Configuration: Don't corrupt abstract socket names in some cases
```

Which I feel cuts to the heart of the matter more.

Also 

> nxt_sockaddr_parse() on the same string.  By calling that function only

technically it's `nxt_sockaddr_unix_parse()` that's doing the damage...

hongzhidao:
LGTM.

alejandro-colomar:
> Now that Zhidao seems OK with the current patches, can I perhaps _suggest_ a slight tweaking of the last commit subject? i.e
> 
> ```
> Configuration: Store abstract Unix sockets with '@' internally
> ```
> 
> For the subject, could I possibly suggest something like
> 
> ```
> Configuration: Don't corrupt abstract socket names in some cases
> ```
> 
> Which I feel cuts to the heart of the matter more.
> 
> Also
> 
> > nxt_sockaddr_parse() on the same string.  By calling that function only
> 
> technically it's `nxt_sockaddr_unix_parse()` that's doing the damage...

Indeed.  I had been thinking these days that the subject only accounts for one of the two bugs being fixed here, but didn't come up with anything better.  I'll take your suggestion.

alejandro-colomar:
v5b changes:

-  Tweak 3rd commit's subject line.  [@ac000 ]

```
$ git range-diff ngx/master gh/unix unix 
1:  a4ffc04f = 1:  a4ffc04f Configuration: Add nxt_conf_get_string_dup()
2:  ccdfd085 = 2:  ccdfd085 Simplify, by calling nxt_conf_get_string_dup()
3:  5269386a ! 3:  ae25fb55 Configuration: Store abstract Unix sockets with '@' internally
    @@ Metadata
     
      ## Commit message ##
    -    Configuration: Store abstract Unix sockets with '@' internally
    +    Configuration: Don't corrupt abstract socket names
     
         The commit that added support for Unix sockets accepts abstract sockets
         using '@' in the config, but we stored it internally using '\0'.
```

src/nxt_conf_validation.c
```diff
@@ -2904,7 +2908,7 @@ nxt_conf_vldt_object_iterator(nxt_conf_validation_t *vldt,
 {
     uint32_t                index;
     nxt_int_t               ret;
-    nxt_str_t               name;
+    nxt_str_t               name, str;
```
ac000, line 2911: Ah, this hunk should be dropped ;)

str is not used...

src/nxt_conf_validation.c
```diff
@@ -2913,13 +2917,11 @@ nxt_conf_vldt_object_iterator(nxt_conf_validation_t *vldt,
...
-
         if (member == NULL) {
             return NXT_OK;
         }
 
         ret = validator(vldt, &name, member);
-
         if (ret != NXT_OK) {
             return ret;
         }
```
ac000, line 2927: And this hunk should probably go also, it's harmless but it's also not really related now...

src/nxt_conf_validation.c
alejandro-colomar: Hmmm, true!

src/nxt_conf_validation.c
alejandro-colomar: I noticed that one, and I liked it.  It's a small reminder of where the validator is called, and where this could have been fixed instead.  We may need that in the future if other functions start needing writable strings.  I'll keep this hunk.

ac000 (approved):
LGTM

```
```

for all three patches.

BTW, I've been taken peoples review approvals as explicit Reviewed-by's...

alejandro-colomar:
Hi Andrew!

On Mon, Feb 05, 2024 at 09:17:37AM -0800, Andrew Clayton wrote:
> @ac000 approved this pull request.
> 
> LGTM
> 
> ```
> Reviewed-by: Andrew Clayton ***@***.***>
> ```
> 
> for all three patches.

Thanks for the picky review!  I wouldn't expect/want less  ;)

> BTW, I've been taken peoples review approvals as explicit Reviewed-by's...

I don't understand this sentence.  Can you please reword/clarify?

Have a lovely day,
Alex

-- 
Looking for a remote C programming job at the moment.

alejandro-colomar:
v5d changes:

-  Review + Cc tags [@ac000 ]

```
$ git range-diff ngx/master gh/unix unix 
1:  a4ffc04f ! 1:  ecd57392 Configuration: Add nxt_conf_get_string_dup()
    @@ Commit message
         This function is like nxt_conf_get_string(), but creates a new copy,
         so that it can be modified without corrupting the configuration string.
     
    -    Cc: Andrew Clayton <a.clayton@nginx.com>
    +    Cc: Zhidao Hong <z.hong@f5.com>
     
      ## src/nxt_conf.c ##
2:  ccdfd085 ! 2:  bb376c68 Simplify, by calling nxt_conf_get_string_dup()
    @@ Commit message
     
         Refactor.
     
    -    Cc: Andrew Clayton <a.clayton@nginx.com>
    +    Cc: Zhidao Hong <z.hong@f5.com>
     
      ## src/nxt_http_static.c ##
3:  c1bdd052 ! 3:  46cef09f Configuration: Don't corrupt abstract socket names
    @@ Commit message
         Fixes: d8e0768a5bae ("Fixed support for abstract Unix sockets.")
         Fixes: e2aec6686a4d ("Storing abstract sockets with @ internally.")
         Link: <https://github.com/nginx/unit/pull/1108>
    -    Cc: Andrew Clayton <a.clayton@nginx.com>
         Cc: Liam Crilly <liam.crilly@nginx.com>
    +    Cc: Zhidao Hong <z.hong@f5.com>
     
```

ac000:
Thanks Alex!

PR #1109, opened 2024-01-31
Fixed some issues found by tests with enabled UndefinedBehaviorSanitizer

andrey-zelenkov:


src/nxt_http_js.c
```diff
@@ -165,8 +167,9 @@ nxt_http_js_ext_get_args(njs_vm_t *vm, njs_object_prop_t *prop,
 
     args = njs_value_arg(&val);
 
-    ret = njs_vm_query_string_parse(vm, r->args->start,
-                                    r->args->start + r->args->length, args);
+    length = r->args->length;
+    start = (length != 0) ? r->args->start : (u_char *) "";
```
ac000, line 171: length can't be < 0. so I'd make it (length > 0)

ac000:
For 505e879 ("Fixed undefined behaviour in left shift of int value")

```
```

ac000 (commented):
EDIT: See below comment...

~Are you saying that if _mmaps->size == 0_ then _mmaps->elts_ is NULL?~

~If so, I'd make the check _(mmaps->elts == NULL)_~

src/nxt_port_memory.c
```diff
@@ -454,6 +454,10 @@ nxt_port_mmap_get(nxt_task_t *task, nxt_port_mmaps_t *mmaps, nxt_chunk_id_t *c,
 
     nxt_thread_mutex_lock(&mmaps->mutex);
 
+    if (nxt_slow_path(mmaps->size == 0)) {
+        goto end;
+    }
```
ac000, line 459: Are you saying that if _mmaps->size == 0_ then _mmaps->elts_ is NULL?

If so, I'd make the check _(mmaps->elts == NULL)_

ac000:
For e059674 ("Initialize port_impl only when it is needed")

```
```

src/nxt_http_request.c
```diff
@@ -957,6 +957,11 @@ nxt_http_arguments_parse(nxt_http_request_t *r)
 
     r->args_decoded.start = dst_start;
 
+    if (nxt_slow_path(r->args->length == 0)) {
+        r->args_decoded.length = 0;
+        goto end;
+    }
+
     start = r->args->start;
     end = start + r->args->length;
```
ac000, line 967: Again, if this is because _r->args->start_ can be NULL, then I'd check for that.

src/nxt_http_route.c
```diff
@@ -2134,6 +2134,10 @@ nxt_http_route_pattern(nxt_http_request_t *r, nxt_http_route_pattern_t *pattern,
         return 0;
     }
 
+    if (nxt_slow_path(length == 0)) {
+        return 1;
+    }
+
```
ac000, line 2140: Again, if this is because _start_ can be NULL, then check for that.

The reason I'm saying this is because it's not that if _length == 0_ is the problem, it's that if the pointer you're applying it to is NULL.

This is all assuming these things can be NULL, are we sure of that?

src/nxt_http_js.c
ac000: But forget that, that's the wrong thing to check, we should check if _r->args->start_ is NULL.

Again, if it even can be?

src/nxt_http_route.c
ac000: However, in _this_ instance, I'd be willing to accept the above check given the commit message.

But I still think the NULL checks are more appropriate where I've mentioned elsewhere.

src/nxt_unit.c
```diff
@@ -3506,6 +3506,10 @@ nxt_unit_mmap_get(nxt_unit_ctx_t *ctx, nxt_unit_port_t *port,
 
     outgoing_size = lib->outgoing.size;
 
+    if (nxt_slow_path(outgoing_size == 0)) {
+        goto skip;
+    }
+
     mm_end = lib->outgoing.elts + outgoing_size;
```
ac000, line 3513: I assume this is because _lib->outgoing.elts_ can be NULL? Then I think it's better/clearer to check for that

src/nxt_http_js.c
andrey-zelenkov: > Again, if it even can be?

Yes, relevant test and Sanitizer's backtrace can be found in commit message to this patch. As an additional verification I can suggest to add assert in code:

```
diff --git a/src/nxt_http_js.c b/src/nxt_http_js.c
--- a/src/nxt_http_js.c
+++ b/src/nxt_http_js.c
@@ -165,6 +165,8 @@ nxt_http_js_ext_get_args(njs_vm_t *vm, n
 
     args = njs_value_arg(&val);
 
+    nxt_assert(r->args->start != NULL);
+
     ret = njs_vm_query_string_parse(vm, r->args->start,
                                     r->args->start + r->args->length, args);
 
```

And run test to make sure:

```
% pytest test/test_rewrite.py::test_rewrite_njs | grep "assertion"
2024/02/01 11:12:03.190 [alert] 85574#19767557 src/nxt_http_js.c:168 assertion failed: r->args->start != NULL
```

ac000: OK, well, I don't know what tool had found these issues and whether they were theoretical or it actually hit them... so good to know these are real...

andrey-zelenkov: tests are available here https://github.com/nginx/unit/tree/master/test and can be run as any standard pytests (i.e. `pytest test/test_static.py`). [UndefinedBehaviorSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) can be enabled just by compiling Unit with `-fsanitize=undefined` option.

alejandro-colomar: As a different approach, maybe we could just avoid `.start` being `NULL` at all, and set it to `(u_char *) ""` whenever length is 0.

That would need modifying `nxt_null_string()` and `nxt_str_null()`, and then fixing other code that is using NULL to use "".

The benefit of this approach is simplicity: less states a nxt_str_t can have.

Although it would need some research.

alejandro-colomar: Currently, there are two states of `nxt_str_t` that mean the same thing (or I haven't seen any code making a meaningful difference from them):

```c
{.length = 0, .start = NULL}
{.length = 0, .start = ""}
```

alejandro-colomar: Oh, now I remember, some strings are NULL due to some zero allocation, so it's not even set.  Forget my suggestion.  :)

src/nxt_http_request.c
alejandro-colomar: This results in less code:

```c
start = r->args->start ?: (u_char) "";
```

ac000: We'll short circuit out the for () loop the only question is do we want to (or not) do the bit of code between the for () loop and the end label in such situations?

alejandro-colomar: Short circuiting optimizes for the case where .length == 0, at the cost of adding some more code (possibly pessimizing the rest of the cases by one op or so).  I guess it depends on how likely each of the cases are.

ac000: Sorry I wasn't clear, the short circuiting is fine. But by not just going to `end:` when ->start is NULL means we'll execute some other stuff that we _may_ not want to...

Looking closer, assuming the `dst = dst_start` in the for loop initialiser happens, then we will execute

```c
r->args_decoded.length = dst - r->args_decoded.start;
```

Which will equate to `dst - dst` which should be `0` and

```c
    if (name_length != 0 || dst != dst_start) {
```

wil equate to false. So I think those are OK.

So yes, the ternary should also do the trick...

hongzhidao: > Again, if this is because _r->args->start_ can be NULL, then I'd check for that.

I can.

For example:
```
/foo: r->args->start is NULL
/foo?: r->args->start is not NULL
```

And for the configuration:
```
{
    "match": {
          "query": ""
    }
}
```
The pattern related to `query` is also NULL.
```
pattern->start = NULL;
```

So it looks like we don't need to touch `r->args_decoded` if there is no UB reported. Otherwise, not sure if we'll have to rethink the parrten->start to query "" value.

I'd would suggest fixing the case of `r->args_start` is NULL. For example:
```
if (nxt_slow_path(r->args->start == NULL)) {
    goto end;
}
```

alejandro-colomar: This reminds me of <https://github.com/mkirchner/linked-list-good-taste>.

ac000: Heh, I actually have that email from Linus saved off for whenever I need to refer to it...

I think what Alex is saying is that with his ternary suggestion the code just flows through and does the right thing ragardless if `r->args->start` is `NULL` or not.

ac000:
If you could add my `Reviewed-by` tags to the commits I've specified, then I can more easily see which things I've already reviewed.

andrey-zelenkov:
Rebased and added `Reviewed-by: Andrew Clayton <a.clayton@nginx.com>` where needed. Was not sure about review status of 27b51c8d so preferred to leave it unreviewed.

```
% git range-diff ca37cf1a...da0802eb
 ...
 1:  2a2e2bbb = 41:  27b51c8d Router: match when pattern and tested string are both zero length
 2:  3519c47f = 42:  58ef7c74 NJS: avoiding arithmetic ops with NULL pointer in r->args
 3:  890ba34b ! 43:  b706f9a0 Fixed undefined behaviour in left shift of int value
    @@ Commit message
     
         SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/nxt_random.c:151:31
     
    +
      ## src/nxt_random.c ##
     @@ src/nxt_random.c: nxt_random(nxt_random_t *r)
              nxt_random_stir(r);
 4:  88a17771 = 44:  6ae05623 Avoiding arithmetic ops with NULL pointer in nxt_http_arguments_parse
 5:  540257e4 = 45:  c104472d Avoiding arithmetic ops with NULL pointer in nxt_port_mmap_get
 6:  4fc83cb7 ! 46:  2d16d02a Initialize port_impl only when it is needed
    @@ Commit message
     
         Found by UndefinedBehaviorSanitizer.
     
    +
      ## src/nxt_unit.c ##
     @@ src/nxt_unit.c: nxt_unit_request_check_response_port(nxt_unit_request_info_t *req,
          pthread_mutex_lock(&lib->mutex);
 7:  ca37cf1a = 47:  da0802eb Avoiding arithmetic ops with NULL pointer in nxt_unit_mmap_get
 ```

File: src/nxt_http_js.c [hunk 6]
```diff
@@ -175,8 +176,8 @@ nxt_http_js_ext_get_args(njs_vm_t *vm, njs_object_prop_t *prop,
 
     args = njs_value_arg(&val);
 
-    ret = njs_vm_query_string_parse(vm, r->args->start,
-                                    r->args->start + r->args->length, args);
+    start = (r->args->start != NULL) ? r->args->start : (u_char *) "";
+    ret = njs_vm_query_string_parse(vm, start, start + r->args->length, args);
 
     if (ret == NJS_ERROR) {
         return NJS_ERROR;
```
ac000, line 183:
```
```

src/nxt_http_route.c
ac000: Hmm, looking at the code again, I think I'd check if start is NULL.

src/nxt_http_request.c
ac000: I'd go with Alex's suggestion and fix it like you did in `NJS: avoiding arithmetic ops with NULL pointer in r->args`

src/nxt_unit.c
```diff
@@ -3502,6 +3502,10 @@ nxt_unit_mmap_get(nxt_unit_ctx_t *ctx, nxt_unit_port_t *port,
 
     pthread_mutex_lock(&lib->outgoing.mutex);
 
+    if (nxt_slow_path(lib->outgoing.elts == NULL)) {
+        goto skip;
+    }
+
 retry:
 
     outgoing_size = lib->outgoing.size;
```
ac000, line 3511:
```
```

src/nxt_http_js.c
ac000: It would be good if GH showed what commit this belongs to...

It's for

```
NJS: avoiding arithmetic ops with NULL pointer in r->args
```

src/nxt_unit.c
ac000: Ditto... this is for

```
Avoiding arithmetic ops with NULL pointer in nxt_unit_mmap_get
```

andrey-zelenkov:
Rebased, two more `Reviewed-by`, check against NULL in ec5a29a1:

```
% git range-diff da0802eb...b82297c4
 -:  -------- >  1:  bca44630 .mailmap: Map Dylan's GitHub address
 -:  -------- >  2:  f2e64475 Wasm-wc: Register a new Wasm component model language module type
 -:  -------- >  3:  f0782722 Wasm-wc: Add core configuration data structure
 -:  -------- >  4:  20ada4b5 Wasm-wc: Core of initial Wasm component model language module support
 -:  -------- >  5:  a9345dd4 Add a .rustfmt.toml file
 -:  -------- >  6:  79c81772 Wasm-wc: Run src/lib.rs through rustfmt
 -:  -------- >  7:  ac3a54d6 Wasm-wc: Improve request buffer handling
 -:  -------- >  8:  98f808af Wasm-wc: Upgrade to wasmtime 17
 -:  -------- >  9:  60eb6c43 Wasm-wc: Allow to use the 'reactor' adaptor again
 -:  -------- > 10:  8d030139 Wasm-wc: Add Cargo.lock
 -:  -------- > 11:  07a0c9a3 Wasm-wc: Wire up the language module to the config system
 -:  -------- > 12:  da44dc00 Fix alignment of wasm options text in auto/help
 -:  -------- > 13:  4e6d7e87 Wasm-wc: Wire it up to the build system
 -:  -------- > 14:  7702293d Docker: Bump rust version to 1.76.0
 -:  -------- > 15:  1297f6f0 Docker: Add wasm-wasi-component to the wasm target
 -:  -------- > 16:  4c558697 Docker: Re-generate Dockerfile.wasm
 -:  -------- > 17:  7883acc6 Tests: Ruby hook tests unstable for version older 3.0
 -:  -------- > 18:  99da2f3c Tests: check for the AddressSanitizer flag during discovery
 -:  -------- > 19:  dbd9d25f Tests: skip some of TLS reconfiguration tests under AddressSanitizer
 -:  -------- > 20:  cabea47d Tests: renamed test_python_procman.py since it's not Python-specific
 -:  -------- > 21:  3f805bc6 Packages: added wasm-wasi-component module packaging for deb-based distros
 -:  -------- > 22:  7a640556 Packages: added wasm-wasi-component module packaging for rpm-based distros
 -:  -------- > 23:  7b13c306 Wasm-wc: Add nxt_unit.o as a dependency in the auto script
 -:  -------- > 24:  d54af163 Wasm-wc: Use the cargo build output as the make target dependency
 -:  -------- > 25:  2f3c7c2c Update third-party java components to their r -:  -------- > 20:  cabea47d Tests: renamed test_python_procman.py since i
t's not Python-specific
 -:  -------- > 21:  3f805bc6 Packages: added wasm-wasi-component module pa
ckaging for deb-based distros
 -:  -------- > 18:  99da2f3c Tests: check for the AddressSanitizer flag du
ring discovery
 -:  -------- > 19:  dbd9d25f Tests: skip some of TLS reconfiguration tests
 under AddressSanitizer
 -:  -------- > 20:  cabea47d Tests: renamed test_python_procman.py since i
t's not Python-specific
 -:  -------- > 21:  3f805bc6 Packages: added wasm-wasi-component module pa
ckaging for deb-based distros
 -:  -------- > 22:  7a640556 Packages: added wasm-wasi-component module pa
ckaging for rpm-based distros
 -:  -------- > 23:  7b13c306 Wasm-wc: Add nxt_unit.o as a dependency in th
e auto script
 -:  -------- > 24:  d54af163 Wasm-wc: Use the cargo build output as the ma
ke target dependency
 -:  -------- > 25:  2f3c7c2c Update third-party java components to their r
ecent versions
 1:  27b51c8d ! 26:  ec5a29a1 Router: match when pattern and tested string are both zero length
    @@ src/nxt_http_route.c: nxt_http_route_pattern(nxt_http_request_t *r, nxt_http_rou
              return 0;
          }
      
    -+    if (nxt_slow_path(length == 0)) {
    ++    if (nxt_slow_path(start == NULL)) {
     +        return 1;
     +    }
     +
 2:  58ef7c74 ! 27:  65a80df6 NJS: avoiding arithmetic ops with NULL pointer in r->args
    @@ Commit message
         Same fix was introduced in NJS:
         <http://hg.nginx.org/njs/rev/4fba78789fe4>
     
    +
      ## src/nxt_http_js.c ##
     @@ src/nxt_http_js.c: static njs_int_t
      nxt_http_js_ext_get_args(njs_vm_t *vm, njs_object_prop_t *prop,
 3:  b706f9a0 = 28:  658d5c0a Fixed undefined behaviour in left shift of int value
 4:  6ae05623 = 29:  589b50c1 Avoiding arithmetic ops with NULL pointer in nxt_http_arguments_parse
 5:  c104472d = 30:  7d10bd59 Avoiding arithmetic ops with NULL pointer in nxt_port_mmap_get
 6:  2d16d02a = 31:  4f374c45 Initialize port_impl only when it is needed
 7:  da0802eb ! 32:  b82297c4 Avoiding arithmetic ops with NULL pointer in nxt_unit_mmap_get
    @@ Commit message
     
         Found by UndefinedBehaviorSanitizer.
     
    +
      ## src/nxt_unit.c ##
     @@ src/nxt_unit.c: nxt_unit_mmap_get(nxt_unit_ctx_t *ctx, nxt_unit_port_t *port,
```

src/nxt_http_route.c
```diff
@@ -2134,6 +2134,10 @@ nxt_http_route_pattern(nxt_http_request_t *r, nxt_http_route_pattern_t *pattern,
         return 0;
     }
 
+    if (nxt_slow_path(start == NULL)) {
+        return 1;
+    }
+
     nxt_assert(pattern->u.pattern_slices != NULL);
 
     pattern_slices = pattern->u.pattern_slices;
```
ac000, line 2143: Looks good.

For

```
Router: match when pattern and tested string are both zero length 
```
```
```

src/nxt_port_memory.c
```diff
@@ -454,6 +454,10 @@ nxt_port_mmap_get(nxt_task_t *task, nxt_port_mmaps_t *mmaps, nxt_chunk_id_t *c,
 
     nxt_thread_mutex_lock(&mmaps->mutex);
 
+    if (nxt_slow_path(mmaps->elts == NULL)) {
+        goto end;
+    }
+
     end_port_mmap = mmaps->elts + mmaps->size;
 
     for (port_mmap = mmaps->elts;
```
ac000, line 463: OK, looks sufficiently like nxt_unit_mmap_get()

```
Avoiding arithmetic ops with NULL pointer in nxt_port_mmap_get
```
```
```

andrey-zelenkov:
```
% git range-diff b82297c4...8eb3653f
1:  ec5a29a1 ! 1:  f80edc37 Router: match when pattern and tested string are both zero length
    @@ Commit message
     
         SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/nxt_http_route.c:2141:17
     
    +
      ## src/nxt_http_route.c ##
     @@ src/nxt_http_route.c: nxt_http_route_pattern(nxt_http_request_t *r, nxt_http_route_pattern_t *pattern,
              return 0;
2:  65a80df6 = 2:  592fe518 NJS: avoiding arithmetic ops with NULL pointer in r->args
3:  658d5c0a = 3:  915fd982 Fixed undefined behaviour in left shift of int value
4:  589b50c1 = 4:  1dafe541 Avoiding arithmetic ops with NULL pointer in nxt_http_arguments_parse
5:  7d10bd59 ! 5:  213864c3 Avoiding arithmetic ops with NULL pointer in nxt_port_mmap_get
    @@ Commit message
         Can be reproduced by test/test_settings.py::test_settings_send_timeout
         with enabled UndefinedBehaviorSanitizer.
     
    +
      ## src/nxt_port_memory.c ##
     @@ src/nxt_port_memory.c: nxt_port_mmap_get(nxt_task_t *task, nxt_port_mmaps_t *mmaps, nxt_chunk_id_t *c,
      
6:  4f374c45 = 6:  e6571b8e Initialize port_impl only when it is needed
7:  b82297c4 = 7:  8eb3653f Avoiding arithmetic ops with NULL pointer in nxt_unit_mmap_get
```

src/nxt_http_request.c
```diff
@@ -946,6 +946,10 @@ nxt_http_arguments_parse(nxt_http_request_t *r)
         return NULL;
     }
 
+    if (nxt_slow_path(r->args->start == NULL)) {
+        goto end;
+    }
+
     hash = NXT_HTTP_FIELD_HASH_INIT;
     name = NULL;
     name_length = 0;
```
ac000, line 955: For

```
Avoiding arithmetic ops with NULL pointer in nxt_http_arguments_parse
```

```
```

andrey-zelenkov:
```
% git range-diff 8eb3653f...887fa467
 -:  -------- >  1:  e2cab032 Remove debug from builds and tests
 -:  -------- >  2:  faa7e792 Packages: Pass CFLAGS to compile wasm modules on all packaging targets
 1:  f80edc37 =  3:  2dfba775 Router: match when pattern and tested string are both zero length
 2:  592fe518 =  4:  edb69d54 NJS: avoiding arithmetic ops with NULL pointer in r->args
 3:  915fd982 =  5:  e9c12c21 Fixed undefined behaviour in left shift of int value
 4:  1dafe541 !  6:  5eeaf1f8 Avoiding arithmetic ops with NULL pointer in nxt_http_arguments_parse
    @@ Commit message
     
         SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/nxt_http_request.c:961:17
     
    +
      ## src/nxt_http_request.c ##
     @@ src/nxt_http_request.c: nxt_http_arguments_parse(nxt_http_request_t *r)
              return NULL;
 5:  213864c3 =  7:  e14f64b3 Avoiding arithmetic ops with NULL pointer in nxt_port_mmap_get
 6:  e6571b8e =  8:  3431dfea Initialize port_impl only when it is needed
 7:  8eb3653f =  9:  887fa467 Avoiding arithmetic ops with NULL pointer in nxt_unit_mmap_get
```

ac000 (approved):

hongzhidao (approved):

Issue #1115, opened 2024-02-06
Incorrect parsing of 64-bit integers


alejandro-colomar:
https://github.com/nginx/unit/blob/46cef09f296d9a3d54b98331d25920fc6322bea8/src/nxt_conf.c#L640

`double` is a 64-bit floating-point type.  This means it cannot represent all 64-bit integers precisely.  Or in other words, converting an `int64_t` to a `double` is a narrowing conversion which loses precision.

So, why this thing?

```c
        case NXT_CONF_MAP_INT32:
        case NXT_CONF_MAP_INT64:
        case NXT_CONF_MAP_INT:
        case NXT_CONF_MAP_SIZE:
        case NXT_CONF_MAP_OFF:
        case NXT_CONF_MAP_MSEC:

            if (v->type != NXT_CONF_VALUE_INTEGER) {
                break;
            }

            num = nxt_strtod(v->u.number, NULL);

            switch (map[i].type) {

            case NXT_CONF_MAP_INT32:
                ptr->i32 = num;
                break;

            case NXT_CONF_MAP_INT64:
                ptr->i64 = num;
                break;

            case NXT_CONF_MAP_INT:
                ptr->i = num;
                break;

            case NXT_CONF_MAP_SIZE:
                ptr->size = num;
                break;

            case NXT_CONF_MAP_OFF:
                ptr->off = num;
                break;

            case NXT_CONF_MAP_MSEC:
                ptr->msec = (nxt_msec_t) num * 1000;
                break;

            default:
                nxt_unreachable();
            }

```

If you specify a very large integer in the JSON config, it seems it's going to be silently mis-parsed, and that may cause all kinds of problems, depending on how important that number is.

alejandro-colomar:
Calling strtoimax(3) would have been better, unless I'm missing something.

alejandro-colomar:
The initial parsing seems to be done here:

<https://github.com/nginx/unit/blob/46cef09f296d9a3d54b98331d25920fc6322bea8/src/nxt_conf.c#L2195>
```c
    nxt_errno = 0;
    end = NULL;

    num = nxt_strtod(value->u.number, &end);

    if (nxt_slow_path(nxt_errno == NXT_ERANGE
        || fabs(num) > (double) NXT_INT64_T_MAX))
    {
        nxt_conf_json_parse_error(error, start,
            "The number is out of representable range.  Such JSON number "
            "value is not supported."
        );

        return NULL;
    }
```

[...]

PR #1116, opened 2024-02-07
Fix some bugs parsing integers

alejandro-colomar:
I still need to test this, but it's compiling.  :)

### DIFF
--- a/src/nxt_conf_validation.c
+++ b/src/nxt_conf_validation.c
@@ -2830,7 +2830,7 @@ nxt_conf_vldt_processes(nxt_conf_validation_t *vldt, nxt_conf_value_t *value,
     nxt_int_t                       ret;
     nxt_conf_vldt_processes_conf_t  proc;
 
-    if (nxt_conf_type(value) == NXT_CONF_NUMBER) {
+    if (nxt_conf_type(value) == NXT_CONF_INTEGER) {
         int_value = nxt_conf_get_number(value);
 
         if (int_value < 1) {


alejandro-colomar:
The regression tests (pytest) are all passing.  Now I just need to check that this does indeed fix the bug I reported yesterday.

hongzhidao:
Hi,
Thanks for the patches.
The number length in conf json is limited to 14, it looks `double` is enough.
Is there any option that takes longer?

hongzhidao:
> Using long double precision makes the code simpler, or less fragile, avoiding hard-coded magic numbers like that 14.

It depends on if there is a requirement that needs long double options. Let us know if there is.
Using 14 can simplify the code like it does now.

I think validation should happens in the control process, otherwise users don't know where the error happened.
For example:
```
"processes": 2147483648, /* INT32_MAX */

"error": "Failed to apply new configuration."
```
If the value is invalid, it should be checked in the controller process and the error can report which option is invalid.

alejandro-colomar:
Hi @hongzhidao ,

> > Using long double precision makes the code simpler, or less fragile, avoiding hard-coded magic numbers like that 14.
> 
> It depends on if there is a requirement that needs long double options. Let us know if there is.

No, there is not.  It's just a cleanup of the code, which right now looks suspicious, even if it's correct.

> Using 14 can simplify the code like it does now.

I would argue the opposite.  Right now, there are obscure reasons why it works, and there's a hidden dependency on the magic value of 14.  Using `long double` that dependency is removed, so the code is simpler.

Regarding raw lines of code, it isn't simpler with `double` either.  Except for the definition of the `nxt_strtold()` wrapper (which I provided to hide the cast), there's no additions or removals (I did remove an unnecessary cast, but that could be removed with `double` either).  Here's that patch:

```diff
commit 3f8b3dac0e3a0c9f28f94c06bb3e09d02dd63a6d
Author: Alejandro Colomar <alx@kernel.org>
Date:   Wed Feb 7 11:56:58 2024 +0100

    Configuration: Use 'long double' precision for parsing numbers
    
    We need to be able to parse correctly both floating-point numbers and
    64-bit integers.  'double' is not appropriate, since it cannot represent
    all 64-bit integers.  'long double' is platform-dependent, but the only
    platform that makes it equivalent to 'double' is Microsoft Visual C++
    for x86.  In platforms that we support, 'long double' is at least
    80-bits wide, with 64 bits for the significand plus 1 for the sign,
    which makes it capable of representing every int64_t or uint64_t.
    
    Previously, this didn't result in bogus parsing due to a magic number,
    NXT_CONF_MAX_NUMBER_LEN, which happens to be the maximum length of an
    integer in base 10 that can be represented by a 'double'.  By using
    'long double', we don't need to rely on the value of that magic number,
    so the code is more robust.
    
    See-also: 68c6b67ffc84 ("Configuration: support for rational numbers.")
    Link: <https://github.com/nginx/unit/issues/1115>
    Link: <https://en.wikipedia.org/wiki/Long_double>
    Reported-by: Alejandro Colomar <alx@kernel.org>
    Suggested-by: Andrew Clayton <a.clayton@nginx.com>
    Cc: Dan Callahan <d.callahan@f5.com>
    Cc: Zhidao Hong <z.hong@f5.com>
    Cc: Valentin Bartenev <vbartenev@gmail.com>
    Signed-off-by: Alejandro Colomar <alx@kernel.org>

diff --git a/src/nxt_conf.c b/src/nxt_conf.c
index 008cb968..99010f00 100644
--- a/src/nxt_conf.c
+++ b/src/nxt_conf.c
@@ -587,9 +587,9 @@ nxt_int_t
 nxt_conf_map_object(nxt_mp_t *mp, nxt_conf_value_t *value, nxt_conf_map_t *map,
     nxt_uint_t n, void *data)
 {
-    double            num;
     nxt_str_t         str, *s;
     nxt_uint_t        i;
+    long double       num;
     nxt_conf_value_t  *v;
 
     union {
@@ -637,7 +637,7 @@ nxt_conf_map_object(nxt_mp_t *mp, nxt_conf_value_t *value, nxt_conf_map_t *map,
                 break;
             }
 
-            num = nxt_strtod(v->u.number, NULL);
+            num = nxt_strtold(v->u.number, NULL);
 
             switch (map[i].type) {
 
@@ -2090,9 +2090,9 @@ static u_char *
 nxt_conf_json_parse_number(nxt_mp_t *mp, nxt_conf_value_t *value, u_char *start,
     u_char *end, nxt_conf_json_error_t *error)
 {
-    u_char  *p, *s, ch, c, *dot_pos;
-    size_t  size;
-    double  num;
+    u_char       *p, *s, ch, c, *dot_pos;
+    size_t       size;
+    long double  num;
 
     s = start;
     ch = *s;
@@ -2192,10 +2192,9 @@ nxt_conf_json_parse_number(nxt_mp_t *mp, nxt_conf_value_t *value, u_char *start,
     nxt_errno = 0;
     end = NULL;
 
-    num = nxt_strtod(value->u.number, &end);
+    num = nxt_strtold(value->u.number, &end);
 
-    if (nxt_slow_path(nxt_errno == NXT_ERANGE
-        || fabs(num) > (double) NXT_INT64_T_MAX))
+    if (nxt_slow_path(nxt_errno == NXT_ERANGE || fabsl(num) > NXT_INT64_T_MAX))
     {
         nxt_conf_json_parse_error(error, start,
             "The number is out of representable range.  Such JSON number "
@@ -2206,13 +2205,13 @@ nxt_conf_json_parse_number(nxt_mp_t *mp, nxt_conf_value_t *value, u_char *start,
     }
 
     if (nxt_slow_path(end == NULL || *end != '\0')) {
-        nxt_thread_log_alert("strtod(\"%s\", %s) failed %E", value->u.number,
+        nxt_thread_log_alert("strtold(\"%s\", %s) failed %E", value->u.number,
                              end == NULL ? (u_char *) "NULL" : end, nxt_errno);
         return NULL;
     }
 
-    value->type = (num == trunc(num)) ? NXT_CONF_VALUE_INTEGER
-                                      : NXT_CONF_VALUE_NUMBER;
+    value->type = (num == truncl(num)) ? NXT_CONF_VALUE_INTEGER
+                                       : NXT_CONF_VALUE_NUMBER;
 
     return p;
 }
diff --git a/src/nxt_string.h b/src/nxt_string.h
index 18ea5490..2173838f 100644
--- a/src/nxt_string.h
+++ b/src/nxt_string.h
@@ -2,6 +2,7 @@
 /*
  * Copyright (C) Igor Sysoev
  * Copyright (C) NGINX, Inc.
+ * Copyright 2024, Alejandro Colomar <alx@kernel.org>
  */
 
 #ifndef _NXT_STRING_H_INCLUDED_
@@ -20,6 +21,9 @@
 #define nxt_strtod(s, endptr)                                                 \
     strtod((char *) s, (char **) endptr)
 
+#define nxt_strtold(s, endptr)                                                \
+    strtold((const char *) s, (char **) endptr)
+
 
 #define nxt_strlen(s)                                                         \
     strlen((char *) s)
```

> 
> I think validation should happens in the control process, otherwise users don't know where the error happened. For example:
> 
> ```
> "processes": 2147483648, /* INT32_MAX */
> 
> "error": "Failed to apply new configuration."
> ```
> 
> If the value is invalid, it should be checked in the controller process and the error can report which option is invalid.

Agree.  But that would be a more aggressive change.  If you welcome a patch for moving the range checks into that code, I can have a look at it.

Cheers,
Alex

hongzhidao:
TBH, I didn't understand if it was an issue.
I'd like to ask others to have a look at it, thanks.

ac000:
With these patches shouldn't `NXT_CONF_MAX_NUMBER_LEN` be removed?

This fix kind of morphed into something bigger with lots of funky macro magic (still trying to understand it all).

If it was just a simple fix the precision thing, i.e just the long double thing without all the other stuff... but (and being brutally honest) I'm kind of meh about the _whole_ thing... certainly not inclined to start another multi-hundred comment thread where we just go round in circles for days...

> "processes": 2147483648, /* INT32_MAX */

Heh, likely way larger than (the default tuned value of) `/proc/sys/kernel/pid_max` (yes, I know it was just to demonstrate a problem).

alejandro-colomar:
Hi Andrew,

On Wed, Feb 07, 2024 at 10:13:53AM -0800, Andrew Clayton wrote:
> With these patches shouldn't `NXT_CONF_MAX_NUMBER_LEN` be removed?
> 
> This fix kind of morphed into something bigger with lots of funky
> macro magic (still trying to understand it all).

(I'll add some comments showing how all the magic works.)

Yeah, I realized I had to resort to those magic macros for the range
checks.  If you have a look at glibc, they have something similar
somewhere, IIRC.

> If it was just a simple fix the precision thing, i.e just the long
> double thing without all the other stuff...

If you want to diggest them in separate PRs (or mailed patches), I can
send them.  Just apply as much as you want, and leave the others for
when you feel like it.

Maybe I should separate the patch that uses strtoimax(3) from the
addition of range checks, since strtoimax(3) already removes Undefined
Behavior (but still misbehaves, since it truncates bignums), and the
range checks which fix the misbehavior.

> but (and being brutally honest) I'm kind of meh about the _whole_
> thing... certainly not inclined to start another multi-hundred comment
> thread where we just go round in circles for days...
> 
> > "processes": 2147483648, /* INT32_MAX */
> 
> Heh, likely way larger than (the default tuned value of)
> `/proc/sys/kernel/pid_max` (yes, I know it was just to demonstrate a
> problem).

Have a lovely night!
Alex

-- 
Looking for a remote C programming job at the moment.

src/nxt_unit_typetraits.h
```diff
@@ -0,0 +1,34 @@
...
+#define _NXT_UNIT_TYPETRAITS_H_INCLUDED_
+
+
+#include <limits.h>
+
+
+#define nxt_widthof(T)      (sizeof(T) * CHAR_BIT)
+
+
+#define nxt_is_unsigned(x)  (((typeof(x)) -1) > 1)
```
alejandro-colomar, line 15: `(typeof(x))` is a cast.

Thisnk of:

```c
((cast) -1 > 1)
```

which is equivalent to

```c
((cast) -1) > 0
```

Now substitue with two types:

```c
((int) -1 > 0)   // of course, this is false
```

and

```c
((unsigned int) -1 > 0)
```
which is
```c
(UINT_MAX > 0)  // of course, true
```

src/nxt_unit_typetraits.h
```diff
@@ -0,0 +1,34 @@
...
+#define nxt_is_signed(x)    (((typeof(x)) -1) < 1)
+
+
+#define nxt_stype_max(T)                                                      \
```
alejandro-colomar, line 19: We can't overflow, so this calculation is tricky.

Let's say our type has 8 bits.

```c
(widthof(T) - 2)  // 6
```

```c
((T) 1 << 6)   // 0b0100_0000
```

```c
((0b0100_0000 - 1) << 1)  // 0x0111_1110
```

```c
(0b0111_1110 + 1)  // 0b0111_1111
```

Works with any signed integer without triggering UB

src/nxt_unit_typetraits.h
```diff
@@ -0,0 +1,34 @@
...
+(                                                                             \
+    (T) (((((T) 1 << (nxt_widthof(T) - 2)) - 1) << 1) + 1)                    \
+)
+
+#define nxt_utype_max(T)    ((T) -1)
```
alejandro-colomar, line 24:
```c
(unsigned) -1
```

is a usual way to get UINT_MAX; I guess this one doesn't need to be explained.

src/nxt_unit_typetraits.h
```diff
@@ -0,0 +1,34 @@
...
+
+#define nxt_type_min(T)     ((T) ~nxt_type_max(T))
```
alejandro-colomar, line 26: For uint8_t:

```c
(uint8_t) ~0xFF  // 0
```

For int8_t:
```c
(int8_t) ~0b0111_1111  // 0b1000_0000, 0x7F, INT8_MIN
```

src/nxt_unit_typetraits.h
```diff
@@ -0,0 +1,34 @@
...
+
+#define nxt_type_max(T)                                                       \
+(                                                                             \
+    (T) (nxt_is_signed(T) ? nxt_stype_max(T) : nxt_utype_max(T))              \
```
alejandro-colomar, line 30: And this is a simple ternary, with little magic.

alejandro-colomar:
The magic macros from `nxt_typetraits.h` come from <https://www.alejandro-colomar.es/src/alx/alx/lib/liba2i.git/tree/include/a2i/typetraits.h>, which itself comes from a patch set for shadow.  That library has comprehensive tests, so I trust these macros.

src/nxt_unit_typetraits.h
alejandro-colomar: See what glibc has:

```c
$ grep -rn -- '- *2)* *- *1)'
atoi/time/tst-mktime2.c:26:	: ((((t) 1 << (sizeof (t) * CHAR_BIT - 2)) - 1) * 2 + 1)))
atoi/include/intprops.h:59:        : ((((t) 1 << (TYPE_WIDTH (t) - 2)) - 1) * 2 + 1)))
atoi/include/intprops.h:72:  (((_GL_INT_CONVERT (e, 1) << (TYPE_WIDTH (+ (e)) - 2)) - 1) * 2 + 1)
null/time/tst-mktime2.c:26:	: ((((t) 1 << (sizeof (t) * CHAR_BIT - 2)) - 1) * 2 + 1)))
null/include/intprops.h:59:        : ((((t) 1 << (TYPE_WIDTH (t) - 2)) - 1) * 2 + 1)))
null/include/intprops.h:72:  (((_GL_INT_CONVERT (e, 1) << (TYPE_WIDTH (+ (e)) - 2)) - 1) * 2 + 1)
sscanf/time/tst-mktime2.c:26:	: ((((t) 1 << (sizeof (t) * CHAR_BIT - 2)) - 1) * 2 + 1)))
sscanf/include/intprops.h:59:        : ((((t) 1 << (TYPE_WIDTH (t) - 2)) - 1) * 2 + 1)))
sscanf/include/intprops.h:72:  (((_GL_INT_CONVERT (e, 1) << (TYPE_WIDTH (+ (e)) - 2)) - 1) * 2 + 1)
master/time/tst-mktime2.c:26:	: ((((t) 1 << (sizeof (t) * CHAR_BIT - 2)) - 1) * 2 + 1)))
master/include/intprops.h:59:        : ((((t) 1 << (TYPE_WIDTH (t) - 2)) - 1) * 2 + 1)))
master/include/intprops.h:72:  (((_GL_INT_CONVERT (e, 1) << (TYPE_WIDTH (+ (e)) - 2)) - 1) * 2 + 1)
```

alejandro-colomar:
BTW, the typertraits macros are also a requirement for the gzip patches, so merging them now would be a win, to reduce the later review for those.  @ac000

ac000:
Hi Alex,

Thanks for the explanations.

widthof() was easy ;), though I would have called it nr_bits() or somesuch to remove ambiguity as to what units it's in.

I did work out the is_unsigned()/signed() ones also...

Hadn't really started to unravel the rest so thanks for saving me some brain cells.

However it's probably not me you need to convince about them...

alejandro-colomar:
> Hi Alex,
> 
> Thanks for the explanations.

You're welcome!

> widthof() was easy ;), though I would have called it nr_bits() or somesuch to remove ambiguity as to what units it's in.

ISO C and GCC are standardizing on "width" for referring to the number of bytes that an object takes in memory.  The Linux man-pages is also following.

> I did work out the is_unsigned()/signed() ones also...
> 
> Hadn't really started to unravel the rest so thanks for saving me some brain cells.

=)

> 
> However it's probably not me you need to convince about them...

Fingers crossed :)

alejandro-colomar:
> Btw, could you be more specific about the configuration?
> 
> ```
> "processes": 2147483648,
> ```
> 
> What's the different result between the current Unit and patches applied for it?

With current Unit, that input produces Undefined Behavior, which on my system today reproduces as Unit trying to spin up too many processes, and dying.

After my patch, Unit detects that this number cannot be represented in an int32_t, as it's `INT32_MAX + 1`, and immediately reports an error in the reconfiguration.

hongzhidao:
> Btw, could you be more specific about the configuration?
"processes": 2147483648, /* INT32_MAX */
What's the different result between the current Unit and patches applied for it?

> When using INT32_MAX, it tries to spin up so many processes, which takes 3 seconds to reconfigure, until it realizes it can't do it.
When reconfiguring with INT32_MAX + 1, it immediately reports an error, due to the new range checks.
With the current master branch, both cases would take a few seconds, since there's no range check.

I was talking about how to test.
Is it correct that the fixing is to throw the error faster? Or is there another option to test it?

alejandro-colomar:
> > Btw, could you be more specific about the configuration?
> > "processes": 2147483648, /* INT32_MAX */
> > What's the different result between the current Unit and patches applied for it?
> 
> > When using INT32_MAX, it tries to spin up so many processes, which takes 3 seconds to reconfigure, until it realizes it can't do it.
> > When reconfiguring with INT32_MAX + 1, it immediately reports an error, due to the new range checks.
> > With the current master branch, both cases would take a few seconds, since there's no range check.
> 
> I was talking about how to test. Is it correct that the fixing is to throw the error faster? Or is there another option to test it?

There are other ways: you can enable the debug log, and see if it tries to run so many processes.

hongzhidao:
Thanks for the explanation.
The first commit looks incomplete since it would be confusing that it supports long double, but actually, the max length of number is 14, it's ok to use `strtod` .
And we need to find a read-world requirement of whether there are options that take longer if we want to change it. Less memory allocation is better.
Do subsequent commits depend on the first?

ac000:
On Wed, 07 Feb 2024 15:07:35 -0800

> ISO C and GCC are standardizing on "width" for referring to the number of bytes that an

I assume you mean _bits_...

hongzhidao:
> When reconfiguring with INT32_MAX + 1, it immediately reports an error, due to the new range checks.

Since there are no specific instructions about the type of `processes`, the configuration is for users, I'm curious if they care what that value means.
Sorry, I'm not sure if that's an issue.

alejandro-colomar:
@hongzhidao 
> Thanks for the explanation. The first commit looks incomplete since it would be confusing that it supports long double, but actually, the max length of number is 14, it's ok to use `strtod` .

Yeah, you're right on that.  Maybe Andrew is right, and that 14 should be updated at the same time.

> And we need to find a read-world requirement of whether there are options that take longer if we want to change it. Less memory allocation is better.

5 bytes per number isn't going to buy us much.  There's other places where we waste a lot more.  But, that commit is the least important of the PR, so I wouldn't mind dropping it.

> Do subsequent commits depend on the first?

No.

@ac000 
> I assume you mean _bits_...

D'oh!  Yep.

hongzhidao:
> 5 bytes per number isn't going to buy us much. 

It depends on whether there is a requirement.

alejandro-colomar:
v2 changes:

-  Update the magic number.

```
$ git range-diff ngx/master gh/strtod strtod 
1:  3f8b3dac ! 1:  5a5c2321 Configuration: Use 'long double' precision for parsing numbers
    @@ Commit message
     
         Previously, this didn't result in bogus parsing due to a magic number,
         NXT_CONF_MAX_NUMBER_LEN, which happens to be the maximum length of an
    -    integer in base 10 that can be represented by a 'double'.  By using
    -    'long double', we don't need to rely on the value of that magic number,
    -    so the code is more robust.
    +    integer in base 10 that can be represented by a 'double' (14 bits).  By
    +    using 'long double', we can now parse the full [u]int64_t range, so the
    +    code is more robust.
    +
    +    Update the value of NXT_CONF_MAX_NUMBER_LEN, so that we can parse the
    +    full range of integers.
     
         See-also: 68c6b67ffc84 ("Configuration: support for rational numbers.")
         Link: <https://github.com/nginx/unit/issues/1115>
    @@ Commit message
     
      ## src/nxt_conf.c ##
    +@@
    + #include <math.h>
    + 
    + 
    +-#define NXT_CONF_MAX_SHORT_STRING  14
    +-#define NXT_CONF_MAX_NUMBER_LEN    14
    ++#define NXT_CONF_MAX_SHORT_STRING  21
    ++#define NXT_CONF_MAX_NUMBER_LEN    21
    + #define NXT_CONF_MAX_STRING        NXT_INT32_T_MAX
    + 
    + #define NXT_CONF_MAX_TOKEN_LEN     256
     @@ src/nxt_conf.c: nxt_int_t
      nxt_conf_map_object(nxt_mp_t *mp, nxt_conf_value_t *value, nxt_conf_map_t *map,
          nxt_uint_t n, void *data)
2:  a3f9dede = 2:  167b89fd Libunit: Add macros that give information about a type
3:  66295132 ! 3:  9310a302 Configuration: Check for overflow when parsing integers
    @@ src/nxt_conf.c
     +#include <nxt_unit_typetraits.h>
      
      
    - #define NXT_CONF_MAX_SHORT_STRING  14
    + #define NXT_CONF_MAX_SHORT_STRING  21
     @@ src/nxt_conf.c: nxt_int_t
      nxt_conf_map_object(nxt_mp_t *mp, nxt_conf_value_t *value, nxt_conf_map_t *map,
          nxt_uint_t n, void *data)
4:  8d186ae1 = 4:  63c157ba Configuration: Use strtoimax(3) for parsing integers
```

alejandro-colomar:
> > 5 bytes per number isn't going to buy us much.
> 
> It depends on whether there is a requirement. Let's assume 14 length is OK for the configuration options, it looks the fourth commit is unnecessary.
> 
> > Reading integers as 'double's and then doing a conversion is very
> > dangerous, and likely to trigger Undefined Behavior.  Using strtoimax(3)
> > is better.  Split the API for getting configuration numbers into an API
> > for double and another one for intmax_t.
> 
> `strtod` is enough for it here.

No.  strtod(3) can invoke undefined behavior when assigning to an int32_t.

```c
int32_t  i32;

i32 = strtod("2147483648", NULL, 10);
```

This code is UB, and so commit 4 is necessary.

hongzhidao:
> No. strtod(3) can invoke undefined behavior when assigning to an int32_t.

> int32_t  i32;
i32 = strtod("2147483648", NULL, 10);

It looks like it's because the field `processes` doesn't use the right type.

hongzhidao:
Hi,
Sorry, I didn't see where the first commit is used. Is it related to the ticket?
> Libunit: Add macros that give information about a type

hongzhidao:
> Configuration: Check for overflow when parsing integers

I feel it's unreasonable to introduce value/type checking in `nxt_conf_map_object()`, the function intends to package an object like its naming. When calling it, the conf value should be valid.

alejandro-colomar:
> Btw, to help others confirm the requirement, could you share an example of a configuration that has an issue?

<https://github.com/nginx/unit/issues/1115#issuecomment-1931007047>

alejandro-colomar:
> > Configuration: Check for overflow when parsing integers
> 
> I feel it's unreasonable to introduce value/type checking in `nxt_conf_map_object()`, the function intends to package an object like its naming.

> When calling it, the conf value should be valid.

I agree that ideally it should already be valid, but the thing is that it isn't valid.

While looking at the validation code to see how much it would be needed to fix this in the validation code, I've found something puzzling.  There are actually tests for that, but they are not triggering somehow:

```c
$ grepc nxt_conf_vldt_processes .
./src/nxt_conf_validation.c:static nxt_int_t nxt_conf_vldt_processes(nxt_conf_validation_t *vldt,
    nxt_conf_value_t *value, void *data);
./src/nxt_conf_validation.c:static nxt_int_t
nxt_conf_vldt_processes(nxt_conf_validation_t *vldt, nxt_conf_value_t *value,
    void *data)
{
    int64_t                         int_value;
    nxt_int_t                       ret;
    nxt_conf_vldt_processes_conf_t  proc;

    if (nxt_conf_type(value) == NXT_CONF_NUMBER) {
        int_value = nxt_conf_get_number(value);
        // If I insert a debugging print here, it won't trigger.  This branch is dead, for some reason.

        if (int_value < 1) {
            return nxt_conf_vldt_error(vldt, "The \"processes\" number must be "
                                       "equal to or greater than 1.");
        }

        if (int_value > NXT_INT32_T_MAX) {
            return nxt_conf_vldt_error(vldt, "The \"processes\" number must "
                                       "not exceed %d.", NXT_INT32_T_MAX);
        }

        return NXT_OK;
    }

    ret = nxt_conf_vldt_object(vldt, value, data);
    if (ret != NXT_OK) {
        return ret;
    }

    proc.spare = 0;
    proc.max = 1;
    proc.idle_timeout = 15;

    ret = nxt_conf_map_object(vldt->pool, value,
                              nxt_conf_vldt_processes_conf_map,
                              nxt_nitems(nxt_conf_vldt_processes_conf_map),
                              &proc);
    if (ret != NXT_OK) {
        return ret;
    }

    if (proc.spare < 0) {
        return nxt_conf_vldt_error(vldt, "The \"spare\" number must not be "
                                   "negative.");
    }

    if (proc.spare > NXT_INT32_T_MAX) {
        return nxt_conf_vldt_error(vldt, "The \"spare\" number must not "
                                   "exceed %d.", NXT_INT32_T_MAX);
    }

    if (proc.max < 1) {
        return nxt_conf_vldt_error(vldt, "The \"max\" number must be equal "
                                   "to or greater than 1.");
    }

    if (proc.max > NXT_INT32_T_MAX) {
        return nxt_conf_vldt_error(vldt, "The \"max\" number must not "
                                   "exceed %d.", NXT_INT32_T_MAX);
    }

    if (proc.max < proc.spare) {
        return nxt_conf_vldt_error(vldt, "The \"spare\" number must be "
                                   "less than or equal to \"max\".");
    }

    if (proc.idle_timeout < 0) {
        return nxt_conf_vldt_error(vldt, "The \"idle_timeout\" number must not "
                                   "be negative.");
    }

    if (proc.idle_timeout > NXT_INT32_T_MAX / 1000) {
        return nxt_conf_vldt_error(vldt, "The \"idle_timeout\" number must not "
                                   "exceed %d.", NXT_INT32_T_MAX / 1000);
    }

    return NXT_OK;
}
```

hongzhidao:
>I agree that ideally it should already be valid, but the thing is that it isn't valid.

I'm asking others to check if it's an issue.
Yep, it needs more work on the controller validation, maybe it's the intention of why only introduce VLDT_INTEGER for the json configuration to simplify the handling. And we need to throw a configuration error but not "failed to apply..." if it's an issue to help users to correct their options.

alejandro-colomar:
v5 changes:

-  Fix incorrect validation of "processes".

After this fix, the validation of "processes" is actually nice:

```
$ sudo unit ctl http PUT /config <proc.json 
{
	"detail": "The \"processes\" number must not exceed 2147483647."
}
```

Now I feel like the other patches are less necessary.  At least we know it works.  So, feel free to ask me dropping some patches.

```
$ git range-diff master gh/strtod strtod 
1:  bf86e8a9 = 1:  bf86e8a9 Libunit: Add macros that give information about a type
2:  aa085d5d = 2:  aa085d5d Configuration: Check for overflow when parsing integers
3:  0f3a215a = 3:  0f3a215a Configuration: Be careful when comparing integers and floating
4:  dfd4b077 = 4:  dfd4b077 Configuration: Use strtoimax(3) for parsing integers
-:  -------- > 5:  2a3b09e3 Configuration: "processes" is an integer, not a floating number
```

alejandro-colomar:
v6 changes:

-  Drop patch dfd4b077 ("Configuration: Use strtoimax(3) for parsing integers")

Now that we keep the 14-digit magic, and we parse as double, using strtod(3) everywhere serves to remind us of the reasons behind the magic number.

```
$ git range-diff master gh/strtod strtod 
1:  bf86e8a9 = 1:  bf86e8a9 Libunit: Add macros that give information about a type
2:  aa085d5d = 2:  aa085d5d Configuration: Check for overflow when parsing integers
3:  0f3a215a = 3:  0f3a215a Configuration: Be careful when comparing integers and floating
4:  dfd4b077 < -:  -------- Configuration: Use strtoimax(3) for parsing integers
5:  2a3b09e3 ! 4:  e11f33e5 Configuration: "processes" is an integer, not a floating number
    @@ src/nxt_conf_validation.c: nxt_conf_vldt_processes(nxt_conf_validation_t *vldt,
      
     -    if (nxt_conf_type(value) == NXT_CONF_NUMBER) {
     +    if (nxt_conf_type(value) == NXT_CONF_INTEGER) {
    -         int_value = nxt_conf_get_integer(value);
    +         int_value = nxt_conf_get_number(value);
      
              if (int_value < 1) {
```

hongzhidao:
> v5 changes:
> 
> * Fix incorrect validation of "processes".
> 
> After this fix, the validation of "processes" is actually nice:
> 
> ```
> $ sudo unit ctl http PUT /config <proc.json 
> {
> 	"error": "Invalid configuration.",
> 	"detail": "The \"processes\" number must not exceed 2147483647."
> }
> ```
> 
> Now I feel like the other patches are less necessary. At least we know it works. So, feel free to ask me to drop some patches.
> 
> ```
> $ git range-diff master gh/strtod strtod 
> 1:  bf86e8a9 = 1:  bf86e8a9 Libunit: Add macros that give information about a type
> 2:  aa085d5d = 2:  aa085d5d Configuration: Check for overflow when parsing integers
> 3:  0f3a215a = 3:  0f3a215a Configuration: Be careful when comparing integers and floating
> 4:  dfd4b077 = 4:  dfd4b077 Configuration: Use strtoimax(3) for parsing integers
> -:  -------- > 5:  2a3b09e3 Configuration: "processes" is an integer, not a floating number
> ```

Good catch, thanks for the job.
It looks like the correct way to fix it.

I agree to fix it only with the commit, thanks.

alejandro-colomar:
v7 changes:

-  Drop the first two commits.  Now that the validation code rejects bad values, we don't need to check again.  [@hongzhidao ]

```
$ git range-diff master gh/strtod strtod 
1:  bf86e8a9 < -:  -------- Libunit: Add macros that give information about a type
2:  aa085d5d < -:  -------- Configuration: Check for overflow when parsing integers
3:  0f3a215a < -:  -------- Configuration: Be careful when comparing integers and floating
-:  -------- > 1:  960ffdc4 Configuration: Be careful when comparing integers and floating
4:  e11f33e5 = 2:  8fae58a6 Configuration: "processes" is an integer, not a floating number
```

alejandro-colomar:
v7b changes:

-  Add some commit metadata

```
$ git range-diff master gh/strtod strtod 
1:  960ffdc4 ! 1:  ed32bf46 Configuration: Be careful when comparing integers and floating
    @@ Commit message
             ^~~~~~~~~~~~~~~~~~~~
             1 warning generated.
     
    +    Fixes: 68c6b67ffc84 ("Configuration: support for rational numbers.")
    +    Link: <https://github.com/nginx/unit/pull/1116>
    +    Link: <https://github.com/nginx/unit/issues/1115>
    +    Cc: Zhidao Hong <z.hong@f5.com>
    +    Cc: Andrew Clayton <a.clayton@nginx.com>
    +    Cc: Dan Callahan <d.callahan@f5.com>
    +    Cc: Valentin Bartenev <vbartenev@gmail.com>
     
      ## src/nxt_conf.c ##
2:  8fae58a6 ! 2:  20b1d7e7 Configuration: "processes" is an integer, not a floating number
    @@ Commit message
     
         Fixes: 68c6b67ffc84 ("Configuration: support for rational numbers.")
         Closes: <https://github.com/nginx/unit/issues/1115>
    +    Link: <https://github.com/nginx/unit/pull/1116>
         Cc: Zhidao Hong <z.hong@f5.com>
         Cc: Dan Callahan <d.callahan@f5.com>
```

hongzhidao:
Hi,
Btw, I would suggest putting the fixing in front of the second one. We can release it if others approve it.
The fixing looks good to me, but I'm not sure for the other one.

alejandro-colomar:
v7c changes:

-  Tweak commit message.  [@hongzhidao ]

```
$ git range-diff master gh/strtod strtod 
1:  ed32bf46 = 1:  ed32bf46 Configuration: Be careful when comparing integers and floating
2:  20b1d7e7 ! 2:  d1831c5d Configuration: "processes" is an integer, not a floating number
    @@ Metadata
     
      ## Commit message ##
    -    Configuration: "processes" is an integer, not a floating number
    +    Configuration: Fix validation of "processes"
    +
    +    It's an integer, not a floating number.
     
         Fixes: 68c6b67ffc84 ("Configuration: support for rational numbers.")
         Closes: <https://github.com/nginx/unit/issues/1115>
```

alejandro-colomar:
v8 changes:

-  Drop patch 1.  [@hongzhidao ]

The reason for patch 1 was that `double` can't hold `INT64_MAX` so the comparison against it would be imprecise.  But since we can't even write that number in 14 digits, the problem is not real.  And if we changed to using a `long double` or an `intmax_t`, the problem would also vanish, it is a purely virtual problem.

```
$ git range-diff master gh/strtod strtod 
1:  ed32bf46 < -:  -------- Configuration: Be careful when comparing integers and floating
2:  d1831c5d = 1:  10c408be Configuration: Fix validation of "processes"
```

alejandro-colomar:
> Good catch, thanks for the job. It looks like the correct way to fix it.
> 
> I agree to fix it only with the commit, thanks.

So, do you approve v8?  Should I add `Reviewed-by: you`?

alejandro-colomar:
Cc: @ac000 

If you don't mind, please have another look.  It seems this is done.

hongzhidao:
> > Good catch, thanks for the job. It looks like the correct way to fix it.
> > I agree to fix it only with the commit, thanks.
> 
> So, do you approve v8? Should I add `Reviewed-by: ` @hongzhidao ?

LGTM. Thanks.

alejandro-colomar:
> > > Good catch, thanks for the job. It looks like the correct way to fix it.
> > > I agree to fix it only with the commit, thanks.
> > 
> > 
> > So, do you approve v8? Should I add `Reviewed-by: ` @hongzhidao ?
> 
> LGTM. Thanks.

Thank you for your strict review!  :)

v8b changes:

-  Add review tag

```
$ git range-diff master gh/strtod strtod 
1:  10c408be ! 1:  8788e4bd Configuration: Fix validation of "processes"
    @@ Commit message
         Fixes: 68c6b67ffc84 ("Configuration: support for rational numbers.")
         Closes: <https://github.com/nginx/unit/issues/1115>
         Link: <https://github.com/nginx/unit/pull/1116>
    -    Cc: Zhidao Hong <z.hong@f5.com>
         Cc: Dan Callahan <d.callahan@f5.com>
         Cc: Valentin Bartenev <vbartenev@gmail.com>
```

ac000:
Bejesus, I go to sleep for a few hours and this is what happens!

But yeah, looks like you got to the heart of the problem!

```
Reviewed-by:  Andrew Clayton <a.clayton@nginx.com>
```

Btw, we decided that URLs in `Closes:` tags should perhaps not be enclosed in `<>`'s simply because GH does not recognise them and won't auto-link/close. See [here](https://github.com/nginx/unit/pull/1020) for some chat on that...

alejandro-colomar:
> Bejesus, I go to sleep for a few hours and this is what happens!
> 
> But yeah, looks like you got to the heart of the problem!

:D

> 
> ```
> Reviewed-by:  Andrew Clayton <a.clayton@nginx.com>
> ```
> 
> Btw, we decided that URLs in `Closes:` tags should perhaps not be enclosed in `<>`'s simply because GH does not recognise them and won't auto-link/close. See [here](https://github.com/nginx/unit/pull/1020) for some chat on that...

Huh!  I've reported a bug to github.  Anyway, I don't mind workarounding that temporarily.

v8c changes:

-  tags

```
$ git range-diff master 8788e4bd strtod 
1:  8788e4bd ! 1:  9e986704 Configuration: Fix validation of "processes"
    @@ Commit message
         It's an integer, not a floating number.
     
         Fixes: 68c6b67ffc84 ("Configuration: support for rational numbers.")
    -    Closes: <https://github.com/nginx/unit/issues/1115>
    +    Closes: https://github.com/nginx/unit/issues/1115
         Link: <https://github.com/nginx/unit/pull/1116>
         Reviewed-by: Zhidao Hong <z.hong@f5.com>
    -    Cc: Andrew Clayton <a.clayton@nginx.com>
         Cc: Dan Callahan <d.callahan@f5.com>
         Cc: Valentin Bartenev <vbartenev@gmail.com>
```

ac000:
On Thu, 08 Feb 2024 06:07:59 -0800

> > Btw, we decided that URLs in `Closes:` tags should perhaps not be enclosed in `<>`'s simply because GH does not recognise them and won't auto-link/close. See [here](https://github.com/nginx/unit/pull/1020) for some chat on that...  
> 
> Huh!  I've reported a bug to github.  Anyway, I don't mind workarounding that temporarily.

Heh, I couldn't even find _where_ to report it!

alejandro-colomar:
> On Thu, 08 Feb 2024 06:07:59 -0800 Alejandro Colomar ***@***.***> wrote: > Btw, we decided that URLs in `Closes:` tags should perhaps not be enclosed in `<>`'s simply because GH does not recognise them and won't auto-link/close. See [here](#1020) for some chat on that... Huh! I've reported a bug to github. Anyway, I don't mind workarounding that temporarily.
> Heh, I couldn't even find _where_ to report it!

It wasn't easy.  I had to click through links several times, and I found something, but it's definitely hidden.  Github "likes" open source, but it seems that the door is open in only one direction...

ac000:
> > Heh, I couldn't even find _where_ to report it!
> 
> It wasn't easy. I had to click through links several times, and I found something, but it's definitely hidden. Github "likes" open source, but it seems that the door is open in only one direction...

Sadly this is all too common...

ac000 (approved):

ac000:
Thanks Alex!

PR #1123, opened 2024-02-12
Add interim GitHub Actions

arbourd:
This pull-request adds GitHub Actions configuration that allows tests to run on pull-requests and `master` branches.

**Details**

- Tests are run in parallel per language for speed and reducing flakes
  - N number of languages and versions can be set
- Module specific tests are set by a string 
  - `go-1.21` will run a job that runs Go tests against Go 1.21, etc
  - Adding `go-1.20` to the list would add an additional test for Go 1.20.
- Module specific tests are run by selecting `test/test_${module}*`
  - Python module runs all tests, including base unit ones, as the Python test suite is used as the default for other tests 
- Tests complete in about 5 minutes

**Caveats**

- Only unprivileged tests are run
- No core dumps
- Only ubuntu


.github/workflows/ci.yml
```diff
@@ -0,0 +1,327 @@
...
+            --tests \
+            --openssl \
+            --njs \
+            --cc-opt="-I njs/src/ -I njs/build"  \
+            --ld-opt="-L njs/build" \
+            --debug
+
+      - name: Make unit
+        run: |
+          make -j4 -k
```
alejandro-colomar, line 136: You probably want `make -j4 -k || make`.  Otherwise, the error output will be ugly.  Cc: @thresheek

ac000: Hmm, from doing some simple testing

```
$ make -j4 -k || make
```
isn't much different from just
```
$ make -j4
```

They both clearly show the first error.

With the `-k` option you do have the opportunity to go through the scrollback to see other errors, but with the second make you'll only see the first error... (unless any other errors just happen to be in the last compiled files).

But in GH,  you're probably going to have to scroll through the compiler output anyway to get to the bottom, so I'm not really sure the second make buys us much...

alejandro-colomar: > Hmm, from doing some simple testing
> 
> ```
> $ make -j4 -k || make
> ```
> 
> isn't much different from just
> 
> ```
> $ make -j4
> ```
> 
> They both clearly show the first error.

No, they don't:

```mk
alx@debian:~/tmp$ cat Makefile 
.PHONY: all a b
all: a b

a:
	$(info	A)
	sleep 1
	echo errors from a
	false

b:
	$(info	B)
	echo errors from b
	false
```

Simple make:
```sh
alx@debian:~/tmp$ make -k
A
sleep 1
echo errors from a
errors from a
false
make: *** [Makefile:8: a] Error 1
B
echo errors from b
errors from b
false
make: *** [Makefile:13: b] Error 1
make: Target 'all' not remade because of errors.
```

Ugly, parallel make:
```
alx@debian:~/tmp$ make -j4 -k
A
sleep 1
B
echo errors from b
errors from b
false
make: *** [Makefile:13: b] Error 1
echo errors from a
errors from a
false
make: *** [Makefile:8: a] Error 1
make: Target 'all' not remade because of errors.
```

Ugly parallel make, followed by simple make:
```
alx@debian:~/tmp$ make -j4 -k || make
A
sleep 1
B
echo errors from b
errors from b
false
make: *** [Makefile:13: b] Error 1
echo errors from a
errors from a
false
make: *** [Makefile:8: a] Error 1
make: Target 'all' not remade because of errors.
A
sleep 1
echo errors from a
errors from a
false
make: *** [Makefile:8: a] Error 1
```
In the last case, you'll see the initial part is mixed, due to parallel make, but you can just ignore it, and start reading the last part.

See <https://github.com/shadow-maint/shadow/commit/97f79e3b2715d78c9342cbb292e1a3e72a23fe62>

> 
> With the `-k` option you do have the opportunity to go through the scrollback to see other errors, but with the second make you'll only see the first error... (unless any other errors just happen to be in the last compiled files).

> 
> But in GH, you're probably going to have to scroll through the compiler output anyway to get to the bottom, so I'm not really sure the second make buys us much...

thresheek: I agree with @alejandro-colomar - the '-k || make' trick is really helpful with multiple build targets' outputs are intertwined.   We've been using it in buildbot builds for a few years too.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,327 @@
...
+
+      ##
+      ## Go
+      ##
+
+      - uses: actions/setup-go@v4
```
thresheek, line 142: It looks like this installs some version of Go that is not shipped with Ubuntu 22.04 but somewhere else. (I think it also applies to other language modules, except for Perl).

On Buildbot we use whatever is shipped on that OS (with a notable exception of Node, which we fetch from nodesource repos).  This helps us ensure our software is compatible with what is broadly used by respective OS users.

Is the intention here to test with with whatever is a "current version" of a language available from actions vendors?  If so, when a new incompatible version of a language is published it means our pipelines will be red and prevent (at least formally) from unrelated PRs to be merged?  I can see some value in having runs with new versions though.

Should we stick to specific known good versions instead?  Should we not be using those "edge" versions but revert to what is available in the OS packages repositories?

thresheek: Oh, I've just read the commit message that describes the versioning, so please disregard that part of the question.  The OS-provided vs actions-provided still stands, though.

.github/workflows/ci.yml
ac000: > No, they don't:

Oh, but I'm afraid they do...

```
$ make -j4 -k || make
...
src/nxt_file.c: In function ‘nxt_file_chown’:
src/nxt_file.c:357:52: error: ‘dummy’ undeclared (first use in this function)
  357 |                                  owner, nxt_errno, dummy,
      |                                                    ^~~~~
src/nxt_log.h:81:46: note: in definition of macro ‘nxt_log_alert’
   81 |         _log_->handler(NXT_LOG_ALERT, _log_, __VA_ARGS__);                    \
      |                                              ^~~~~~~~~~~
src/nxt_file.c:356:13: note: in expansion of macro ‘nxt_thread_log_alert’
  356 |             nxt_thread_log_alert("getpwnam_r(\"%s\", ...) failed %E %s",
      |             ^~~~~~~~~~~~~~~~~~~~
src/nxt_file.c:357:52: note: each undeclared identifier is reported only once for each function it appears in
  357 |                                  owner, nxt_errno, dummy,
      |                                                    ^~~~~
src/nxt_log.h:81:46: note: in definition of macro ‘nxt_log_alert’
   81 |         _log_->handler(NXT_LOG_ALERT, _log_, __VA_ARGS__);                    \
      |                                              ^~~~~~~~~~~
src/nxt_file.c:356:13: note: in expansion of macro ‘nxt_thread_log_alert’
  356 |             nxt_thread_log_alert("getpwnam_r(\"%s\", ...) failed %E %s",
      |             ^~~~~~~~~~~~~~~~~~~~
make: *** [build/Makefile:226: build/src/nxt_file.o] Error 1
$
```

```
$ make -j4
...
src/nxt_file.c: In function ‘nxt_file_chown’:
src/nxt_file.c:357:52: error: ‘dummy’ undeclared (first use in this function)
  357 |                                  owner, nxt_errno, dummy,
      |                                                    ^~~~~
src/nxt_log.h:81:46: note: in definition of macro ‘nxt_log_alert’
   81 |         _log_->handler(NXT_LOG_ALERT, _log_, __VA_ARGS__);                    \
      |                                              ^~~~~~~~~~~
src/nxt_file.c:356:13: note: in expansion of macro ‘nxt_thread_log_alert’
  356 |             nxt_thread_log_alert("getpwnam_r(\"%s\", ...) failed %E %s",
      |             ^~~~~~~~~~~~~~~~~~~~
src/nxt_file.c:357:52: note: each undeclared identifier is reported only once for each function it appears in
  357 |                                  owner, nxt_errno, dummy,
      |                                                    ^~~~~
src/nxt_log.h:81:46: note: in definition of macro ‘nxt_log_alert’
   81 |         _log_->handler(NXT_LOG_ALERT, _log_, __VA_ARGS__);                    \
      |                                              ^~~~~~~~~~~
src/nxt_file.c:356:13: note: in expansion of macro ‘nxt_thread_log_alert’
  356 |             nxt_thread_log_alert("getpwnam_r(\"%s\", ...) failed %E %s",
      |             ^~~~~~~~~~~~~~~~~~~~
make: *** [build/Makefile:226: build/src/nxt_file.o] Error 1
make: *** Waiting for unfinished jobs....
$
```

ac000: > See [shadow-maint/shadow@97f79e3](https://github.com/shadow-maint/shadow/commit/97f79e3b2715d78c9342cbb292e1a3e72a23fe62)

Hmm, you did it slightly differently there and not that Iker was really convinced either way...

ac000:
A better way may be something like

```shell
$ make -j4 -k 2>errors.txt
```
Then you can display that file (with a nice big banner or whatever) if there was any errors... plus you get _all_ the errors front and centre.

ac000:
If we used kernel style make output, any errors/warnings would stand out much more...

.github/workflows/ci.yml
alejandro-colomar: > > See [shadow-maint/shadow@97f79e3](https://github.com/shadow-maint/shadow/commit/97f79e3b2715d78c9342cbb292e1a3e72a23fe62)
> 
> Hmm, you did it slightly differently there and not that Iker was really convinced either way...

Yeah, take your favourite one.  Each one has advantages and disadvantages.  I suggested `|| make` because that's what BB is doing already, as K said.

> > No, they don't:
> 
> Oh, but I'm afraid they do...
> 

What you're seeing is racy, and the race just happens to behave relatively well in your case.  In your case, probably just the command is separate from the error, which is bad, but not the worst thing that can happen  Output from different compilations may get mixed very badly (I've seen it in other projects).

See:

```
$ MANWIDTH=72 man make | sed -n '/^ *-O/,/^$/p'
     -O[type], --output-sync[=type]
          When  running  multiple jobs in parallel with ‐j, ensure the
          output of each job is collected together rather than  inter‐
          spersed  with output from other jobs.  If type is not speci‐
          fied or is target the output from the entire recipe for each
          target is grouped together.  If type is line the output from
          each command line within a recipe is grouped  together.   If
          type  is  recurse  output  from  an entire recursive make is
          grouped together.  If type is none output synchronization is
          disabled.

```

From that documentation it's unclear what the default is, but it's `none`.  You can check GNU Make's source code:

```c
$ grepc decode_output_sync_flags src/
src/main.c:static void
decode_output_sync_flags (void)
{
#ifdef NO_OUTPUT_SYNC
  output_sync = OUTPUT_SYNC_NONE;
#else
  if (output_sync_option)
    {
      if (streq (output_sync_option, "none"))
        output_sync = OUTPUT_SYNC_NONE;
      else if (streq (output_sync_option, "line"))
        output_sync = OUTPUT_SYNC_LINE;
      else if (streq (output_sync_option, "target"))
        output_sync = OUTPUT_SYNC_TARGET;
      else if (streq (output_sync_option, "recurse"))
        output_sync = OUTPUT_SYNC_RECURSE;
      else
        OS (fatal, NILF,
            _("unknown output-sync type '%s'"), output_sync_option);
    }

  if (sync_mutex)
    osync_parse_mutex (sync_mutex);
#endif
}
```

alejandro-colomar: Alternatively, you may want to just specify `-Otarget`

arbourd: I'm happy to use `make -j4 -k || make` for the time being to match what we're doing elsewhere. We can do improvements in future PRs for those who are passionate about improving our error visibility.

Again, this is just an interim / first-pass at an Actions build for Unit alone, and I suspect much will change with time 😄

.github/workflows/ci.yml
callahad: I'm on holiday and setting a bad example, but my preference for the testing aligns with Dylan's:

- I want us to easily test against arbitrary versions of upstream runtimes, separate from distro-provided versions.

This isn't exclusive to testing against distro-provided dependencies; we should _also_ be testing that we successfully build on platforms we want to support, but that can be a second phase of this effort. For now, the existing BuildBot infrastructure covers us there.

thresheek: Thanks for clarifications @arbourd!  I am not against these changes, but I wanted to have clarity on what this PR tries to achieve.

> > Should we stick to specific known good versions instead? Should we not be using those "edge" versions but revert to what is available in the OS packages repositories?
> 
> Is there any reason not to do both? Is it an XOR? I hope I've laid out a competent argument about why I think testing against the non-default versions is important for Unit. Is there any context I'm missing that makes this untenable?

Nope, I think this is a correct approach.  We can start with the tests that make sense from development point of view (new or "currently stable" language versions), and then build up other parts of CI that will test the changes on what's shipped in the OS we need to support, just like we do on BB.

To add some info, for BB, we kinda sorta got away with testing our code on platforms like Alpine and FreeBSD that typically offer up-to-date language runtimes.

From my point of view, this conversation can be marked as resolved.

.github/workflows/ci.yml
alejandro-colomar: @ac000 

Today, I've seen a mixed line in a multi-process make, while testing some improvement to the Linux man-pages' build system (I'm adding a `distcheck` target).

Here's a line that was printed:
```
MAKE build-pdf (2>/dev/null):MAKE build-ps (-i): GROPS	.tmp/man/man7/koi8-r.7.ps
 GROPDF	.tmp/man/man4/sd.4.pdf
```
while it should have been
```
MAKE build-pdf (2>/dev/null): GROPDF	.tmp/man/man4/sd.4.pdf
MAKE build-ps (-i): GROPS	.tmp/man/man7/koi8-r.7.ps
```

The command was
```sh
$ make distcheck TMPDIR1=/tmp/tmp1 TMPDIR2=/tmp/tmp2 -j24 -k
```
You can guess that 24 processes competing for /dev/stdout don't behave well.

And the relevant makefile code is:
```mk
$(builddir)/distcheck.build-ps.touch: $(builddir)/distcheck.build-pre.touch
$(builddir)/distcheck.build-ps.touch: $(TMPDIR1)/$(DISTNAME) | $$(@D)/
	$(info	MAKE		build-ps)
	$(MAKE) -C $< -k build-ps $(HIDE_ERR) \
	| $(TS) 'MAKE build-ps (2>/dev/null):' \
	|| $(TRUE)
	$(MAKE) -C $< -i nothing $(DISTCHECK_IGNORE_BUILD_PS) $(HIDE_ERR) \
	| $(TS) 'MAKE build-ps (-i):'
	$(MAKE) -C $< build-ps \
	| $(TS) 'MAKE build-ps:'
	$(TOUCH) $@

$(builddir)/distcheck.build-pdf.touch: $(builddir)/distcheck.build-pre.touch
$(builddir)/distcheck.build-pdf.touch: $(TMPDIR1)/$(DISTNAME) | $$(@D)/
	$(info	MAKE		build-pdf)
	$(MAKE) -C $< -k build-pdf $(HIDE_ERR) \
	| $(TS) 'MAKE build-pdf (2>/dev/null):' \
	|| $(TRUE)
	$(MAKE) -C $< -i nothing $(DISTCHECK_IGNORE_BUILD_PDF) $(HIDE_ERR) \
	| $(TS) 'MAKE build-pdf (-i):'
	$(MAKE) -C $< build-pdf \
	| $(TS) 'MAKE build-pdf:'
	$(TOUCH) $@
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,327 @@
...
+          - build: perl
+            os: ubuntu-latest
+          - build: php-8.3
+            os: ubuntu-latest
+          - build: python-3.12
+            os: ubuntu-latest
+          - build: ruby-3.3
+            os: ubuntu-latest
+          - build: wasm
+            os: ubuntu-latest
```
callahad, line 39: Should we update this matrix to match our Dockerfiles? That would currently mean:

- [ ] Add Go 1.22
- [ ] Add Node 20
- [ ] Add PHP 8.2
- [ ] Add Python 3.11
- [ ] Add Ruby 3.2
- [ ] Add Perl 5.36 and 5.38
- [ ] Drop Perl 5.34 (default in `ubuntu-latest`)

I'm... not sure that makes sense right now. Maybe best to merge as is and discuss coverage after release. Thoughts, @arbourd ?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,327 @@
...
+        if: steps.metadata.outputs.module == 'ruby'
+
+      ##
+      ## Wasm
+      ##
+
+      - name: Make wasmtime
+        run: |
+          make -C pkg/contrib .wasmtime
+        if: steps.metadata.outputs.module == 'wasm'
```
ac000, line 296: What version of wasmtime does this use?

callahad: Looks like we're at Wasmtime 11 in `pkgs/contrib`

ac000:
On Tue, 20 Feb 2024 08:30:53 -0800
Dan Callahan ***@***.***> wrote:

> @callahad commented on this pull request.
> 
> 
> 
> > +      - name: Make wasmtime  
> +        run: |
> +          make -C pkg/contrib .wasmtime
> +        if: steps.metadata.outputs.module == 'wasm'
> 
> Looks like we're at Wasmtime 11 in `pkgs/contrib`

Hmm, that's pretty old. Any chance of using something more recent? In
fact I think I'd be OK with using whatever the latest release is for
now...

.github/workflows/ci.yml
arbourd: Ya I'm happy to match those. 

Couple of issues with Perl and PHP with the default Actions runner:

- I haven't been able to get the community `setup-perl` action to play nice yet
- The community `setup-php` is also broken for anything other than 8.3 due to potential missing SAPI's.

Maybe net-add the missing ones and then follow up with Perl and PHP after?

callahad: Works for me. I have some ideas for preventing our matrix from exploding, but for now, let the combinatorics explode.

callahad (approved):
This is clearly Good Enough to be useful; let's merge now and refine later.

callahad:
At your discretion, consider adding a richer commit message / subject line. It could be useful to have some of the caveats spelled out there (mainly, that this is an initial pass and not intended to be comprehensive; the emphasis is on providing a Good Enough smoke test for development purposes)

.github/workflows/ci.yml
alejandro-colomar: We've merged the patch with `-Orecurse` in shadow:

-  <https://github.com/shadow-maint/shadow/pull/950>

You may be interested in doing something similar here.

alejandro-colomar:
We've merged some

> A better way may be something like
> 
> ```shell
> $ make -j4 -k 2>errors.txt
> ```
> 
> Then you can display that file (with a nice big banner or whatever) if there was any errors... plus you get _all_ the errors front and centre.

That doesn't show the errors together with the commands that produced them.  Plus, that doesn't solve the races.

ac000:
On Wed, 13 Mar 2024 09:38:12 -0700

> > A better way may be something like
> > 
> > ```shell
> > $ make -j4 -k 2>errors.txt
> > ```
> > 
> > Then you can display that file (with a nice big banner or whatever) if there was any errors... plus you get _all_ the errors front and centre.  
> 
> That doesn't show the errors together with the commands that produced them.  Plus, that doesn't solve the races.
> 

This whole thing is pretty much moot now we have prettified output...

alejandro-colomar:
> On Wed, 13 Mar 2024 09:38:12 -0700 Alejandro Colomar ***@***.***> wrote: > A better way may be something like > > ```shell > $ make -j4 -k 2>errors.txt > ``` > > Then you can display that file (with a nice big banner or whatever) if there was any errors... plus you get _all_ the errors front and centre. That doesn't show the errors together with the commands that produced them. Plus, that doesn't solve the races.
> This whole thing is pretty much moot now we have prettified output...

Not really.  You're still subject to races.  And redirecting stderr still leaves you without the corresponding `  CC  ...` next to the error.

So, think that the following may happen:

```
  CC  foo.o
some error line from compiling foo.o
  CC bar.o
another error line from compiling foo.o
```

ac000:
On Wed, 13 Mar 2024 09:53:49 -0700

> So, think that the following may happen:
> 
> ```
>   CC  foo.o
> some error line from compiling foo.o
>   CC bar.o
> another error line from compiling foo.o
> ```
> 

But the error itself will show where the problem lies...

alejandro-colomar:
> But the error itself will show where the problem lies...

Yeah, but if several error reports are intertwined due to a race, it will be hard to read each of them.

Think of:

```
error line 1 of foo.o
error line 1 of bar.o
error line 2 of foo.o
error line 2 of bar.o
error line 3 of bar.o
error line 4 of bar.o
error line 3 of foo.o
```

I see no reasons to avoid `-Otarget`, especially when it doesn't impose observable performance degradation (I couldn't measure any degradation in projects that I use often, including the Linux man-pages and shadow).

Issue #982, opened 2023-10-24
Docker images should redirect error log message to stderr (not stdout)


lcrilly:
To facilitate effective log streaming from Unit containers it would be better if the Unit log file was symlinked to /dev/stderr instead of /dev/stdout.

This leaves space for stdout to be used for [access logging](https://unit.nginx.org/configuration/#access-log).

Originated in https://github.com/nginx/unit/issues/973

Issue #973, opened 2023-10-17
Startup message logged to std


andrejohansson:
I'm running the unit docker image and when it starts up it logs a message to **stderr** which in turns causes the logs to be picked up by our gcp logging streamer which sees it as an error.

```
2023/10/17 08:18:46 [info] 1#1 unit 1.29.1 started"
```

Why is it defaulting to stderr for that info message?
Can I turn it of or redirect it somehow?

I see that I can redirect application logs here on an *application level* but how do I do this on a global level (which I guess is were startup log messages reside):
https://unit.nginx.org/configuration/#settings

andrejohansson:
Thanks for your reply and the information!

But I fail to understand your link, you say that logs are _typically redirected to **stderr**_ but when I check [the code](https://github.com/nginx/unit/blob/8c4425ccb9a413e8d0506e0254f0e84bd89a32a6/pkg/docker/Dockerfile.minimal#L80) it´s redirecting to **stdout** (which would have been fine for us):

```
    && ln -sf /dev/stdout /var/log/unit.log
```

PR #1128, opened 2024-02-14
Docker: redirect logs to stderr

thresheek:
This allows to use /dev/stdout as application logs if users choose to do so.

Fixes https://github.com/nginx/unit/issues/982

### DIFF
--- a/pkg/docker/template.Dockerfile
+++ b/pkg/docker/template.Dockerfile
@@ -77,7 +77,7 @@ RUN set -ex \
     && apt-get purge -y --auto-remove build-essential \
     && rm -rf /var/lib/apt/lists/* \
     && rm -f /requirements.apt \
-    && ln -sf /dev/stdout /var/log/unit.log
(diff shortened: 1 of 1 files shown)


ac000:
Please use `Closes:` instead of `Fixes`. We're using `Fixes:` for when referring back to previous commits that introduced some issue.

ac000:
No need for the trailing `.` on the subject line.

thresheek:
Removed trailing dot on the subject line, and changed wording to use 'Closes:' instead.

Thanks!

tippexs (approved):
LGTM

callahad:
This seems somewhat weird to me, but it's a small and relatively innocuous change. We'll want to include it in the release notes and come back around to more sophisticated logging configurations when we address #973

callahad:
We'll also need to regenerate all Dockerfiles from the template, but there are a few other incoming Docker changes that I'm about to merge. I'll regenerate once those are all in.

PR #1131, opened 2024-02-16
Docker: Switch to github

thresheek:


ac000 (approved):
That leaves us in a detached head state, but I guess that's fine for this use case...

arbourd (approved):

PR #1132, opened 2024-02-16
Packages: Move dist target to git archive

thresheek:


.gitattributes
```diff
@@ -1,2 +1,6 @@
 *.c	diff=cpp
 *.h	diff=cpp
+.hg*	export-ignore
```
ac000, line 3: Maybe a blank line between these two.

thresheek: Yep, agreed. Done.

ac000 (approved):

PR #1135, opened 2024-02-19
Node.js: Use console.warn instead of stderr.write

callahad:
Functionally identical, but marginally more idiomatic.

Refines: fbeb2065b180e2376088387ee150d3975dc08cd5

cc: @javorszky

For reference, Node's console docs are here: https://nodejs.org/docs/latest-v20.x/api/console.html

`console.warn` is an alias for `console.error`, and `console.error` is documented as: "Prints to stderr with newline." The `console` global has been present since the very earliest releases of Node.

### DIFF
--- a/src/nodejs/unit-http/http_server.js
+++ b/src/nodejs/unit-http/http_server.js
@@ -5,7 +5,6 @@
 
 'use strict';
 
-const { stderr } = require('process');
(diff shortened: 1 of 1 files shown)


javorszky (approved):
this makes a lot more sense here, thank you!

PR #1127, opened 2024-02-14
Bump docker image language runtimes

thresheek:

PR #1137, opened 2024-02-19
Docker: Switch python3.12 to using github

andrey-zelenkov:
Forgotten in
<https://github.com/nginx/unit/commit/c3af21e970ca3c822004cfda7c5b56ec07d99da9>

### DIFF
--- a/pkg/docker/Dockerfile.python3.12
+++ b/pkg/docker/Dockerfile.python3.12
@@ -11,11 +11,11 @@ LABEL org.opencontainers.image.version="1.31.1"
 RUN set -ex \
     && savedAptMark="$(apt-mark showmanual)" \
     && apt-get update \
-    && apt-get install --no-install-recommends --no-install-suggests -y ca-certificates mercurial build-essential libssl-dev libpcre2-dev curl pkg-config \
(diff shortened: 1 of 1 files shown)


callahad (approved):

callahad:
> Forgotten in c3af21e970ca3c822004cfda7c5b56ec07d99da9

Not so much _forgotten..._ It's more the product of out-of-order merges.

#1050 added python 3.12
#1131 switched the Dockerfiles to git

...but #1131 was based on `master` _before_ #1050 was merged, so its regenerated the Dockerfiles didn't include 3.12.

We're actually in an inconsistent state right now, because https://github.com/nginx/unit/commit/822303e23cb489efdc4fa3ca321f8468a2dd17fa effectively bumped NXT_VERSION to 1.32.0, which is used in `make dockerfiles`, so you either have to edit that or override the variable when regenerating the Dockerfiles locally.

And we're likely to have more inconsistencies once we merge #1127.

I wonder if the answer is to stop tracking the generated Dockerfiles in our repo?

callahad:
Oof. Thank you for the citation. So I guess the next best thing is a CI job that verifies that regenerating our Dockerfiles is a no-op. Will file an issue to think this through later.

PR #1143, opened 2024-02-20
Update wasmtime to v18

arbourd:
Updates wasmtime to latest major release `18.0.0` as per: https://github.com/nginx/unit/pull/1123#discussion_r1496240064


arbourd:
I think waiting on the CI stuff to see if it still builds is a good idea.

thresheek:
You also need to fix the checksum in SHA512SUMS.

We use the contrib version to build our packages for all supported distros and docker images, so makes sense to also test if the builds don't fail (we don't have a BB integration for that).

callahad:
I _think_ we should skip this update... this bundled copy of wasmtime is currently only used by our bespoke libunit-wasm stuff, which is likely an evolutionary dead end over the long run as WASI progresses...

PR #1147, opened 2024-02-22
Packaging for wasm wasi component module

thresheek:
Mostly mirrors the method used in the dockerfiles

Tested on RHEL 8, RHEL 9, Amazon Linux 2, Ubuntu 20.04, Ubuntu 22.04, Ubuntu 23.10.

Compared to `wasm` module, for `wasm-wasi-component` wasmtime and dependencies are built twice: once for debug and once for non-debug build. 
Also, both resulting module shared objects ship their own copy of wasmtime, making it three copies of wasmtime (one 11.0.1 and two 17.0.1) shipped in the package.


ac000 (approved):
LGTM

PR #1151, opened 2024-02-22
Remove debug from builds and tests

arbourd:
The info and above errors should be more than enough for debugging
failures in GitHuB Actions CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,8 +136,7 @@ jobs:
             --openssl \
             --njs \
             --cc-opt="-I njs/src/ -I njs/build"  \
-            --ld-opt="-L njs/build" \
-            --debug
+            --ld-opt="-L njs/build"
 
       - name: Make unit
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -136,8 +140,7 @@ jobs:
             --openssl \
             --njs \
             --cc-opt="-I njs/src/ -I njs/build"  \
-            --ld-opt="-L njs/build" \
-            --debug
+            --ld-opt="-L njs/build"
 
       - name: Make unit
         run: |
```
ac000, line 146: Looks good. But perhaps this should be split out into own commit separate from the language updates...

arbourd: I'm messing around with it to force a failure, but ya I agree! I'll undraft this when it's good to go.

ac000:
LGTM

ac000 (approved):

PR #1154, opened 2024-02-23
Clang static analyser fixes

ac000:
These commits fix a couple of issues found by the clang static analyser.

- Var: Remove a dead assignment in nxt_var_interpreter()

The first removes a dead assignment whereby we store the return from a nxt_cpymem() bu then return from the function before doing anything further with it.

- Avoid potential NULL pointer dereference in nxt_router_temp_conf()

The second fixes a potential NULL pointer dereference.

### DIFF
--- a/src/nxt_router.c
+++ b/src/nxt_router.c
@@ -1077,14 +1077,14 @@ nxt_router_temp_conf(nxt_task_t *task)
 
     rtcf = nxt_mp_zget(mp, sizeof(nxt_router_conf_t));
     if (nxt_slow_path(rtcf == NULL)) {
-        goto fail;
+        goto out_free_mp;
     }
 
     rtcf->mem_pool = mp;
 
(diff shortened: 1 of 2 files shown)


hongzhidao (approved):
LGTM.

And just another way.
```
    if (rtcf != NULL && rtcf->tstr_state != NULL) {
        nxt_tstr_state_release(rtcf->tstr_state);
    }
```

Both are ok, fix it in your preferred way.

ac000:
> LGTM.
> 
> And just another way.
> 
> ```
>     if (rtcf != NULL && rtcf->tstr_state != NULL) {
>         nxt_tstr_state_release(rtcf->tstr_state);
>     }
> ```
> 
> Both are ok, fix it in your preferred way.

Yes, that also works. However I like the overall goto label clean up which i think improves the whole function...

PR #1173, opened 2024-03-05
Test with root access in GitHub workflows

arbourd:
To enable tests that require privileged root access, this commit tests with `sudo`. The Java and Python jobs have additional permissions issues, so they are also configured and made with `sudo`.

A small permissions fix is required before running tests to allow non-root users to execute within the `/home/runner` directory.

This change also removes the custom directories that were required without root access.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,22 +53,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      # Creates and outputs directories used by tests (/usr/local is unfriendly)
-      - name: Configure directories
-        id: dir
-        run: |
-          PREFIX=${HOME}/.unit
-          BIN=${PREFIX}/bin
-          VAR=${PREFIX}/var
-          mkdir -p $BIN
-          mkdir -p $VAR
-
-          echo "prefix=${PREFIX}" >> "$GITHUB_OUTPUT"
-          echo "bin=${BIN}" >> "$GITHUB_OUTPUT"
-          echo "bin=${BIN}" >> "$GITHUB_PATH"
-          echo "var=${VAR}" >> "$GITHUB_OUTPUT"
-          cat "$GITHUB_OUTPUT"
-
       # Provides module, language version and testpath from build name
       - name: Output build metadata
         id: metadata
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -126,16 +110,7 @@ jobs:
...
-            --log=${{ steps.dir.outputs.var }}/log/unit/unit.log \
-            --runstatedir=${{ steps.dir.outputs.var }}/run \
-            --pid=${{ steps.dir.outputs.var }}/run/unit/unit.pid \
-            --control=unix:${{ steps.dir.outputs.var }}/run/unit/control.sock \
-            --modules=${{ steps.dir.outputs.prefix }}/lib/unit/modules \
-            --statedir=${{ steps.dir.outputs.var }}/state/unit \
+          sudo ./configure \
             --tests \
             --openssl \
             --njs \
```
ac000, line 116: Right, Unit doesn't have to be installed to run the pytests...

.github/workflows/ci.yml
```diff
@@ -309,18 +284,25 @@ jobs:
...
       ## Tests
       ##
 
+      # /home/runner will be root only after calling sudo above
+      # Ensure all users and processes can execute
+      - name: Fix permissions
+        run: |
+          sudo chmod -R +x /home/runner
+          namei -l ${{ github.workspace }}
+
```
ac000, line 305: Heh, you passed the test and didn't just chmod 777 the whole thing!

ac000 (commented):
A more general question is do we need to do all the configuring and compiling as root?

I know this is just a throwaway environment, but from a principle of least privilege, is it really only the actual pytests that need to be run as root?

arbourd:
Re: @ac000 

> A more general question is do we need to do all the configuring and compiling as root?
>
> I know this is just a throwaway environment, but from a principle of least privilege, is it really only the actual pytests that need to be run as root?

This is actually how I started, but, there are issues with the Java and Python tests: for whatever reasons there are issues with requiring `sudo` either during the configure or make steps. I am happy to try and make it work though, if we want to try and use `sudo` as little as possible.

.github/workflows/ci.yml
arbourd: I think these have defaults as well in `/usr` yes?

ac000: Yes, `/usr/local`

ac000:
Even if we can do _most_ of it without sudo and only use sudo for the bits that currently require it...

arbourd:
@ac000 All done. I will rebase and clean up the commits before merging. 2nd commit is the specific-sudoing.

ac000:
I _would_ like to see the final result before mering...

ac000 (approved):
OK, looks good. Thanks!

PR #1174, opened 2024-03-05
Add dependabot.yml

arbourd:
We already [use dependabot for security related patches](https://github.com/nginx/unit/pull/1170), by default.

This change adds a dependabot.yml configuration file that explicitly enables the service to manage versions of Actions in GitHub Actions. This ensures that Actions like `setup-go` are updated timely.

This change does not affect how Dependabot manages versions for Go, Rust, etc. The file can be used to configure that for additional package managers and languages in the future, if desired.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: daily


ac000:
So how does this actually look like?

[This](https://github.com/nginx/unit/pull/1174)?

As long as it won't be doing something funky like automatically committing changes to the repository...

arbourd:
It wouldn't update the repository directly @ac000, but it will open a PR. It would look something like this: https://github.com/arbourd/concourse-slack-alert-resource/pull/115

ac000 (approved):
OK, let's see how this goes...

PR #1175, opened 2024-03-06
Add GitHub workflow for wasm-wasi-component

ac000:
This adds a GitHub CI workflow for the new wasm-wasi-component language module.
    
Some things of note.
    
1) We need to special case 'wasm-wasi-component' in the 'Output build
       metadata' section as we are splitting the module names on '-' to
       split them into name and version.
    
2) Apart from needing to tell bindgen about the njs include paths, we
       also need to explicitly specify which version of clang to use to
       work around an [issue](https://gitlab.freedesktop.org/mesa/mesa/-/issues/7268) with multiple versions of clang installed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,8 @@ jobs:
             os: ubuntu-latest
           - build: wasm
             os: ubuntu-latest
+          - build: wasm-wasi-component
+            os: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v4
@@ -57,8 +59,12 @@ jobs:
       - name: Output build metadata
         id: metadata
         run: |
-          # Split the build name by '-' into module and version
-          IFS='-' read -r module version <<< "${{ matrix.build }}"
+          if [ "${{ matrix.build }}" = "wasm-wasi-component" ]; then
+            module="wasm-wasi-component"
+          else
+            # Split the build name by '-' into module and version
+            IFS='-' read -r module version <<< "${{ matrix.build }}"
+          fi
 
           testpath="test/test_${module}*"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -317,6 +323,27 @@ jobs:
...
+        run: |
+          ./configure wasm-wasi-component
+        if: steps.metadata.outputs.module == 'wasm-wasi-component'
+
+      - name: Make wasm-wasi-component
+        run: |
+          CLANG_PATH=/usr/bin/clang-15 \
+          BINDGEN_EXTRA_CLANG_ARGS="-I../../njs/src -I../../njs/build" \
+          make wasm-wasi-component
+        if: steps.metadata.outputs.module == 'wasm-wasi-component'
```
arbourd, line 345: Does it make sense just to include this in the `wasm` step?

.github/workflows/ci.yml
```diff
@@ -325,15 +352,18 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: '3'
-        if: steps.metadata.outputs.module != 'wasm'
+        if: steps.metadata.outputs.module != 'wasm' &&
+            steps.metadata.outputs.module != 'wasm-wasi-component'
```
arbourd, line 356: You could also do

```suggestion
        if: startsWith(steps.metadata.outputs.module, 'wasm')
```

`startsWith` is a [builtin](https://docs.github.com/en/actions/learn-github-actions/expressions#startswith)

.github/workflows/ci.yml
```diff
@@ -73,8 +75,12 @@ jobs:
...
         id: metadata
         run: |
-          # Split the build name by '-' into module and version
-          IFS='-' read -r module version <<< "${{ matrix.build }}"
+          if [ "${{ matrix.build }}" = "wasm-wasi-component" ]; then
+            module="wasm-wasi-component"
+          else
+            # Split the build name by '-' into module and version
+            IFS='-' read -r module version <<< "${{ matrix.build }}"
+          fi
```
arbourd, line 83: I think we'll need to improve upon my hacky `language-version` idea eventually, but this works for now! 😄

arbourd (approved):

.github/workflows/ci.yml
ac000: I did at first, but then thought it was sufficiently different...

arbourd: Cool! Looks to good to me 👍🏻

ac000:
- Rebased with master

```
$ git range-diff 16bd12b3...0cee7d1a
-:  -------- > 1:  8032ce31 Test with root access in GitHub workflows
1:  16bd12b3 ! 2:  0cee7d1a Add GitHub workflow for wasm-wasi-component
    @@ .github/workflows/ci.yml: jobs:
      
            - name: Install pytest
              run: |
    -           pip install pytest
    +           sudo -H pip install pytest
     -        if: steps.metadata.outputs.module != 'wasm'
     +        if: steps.metadata.outputs.module != 'wasm' &&
     +            steps.metadata.outputs.module != 'wasm-wasi-component'
      
            - name: Run ${{ steps.metadata.outputs.module }} tests
              run: |
    -           pytest --print-log ${{ steps.metadata.outputs.testpath }}
    +           sudo -E pytest --print-log ${{ steps.metadata.outputs.testpath }}
              # Skip pytest if wasm build, as there are no tests yet
     -        if: steps.metadata.outputs.module != 'wasm'
     +        if: steps.metadata.outputs.module != 'wasm' &&
```

PR #1180, opened 2024-03-07
.mailmap: Map Dylan's 2nd GitHub address

arbourd:
I changed a setting and now GitHub will recognize both the legacy numberless version, and the newer version with UserID.

The with-UserID version will be used by the any changes stemming from the GitHub GUI.


ac000 (commented):
LGTM. You can add my `Reviewed-by: Andrew Clayton <a.clayton@nginx.com>` (Goes above your s-o-b).

PR #1184, opened 2024-03-09
Fix RUSTSEC-2024-0019

rex4539:
https://rustsec.org/advisories/RUSTSEC-2024-0019


ac000:
Thanks for the PR, however I believe this is the same as https://github.com/nginx/unit/pull/1170

However this is a Windows only issue and so doesn't effect Unit... so there was no mad rush and we were having some internal discussions about the use of dependabot...

PR #1188, opened 2024-03-11
Tests: NJS cacheable variables with access log

andrey-zelenkov:
Reproduces issue https://github.com/nginx/unit/issues/1169.

### DIFF
--- a/test/test_njs.py
+++ b/test/test_njs.py
@@ -116,6 +116,24 @@ def test_njs_variables_cacheable(temp_dir):
     check_rewrite('/str', '${vars.uri}')
 
 
+def test_njs_variables_cacheable_access_log(findall, temp_dir):
+    assert 'success' in client.conf({"return": 200}, 'routes/0/action')
+
+    assert 'success' in client.conf(
+        {
+            'path': f'{temp_dir}/access.log',
+            'format': '`${vars.host}, ${vars.status}\n`',
+        },
+        'access_log'
+    ), 'access_log configure'
+
+    reqs = 50
+    for _ in range(reqs):
+        client.get()
+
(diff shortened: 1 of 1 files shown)


andrey-zelenkov:
Rebased and removed unused iterator.

```
% git range-diff afe8b31d...287fb378
 -:  -------- >  1:  0d99744d Router: match when pattern and tested string are both zero length
 -:  -------- >  2:  fdc46759 NJS: avoiding arithmetic ops with NULL pointer in r->args
 -:  -------- >  3:  8844d33c Fixed undefined behaviour in left shift of int value
 -:  -------- >  4:  7dcd6c0e Avoiding arithmetic ops with NULL pointer in nxt_http_arguments_parse
 -:  -------- >  5:  264b3755 Avoiding arithmetic ops with NULL pointer in nxt_port_mmap_get
 -:  -------- >  6:  c9461a6b Initialize port_impl only when it is needed
 -:  -------- >  7:  dd701fb4 Avoiding arithmetic ops with NULL pointer in nxt_unit_mmap_get
 -:  -------- >  8:  abcfc4cd Fix the security-alert email link in the README
 -:  -------- >  9:  9993814d NJS: loader should be registered using njs_vm_set_module_loader()
 1:  afe8b31d ! 10:  287fb378 Tests: NJS cacheable variables with access log
    @@ test/test_njs.py: def test_njs_variables_cacheable(temp_dir):
     +    ), 'access_log configure'
     +
     +    reqs = 50
    -+    for i in range(reqs):
    ++    for _ in range(reqs):
     +        client.get()
     +
     +    assert len(findall(r'localhost, 200', 'access.log')) == reqs
  ```

PR #1203, opened 2024-03-27
Merged with the 1.32 branch

andrey-zelenkov:


ac000:
I don't think this is right!

As has been previously said, what wasn't applied to 1.32.1 is 1.33.0 material...

ac000 (changes requested):
Putting a block on here as a safety net...

callahad:
Yeah, looks like the merge is just reversed; this is going master -> 1.32, instead of 1.32 -> master.

ac000:
> Yeah, looks like the merge is just reversed; this is going master ->
> 1.32, instead of 1.32 -> master.

But that's also wrong.

We'd end up with duplicate commits like we did once before...

I'm not sure why anything is needed merged with anything else at this
stage...

master and 1.32 are now distinct branches.

andrey-zelenkov:
I was hoping that the green button would understand what I want and do all work (I was going to send PR with merge 1.32 to master as @callahad assumed) but it didn't happen. So, anyway this PR is incorrect and should be closed. Will send another PR shortly.

PR #1211, opened 2024-04-04
Tests: added $request_uri tests with proxy

andrey-zelenkov:


test/test_variables.py
```diff
@@ -108,6 +108,51 @@ def check_request_uri(req_uri):
...
+            "access_log": {
+                "path": f'{temp_dir}/access.log',
+                "format": "$remote_addr $uri $request_uri",
+            },
+        }
+    )
+
+    assert search_in_file(r'::1', 'access.log') is None
+
+    assert client.get(url='/blah%2Fblah?a=b')['status'] == 200
```
hongzhidao, line 146: Looks good.
But I can't find this type of test for normal proxy. It's useful to add such uri test without "rewrite", for example:
```
"action": {
    "proxy": "http://[::1]:8081",
}

assert client.get(url='/blah%2Fblah?a=b')['status'] == 200

# search /blah%2Fblah?a=b
assert (
        wait_for_record(fr'...', 'access.log') is not None
    ), 'req 8081 (proxy)'
```

hongzhidao:
About the title, there is no `$rewrite_uri` variable in Unit.

test/test_variables.py
```diff
@@ -108,6 +108,51 @@ def check_request_uri(req_uri):
...
+        {
+            "listeners": {
+                "*:8080": {"pass": "routes/a"},
+                "[::1]:8081": {"pass": "routes/b"},
+            },
+            "routes": {
+                "a": [
+                    {
+                        "action": {
+                            "rewrite": "/foo",
```
hongzhidao, line 123: I feel it's necessary to test URI like `/foo%2Ffoo. For example:
```
"rewrite": "/foo%2Ffoo",
```

andrey-zelenkov: Added

andrey-zelenkov:
Patch reworked after @hongzhidao's comment and renamed. Commit message updated accordingly.

hongzhidao (approved):

PR #1214, opened 2024-04-09
HTTP: Refactoring nxt_h1p_peer_header_send()

hongzhidao:
Previously, proxy request was constructed based on the `r->target` field. However, r->target will remain unchanged in the future, even in cases of URL rewriting.
To accommodate this, I refactored the handling of the proxy target.

### DIFF
--- a/src/nxt_h1proto.c
+++ b/src/nxt_h1proto.c
@@ -90,6 +90,8 @@ static void nxt_h1p_peer_connect(nxt_task_t *task, nxt_http_peer_t *peer);
 static void nxt_h1p_peer_connected(nxt_task_t *task, void *obj, void *data);
 static void nxt_h1p_peer_refused(nxt_task_t *task, void *obj, void *data);
 static void nxt_h1p_peer_header_send(nxt_task_t *task, nxt_http_peer_t *peer);
+static nxt_int_t nxt_h1p_peer_request_target(nxt_http_request_t *r,
+    nxt_str_t *target);
 static void nxt_h1p_peer_header_sent(nxt_task_t *task, void *obj, void *data);
 static void nxt_h1p_peer_header_read(nxt_task_t *task, nxt_http_peer_t *peer);
 static ssize_t nxt_h1p_peer_io_read_handler(nxt_task_t *task, nxt_conn_t *c);
@@ -653,6 +655,8 @@ nxt_h1p_header_process(nxt_task_t *task, nxt_h1proto_t *h1p,
 
     r->target.start = h1p->parser.target_start;
     r->target.length = h1p->parser.target_end - h1p->parser.target_start;
+
+    r->quoted_target = h1p->parser.quoted_target;
 
     if (h1p->parser.version.ui64 != 0) {
         r->version.start = h1p->parser.version.str;
@@ -2263,6 +2267,8 @@ nxt_h1p_peer_header_send(nxt_task_t *task, nxt_http_peer_t *peer)
 {
     u_char              *p;
     size_t              size;
+    nxt_int_t           ret;
+    nxt_str_t           target;
     nxt_buf_t           *header, *body;
     nxt_conn_t          *c;
     nxt_http_field_t    *field;
@@ -2272,7 +2278,12 @@ nxt_h1p_peer_header_send(nxt_task_t *task, nxt_http_peer_t *peer)
 
     r = peer->request;
 
-    size = r->method->length + sizeof(" ") + r->target.length
+    ret = nxt_h1p_peer_request_target(r, &target);
+    if (nxt_slow_path(ret != NXT_OK)) {
+        goto fail;
+    }
+
+    size = r->method->length + sizeof(" ") + target.length
            + sizeof(" HTTP/1.1\r\n")
            + sizeof("Connection: close\r\n")
            + sizeof("\r\n");
@@ -2288,8 +2299,7 @@ nxt_h1p_peer_header_send(nxt_task_t *task, nxt_http_peer_t *peer)
 
     header = nxt_http_buf_mem(task, r, size);
     if (nxt_slow_path(header == NULL)) {
-        r->state->error_handler(task, r, peer);
-        return;
+        goto fail;
     }
 
     p = header->mem.free;
@@ -2328,8 +2338,7 @@ nxt_h1p_peer_header_send(nxt_task_t *task, nxt_http_peer_t *peer)
         }
 
         if (nxt_slow_path(body == NULL)) {
-            r->state->error_handler(task, r, peer);
-            return;
+            goto fail;
         }
 
         header->next = body;
@@ -2353,6 +2362,61 @@ nxt_h1p_peer_header_send(nxt_task_t *task, nxt_http_peer_t *peer)
     }
 
     nxt_conn_write(task->thread->engine, c);
+
+    return;
+
+fail:
+
+    r->state->error_handler(task, r, peer);
+}
+
+
+static nxt_int_t
+nxt_h1p_peer_request_target(nxt_http_request_t *r, nxt_str_t *target)
+{
+    u_char  *p;
+    size_t  size, encode;
+
+    if (!r->uri_changed) {
+        *target = r->target;
+        return NXT_OK;
+    }
+
+    if (!r->quoted_target && r->args->length == 0) {
+        *target = *r->path;
+        return NXT_OK;
+    }
+
+    if (r->quoted_target) {
+        encode = nxt_encode_complex_uri(NULL, r->path->start,
+                                        r->path->length);
+    } else {
+        encode = 0;
+    }
+
+    size = r->path->length + encode * 2 + 1 + r->args->length;
+
+    target->start = nxt_mp_nget(r->mem_pool, size);
+    if (target->start == NULL) {
+        return NXT_ERROR;
+    }
+
+    if (r->quoted_target) {
+        p = (u_char *) nxt_encode_complex_uri(target->start, r->path->start,
+                                              r->path->length);
+
+    } else {
+        p = nxt_cpymem(target->start, r->path->start, r->path->length);
+    }
+
+    if (r->args->length > 0) {
+        *p++ = '?';
+        p = nxt_cpymem(p, r->args->start, r->args->length);
+    }
+
+    target->length = p - target->start;
+
+    return NXT_OK;
 }
 
 
--- a/src/nxt_http_parse.h
+++ b/src/nxt_http_parse.h
@@ -61,9 +61,9 @@ struct nxt_http_request_parse_s {
 
     /* target with "/." */
     uint8_t                   complex_target;         /* 1 bit */
-#if 0
     /* target with "%" */
     uint8_t                   quoted_target;          /* 1 bit */
+#if 0
     /* target with " " */
     uint8_t                   space_in_target;        /* 1 bit */
 #endif
--- a/src/nxt_http.h
+++ b/src/nxt_http.h
@@ -192,6 +192,8 @@ struct nxt_http_request_s {
     nxt_http_status_t               status:16;
 
     uint8_t                         log_route;    /* 1 bit */
+    uint8_t                         quoted_target;  /* 1 bit */
+    uint8_t                         uri_changed;  /* 1 bit */
 
     uint8_t                         pass_count;   /* 8 bits */
     uint8_t                         app_target;
--- a/src/nxt_http_parse.c
+++ b/src/nxt_http_parse.c
@@ -286,13 +286,11 @@ continue_target:
         case NXT_HTTP_TARGET_SPACE:
             rp->target_end = p;
             goto space_after_target;
-#if 0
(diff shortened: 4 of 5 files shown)


ac000 (commented):
Hi Zhidao,

> Previously, proxy request was constructed based on the `r->target`
field. However, r->target will remain unchanged in the future,
even in cases of URL rewriting.
To accommodate this, I refactored the handling of the proxy target.

This doesn't really tell me _why_ `r->target` remaining unchanged will be an issue.

src/nxt_h1proto.c
```diff
@@ -90,6 +90,8 @@ static void nxt_h1p_peer_connect(nxt_task_t *task, nxt_http_peer_t *peer);
 static void nxt_h1p_peer_connected(nxt_task_t *task, void *obj, void *data);
 static void nxt_h1p_peer_refused(nxt_task_t *task, void *obj, void *data);
 static void nxt_h1p_peer_header_send(nxt_task_t *task, nxt_http_peer_t *peer);
+static nxt_int_t nxt_h1p_peer_request_target(nxt_http_request_t *r,
+    nxt_str_t *target);
```
ac000, line 94: This change seems unrelated to the rest. I.e I don't see you adding this function for example.

Does adding this get rid of some warning introduced by this commit?

src/nxt_h1proto.c
```diff
@@ -654,6 +656,8 @@ nxt_h1p_header_process(nxt_task_t *task, nxt_h1proto_t *h1p,
     r->target.start = h1p->parser.target_start;
     r->target.length = h1p->parser.target_end - h1p->parser.target_start;
 
+    r->quoted_target = h1p->parser.quoted_target;
+
```
ac000, line 660: What does `quoted_target` mean?

hongzhidao:
Hi Andrew,
For now, `r->target` is changed during URL rewriting. It will be constant because we need to keep `$request_uri` unchanged. Here's the change.
https://github.com/nginx/unit/pull/1162/files#diff-8f22b5e7ac8aa80082722fd822e96917284058f7aa449aeffec406a5e0520beeL80 
Let me know if it's necessary to add something like above to the commit log.

src/nxt_h1proto.c
```diff
@@ -2353,6 +2362,61 @@ nxt_h1p_peer_header_send(nxt_task_t *task, nxt_http_peer_t *peer)
...
+    size = r->path->length + encode * 2 + 1 + r->args->length;
+
+    target->start = nxt_mp_nget(r->mem_pool, size);
+    if (target->start == NULL) {
+        return NXT_ERROR;
+    }
+
+    if (r->quoted_target) {
+        p = (u_char *) nxt_encode_complex_uri(target->start, r->path->start,
+                                              r->path->length);
```
ac000, line 2406: This begged the question, will a `size_t` always be big enough to hold a `uintptr_t`.

While I don't believe the C standard makes any guarantees that that will be the case, I think in practice it will be, at least for any system we care about.

On 64bit Linux (and macOS) they are defined like

```
$ echo | gcc -E -xc -include 'stddef.h' - | grep size_t
typedef long unsigned int size_t;
$ echo | gcc -E -xc -include 'stdint.h' - | grep uintptr_t
typedef unsigned long int uintptr_t;
```

On 32bit Linux

```
$ echo | gcc -E -xc -include 'stddef.h' - | grep size_t
typedef unsigned int size_t;
$ echo | gcc -E -xc -include 'stdint.h' - | grep uintptr_t
typedef unsigned int uintptr_t;
```

src/nxt_http_parse.h
```diff
@@ -61,12 +61,10 @@ struct nxt_http_request_parse_s {
 
     /* target with "/." */
     uint8_t                   complex_target;         /* 1 bit */
-#if 0
     /* target with "%" */
     uint8_t                   quoted_target;          /* 1 bit */
     /* target with " " */
     uint8_t                   space_in_target;        /* 1 bit */
-#endif
```
ac000, line 69: Curious why we are only now enabling these two things?

ac000:
> For now, `r->target` is changed during URL rewriting. It will be
> constant because we need to keep `$request_uri` unchanged. Here's the
> change.

Yes, but I'm more asking _why_ it remaining unchanged will be an issue
for the proxy going forward.

hongzhidao:
> This doesn't really tell me why r->target remaining unchanged will be an issue.
Yes, but I'm more asking why it remaining unchanged will be an issue
for the proxy going forward.

There is no issue currently for proxy.

Ideally, the `$request_uri` will always correspond to `r->target` which happens in `nxt_http_variables.c`.
During URL rewriting, `r->target` is changed, and the same is for `$request_uri`.
We decided to make `$request_uri` constant, which means `r->target` will also be unchanged, even in case of URL rewriting. The logic of touching `r->target` will be removed.
https://github.com/nginx/unit/pull/1162/files#diff-8f22b5e7ac8aa80082722fd822e96917284058f7aa449aeffec406a5e0520beeL80

To make it short, the `r->target` should be designed to be constant, but Unit needs to pass a changeable URL to upstream server in proxy module. So proxy can't depend on `r->target`.

About the quoted_target, take the following as an example.
```
client request: /blah%25blah?a=b
r->path: /blah%blah
r->target:/blah%2Fblah?a=b
```
`quoted_target` means it contains percent-encoding char like `%25`.
If it's quoted, it means we need to encode `r->path`.
And we also need to take into arguments account, if there is argument, we need to append it.
The logic is something like this:
```
if (quoted_target || args->length > 0) {
    we need to generate a new URL.
}
```

src/nxt_http_parse.h
hongzhidao: See my new comment, but it looks like `space_in_target` is unused. Will check.

src/nxt_h1proto.c
hongzhidao: See my new comment.
It happens during http parsing.

src/nxt_h1proto.c
hongzhidao: Sorry, I don't understand what you mean.
`nxt_h1p_peer_request_target()` is called in `nxt_h1p_peer_header_send()`.

ac000: Ah, I see it.

<rant>
If ordered our functions correctly we wouldn't need all these prototypes...
</rant>

ac000:
> To make it short, the `r->target` should be designed to be constant,
> but Unit needs to pass a changeable URL to upstream server in proxy
> module. So proxy can't depend on `r->target`.

Right this is what I was after.

So, the proxy needs to see the re-written URL... gotcha.

That's a key bit of information missing from the commit message, please
add it.

ac000:
> About the quoted_target, take the following as an example.

> ```
> client request: /blah%25blah?a=b
> r->path: /blah%blah
> r->target:/blah%2Fblah?a=b
> ```

> quoted_target means it contains percent-encoding char like `%25`.  If
> it's quoted, it means we need to encode `r->path`.  And we also need
> to take into arguments account, if there is argument, we need to
> append it.

OK, so _quoted_target_ means _percent_encoded_. It's just badly named.

hongzhidao:
I think it's just a refactoring patch, there are no functional changes introduced.
When proxy was introduced, there is no rewrite feature at that time, it makes sense to use `r->target` which is equal to `r->path` + `r->args`. So what I did here is to use a more proper way of creating proxy sending header.

Btw, I'm thinking about separating quoted_target related code into a new commit.

src/nxt_h1proto.c
hongzhidao: I replaced it with `nxt_uint_t` that referred to `nxt_http_return.c`.
I feel both `nxt_uint_t` and `size_t` are correct since the max size is limited by the `large_header_buffer_size` option that is validated by the Unit JSON. And the max number length is 14.
```
#define NXT_CONF_MAX_NUMBER_LEN    14
```

src/nxt_h1proto.c
hongzhidao: Hmm, the other way is to put the function definition before the call, both the two ways are fine.

ac000: > Hmm, the other way is to put the function definition before the call, both the two ways are fine.

If you mean this, then yes...

```c
static void func_c(void)
{
        /* ... */
}

static void func_b(void)
{
        func_c();
}

static void func_a(void)
{
        func_b();
}

int main(void)
{
        func_a();

        return 0;
}
```

Following that pattern you rarely need to provide prototypes as the compiler will see the function definition _before_ it's used.

ac000 (commented):
**EDIT:** This is in relation to [HTTP: Introduce quoted target marker in HTTP parsing](https://github.com/nginx/unit/pull/1214/commits/28ed5912fcd2f55eb7dd07e4b77e34a4000fa6e4)

Perhaps a bit of text in the commit message about _why_ we need this now and we didn't need it before...

src/nxt_h1proto.c
ac000: That makes me more nervous! What if that constant ever increases... I think it was **safer** with `size_t`.

ac000:
> I think it's just a refactoring patch, there are no functional changes
> introduced.

But it's making use of the `quoted_target` thing which it wasn't before.
Does that not equate to a change in behaviour?

src/nxt_h1proto.c
hongzhidao: I kept `size_t`.

hongzhidao:
> **EDIT:** This is in relation to [HTTP: Introduce quoted target marker in HTTP parsing](https://github.com/nginx/unit/pull/1214/commits/28ed5912fcd2f55eb7dd07e4b77e34a4000fa6e4)
> 
> Perhaps a bit of text in the commit message about _why_ we need this now and we didn't need it before...

I don't know exactly why, because this part of the code was commented out, probably because it wasn't needed at the time.
But I added a note about when it can be used, that's its intention.

hongzhidao:
> **EDIT:** This is in relation to [HTTP: Introduce quoted target marker in HTTP parsing](https://github.com/nginx/unit/pull/1214/commits/28ed5912fcd2f55eb7dd07e4b77e34a4000fa6e4)
> 
> Perhaps a bit of text in the commit message about _why_ we need this now and we didn't need it before...

I don't know exactly why, because this part of the code was commented out, probably because it wasn't needed at the time.
But I added a note about when it can be used, that's its intention.

hongzhidao:
> > I think it's just a refactoring patch, there are no functional changes
> > introduced.
> 
> But it's making use of the `quoted_target` thing which it wasn't before. Does that not equate to a change in behaviour?

I split it into two commits, for the first one, maybe it's a code change rather than refactoring.
And I think it's a refactoring for the second commit.

ac000:
For me the second patch ((HTTP: Refactoring nxt_h1p_peer_header_send()](https://github.com/nginx/unit/pull/1214/commits/95727a1bae2224b1aa1a2c245583a1429eb14342)) is not a refactoring.

I would either split it up to do the refactor bit, i.e without the `quoted_target` stuff, then a second patch making use of the `quoted_target` stuff.

We didn't use `quoted_target` before, a _refactor_ should not suddenly start using it I think...

Or just admit it isn't a refactor and adjust the commit message...

hongzhidao:
> Or just admit it isn't a refactor and adjust the commit message...

Ok, the same for the first one. Removed "No functional changes".

src/nxt_h1proto.c
```diff
@@ -2353,6 +2362,61 @@ nxt_h1p_peer_header_send(nxt_task_t *task, nxt_http_peer_t *peer)
...
+
+    if (r->quoted_target) {
+        encode = nxt_encode_complex_uri(NULL, r->path->start,
+                                        r->path->length);
+    } else {
+        encode = 0;
+    }
+
+    size = r->path->length + encode * 2 + 1 + r->args->length;
+
```
ac000, line 2398: Just want to confirm that the required calculation is (taking operator precedence into account) is

```c
size = r->path->length + (encode * 2) + 1 + r->args->length;
```

e.g

```
5 + (7*2) + 1 + 8 = 28
```

hongzhidao: Yes, I did.

ac000: Final question about this.

Why the * 2?

hongzhidao: According to its implementation.
https://github.com/nginx/unit/blob/master/src/nxt_string.c#L607
Take `/a b c` as an example, the allocation length is `r->path->length` + `encoded_number * 2`.
```
/a b c => /a%20b%20c
r->path: "/a b c"
encoded chars: ' ', ' '
total: 6 + 2 * 2
```

ac000: Got it.

r->path->length already includes the to be encoded chars, and each to be encoded char adds two bytes.

ac000 (approved):

src/nxt_h1proto.c
hongzhidao: Correct.

PR #1218, opened 2024-04-16
Add key points to Readme.

avahahn:
I noted that our Docker instructions don't provide a complete quickstart, and leaves the user thinking about what additional steps need to be taken to host their application. While this isn't too hard to figure out on it's own I find that just having the instruction there saves time otherwise spent looking at examples and finding/reading the Dockerfile.

Additionally, after stumbling on our deprecated WASM functionality before the supported component; I found it prudent to show our WASM documentation front and center. Hopefully this will spare potential users any confusion they get finding the wrong thing first.

Finally, I wanted to point out the configuration docs in their own section, because I missed the single inline link we have to them the first two times I read through our Readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ For details and available language packages, see the
 ### Docker
 
 ``` console
-$ docker pull unit
+$ docker pull unit:<TAG>
 $ mkdir /tmp/unit-control # customize as needed.
 $ docker run -d \
       --mount type=bind,src=/tmp/unit-control,dst=/var/run \
@@ -53,6 +53,8 @@ $ docker run -d \
 For a description of image tags, see the
 [docs](https://unit.nginx.org/installation/#docker-images).
 
+WARNING: latest image tag may not provide support for specific language modules,
+*do* check the available image tags from the link above before pulling your image.
 
 Your current working directory will now be mounted to the Unit image at `/www`.
 You can reach its socket at `/tmp/unit-control/control.unit.sock` assuming no 
--- a/README.md
+++ b/README.md
@@ -42,11 +42,21 @@ For details and available language packages, see the
 
 ``` console
 $ docker pull unit
+$ mkdir /tmp/unit-control # customize as needed.
+$ docker run -d \
+      --mount type=bind,src=/tmp/unit-control,dst=/var/run \
+      --mount type=bind,src=.,dst=/www \
+      --network host \
+      unit
 ```
 
 For a description of image tags, see the
 [docs](https://unit.nginx.org/installation/#docker-images).
 
+
+Your current working directory will now be mounted to the Unit image at `/www`.
+You can reach its socket at `/tmp/unit-control/control.unit.sock` assuming no 
+further customizations have been made.
 
 ### Amazon Linux, Fedora, Red Hat
 
@@ -71,6 +81,15 @@ $ wget https://raw.githubusercontent.com/nginx/unit/master/tools/setup-unit && c
 For details and available language packages, see the
 [docs](https://unit.nginx.org/installation/#official-packages).
 
+## Configuration
+
+NGINX Unit provides a RESTful API for dynamic configuration.
+See the [control API documentation](https://unit.nginx.org/controlapi/)
+for more information on what endpoints are available and how to use them.
+
+
+For full details of configuration management, see the
+[docs](https://unit.nginx.org/configuration/#configuration-management).
 
 ## Running a Hello World App
 
@@ -161,8 +180,10 @@ Unit's output should contain both snippets, neatly organized:
 }
 ```
 
-For full details of configuration management, see the
-[docs](https://unit.nginx.org/configuration/#configuration-management).
+## WebAssembly
+Unit supports running WebAssembly Components (WASI 0.2).
+For more information see the
+[Unit Configuration Docs](https://unit.nginx.org/configuration/#configuration-wasm).
 
 ## OpenAPI Specification
 


README.md
```diff
@@ -42,12 +42,21 @@ For details and available language packages, see the
...
+      --network host \
+      unit
 ```
 
 For a description of image tags, see the
 [docs](https://unit.nginx.org/installation/#docker-images).
 
 
+Your current working directory will now be mounted to the Unit image at `/www`.
+You can reach its socket at `/tmp/unit-control/control.unit.sock` assuming no further customizations have been made.
```
ac000, line 58: Probably wants a line-break after the `no`.

README.md
```diff
@@ -71,6 +80,14 @@ $ wget https://raw.githubusercontent.com/nginx/unit/master/tools/setup-unit && c
 For details and available language packages, see the
 [docs](https://unit.nginx.org/installation/#official-packages).
 
+## Configuration
+
+NGINX Unit provides a RESTful API for dynamic configuration.
+See the [control API documentation](https://unit.nginx.org/controlapi/) for more information on what endpoints are available and how to use them.
```
ac000, line 86: Line break after the URL...

README.md
```diff
@@ -161,8 +178,9 @@ Unit's output should contain both snippets, neatly organized:
 }
 ```
 
-For full details of configuration management, see the
-[docs](https://unit.nginx.org/configuration/#configuration-management).
+## WebAssembly
+Unit supports deploying WASM as a supported application.
+For more information see [the docs](https://unit.nginx.org/configuration/#configuration-wasm).
```
ac000, line 183: Line break after `see`. Or `the` and don't include it in the link (for consistency with the others)...

ac000 (changes requested):

ac000:
Hi Ava, just some general comments.

We try to keep commit subjects in the 50-70 character range (full documentation covering all this is coming!).

Perhaps you could make the subject `Add additional information to the README` or some such and use the existing subject as the body (keeping the body to 72 or less character lines).

Bonus points for the `s-o-b` though!

About the changes themselves, if you look closely, you'll notice that we try to keep the text in the README to < 80 character lines (except for things like URLs, commands and output etc).

README.md
Jcahilltorre: It's generally bad practice not to specify where a link goes:
❌: "go here" , "see the doc", "check this"
✔: "check the wasm configuration documentation", "check the Microsoft Azure docs", etc

I would change this to `For more information see the [Unit configuration docs]`

danielledeleo (approved):
Minor style concerns as mentioned by others aside, these changes are great. Thanks Ava!

README.md
```diff
@@ -161,8 +178,9 @@ Unit's output should contain both snippets, neatly organized:
 }
 ```
 
-For full details of configuration management, see the
-[docs](https://unit.nginx.org/configuration/#configuration-management).
+## WebAssembly
+Unit supports deploying WASM as a supported application.
```
lcrilly, line 182: 1. Above we talk about "running apps" but here we're talking about "deploying".
2. I'm not sure that calling out WebAssembly is warranted over all the other runtimes. I do like the idea of promoting it but retain a small reservation.
3. The abbreviation is Wasm so we should either avoid using two terms or introduce the abbreviation if we intend to use it again.
```suggestion
Unit supports running WebAssembly (Wasm) Components.
```

avahahn: About your second point:
I did deliberately call out WebAssembly where other languages are not. My rationale is articulated in the merge request description. My goal is not to promote WebAssembly, but to solve a discovery issue that we have.

Coming to the project, I encountered the unit-wasm repo first. I did not know it was deprecated. I was confused as to why it wasn't mentioned in our Readme since it is a separate repository. I did not realize we had a more up to date module in the Unit repository until someone told me.

Most people are going to find us through Github before they look at our docs site. I am willing to hazard a bet that people will find unit-wasm before they find our largely undocumented WASI module.

Perhaps a better solution is to update unit-wasm with a deprecation notice that redirects to whatever documentation we do have for our WASI module.

Additionally, I think it might be worth explicitly listing all of the languages we support (and probably separate quick starts for each) in the Readme; lest someone skimming our readme miss the one single line we have saying that we support multiple languages. They would come off thinking we only supported PHP.

Happy to either leave this here, or to go update unit-wasm. Please let me know which is preferrable.

ac000: I _am_ hesitant to explicitly mark the `wasm` language module as deprecated at the moment as it handles some cases that the new one doesn't. E.g ingress/egress of > 4GiB data... we also don't get the same amount of request information, well, not that I found in the C API anyway, see <https://gist.github.com/ac000/906707af37b42cdcc9d3cc4c2715da99#file-000-wasi-http-txt-L21>

lcrilly: Your observations are spot-on.

IMO, the job of the readme is to let people know if this is something they should pay attention to, and if so, to get them up and running very quickly. This PR is compatible with that :)

Deprecating unit-wasm and libunitwasm are equally important to that goal.

ac000: > Deprecating unit-wasm and libunitwasm are equally important to that goal.

See above.

I _could_ add a note to the unit-wasm README that points people to `wasm-wasi-component` _if_ they are happy with its caveats (and yes that includes requiring _all_ the rust stuff, or running in memory constrained environments)...

ac000: About the memory constraints.

When uploading a file and simply writing it to disk...

>Yes, this does mean that the wasm-wasi-component application process does end   
up consuming quite a bit of RAM. When trying to upload the 2GiB file, I clock   
it using about ~4GiB, though this does get freed afterwards.                    
                                                                                
> For comparison, the C language module consumes a steady ~70MiB.

This _may_ be fixable...

avahahn: > I could add a note to the unit-wasm README that points people to wasm-wasi-component if they are happy with its caveats (and yes that includes requiring all the rust stuff, or running in memory constrained environments)...

I think a note would be excellent. I am happy to add one myself. I don't think it's worth unsubstantiated fearmongering about Rust in memory constrained environments in that comment. That would only either confuse new users or potentially turn off people who are using Rust.

Also, in such environments would a user really be deploying a WebAssembly component?

avahahn: It sounds like these performance issues should be an issue, bug, or some other tracked work we can prioritize though.

ac000:
You can add my

```
Acked-by: Andrew Clayton <a.clayton@nginx.com>
```

README.md
```diff
@@ -41,12 +41,24 @@ For details and available language packages, see the
...
+      --mount type=bind,src=/tmp/unit-control,dst=/var/run \
+      --mount type=bind,src=.,dst=/www \
+      --network host \
+      unit
 ```
 
 For a description of image tags, see the
 [docs](https://unit.nginx.org/installation/#docker-images).
 
+WARNING: latest image tag may not provide support for apecific language models,
```
callahad, line 56:
```suggestion
WARNING: latest image tag may not provide support for specific language modules,
```

callahad (approved):
lgtm ✅

PR #1221, opened 2024-04-17
Pointer constification

ac000:
This patch series marks various pointer function arguments as `const` in the configuration sub-system.

The main impetus of this was to convert a bunch of

```c
static nxt_str_t ...
```

variable declarations to

```c
static const nxt_str_t ...
```

Having these _only_ `static` seems a bit weird, given what that means and that they we're really being treated as string literals.

Making them `static const` really makes them  const, and remove the thread safety issue, in that if you try to write to them then the compiler will yell at you!

There is also precedent for this in Unit as we already had

```
src/nxt_conf_validation.c:2032:    static const nxt_str_t  wsgi = nxt_string("wsgi");
src/nxt_conf_validation.c:2033:    static const nxt_str_t  asgi = nxt_string("asgi");
src/nxt_conf_validation.c:2424:    static const nxt_str_t  http = nxt_string("http");
src/nxt_conf_validation.c:2425:    static const nxt_str_t  https = nxt_string("https");
src/nxt_controller.c:513:    static const nxt_str_t json = nxt_string(
src/nxt_controller.c:1281:    static const nxt_str_t empty_obj = nxt_string("{}");
src/nxt_h1proto.c:865:    static const nxt_str_t tmp_name_pattern = nxt_string("/req-XXXXXXXX");
src/nxt_h1proto.c:1230:    static const nxt_str_t  connection[3] = {
src/nxt_http_variables.c:521:    static const nxt_str_t  connection[3] = {
src/nxt_php_sapi.c:196:    static const nxt_str_t  chdir = nxt_string("chdir");
src/nxt_router.c:273:static const nxt_str_t http_prefix = nxt_string("HTTP_");
src/nxt_router.c:274:static const nxt_str_t empty_prefix = nxt_string("");
```

Why?

Well it helps improve the code quality by helping to reduce programmer errors, by enabling the compiler to warn you when you are trying to modify something you shouldn't.

As Linus [said](https://yarchive.net/comp/const.html)

> - Anything that *can* take a const pointer should always do so.

With the current fad of C bashing about it being _unsafe_, let's show the doom-sayers that there are things we can do in C (and more so with modern C) that can help eliminate whole classes of problems and this is just _one small_ part of that.

### DIFF
--- a/src/nxt_conf.c
+++ b/src/nxt_conf.c
@@ -130,16 +130,17 @@ static nxt_int_t nxt_conf_copy_array(nxt_mp_t *mp, const nxt_conf_op_t *op,
 static nxt_int_t nxt_conf_copy_object(nxt_mp_t *mp, const nxt_conf_op_t *op,
     nxt_conf_value_t *dst, const nxt_conf_value_t *src);
 
-static size_t nxt_conf_json_string_length(nxt_conf_value_t *value);
-static u_char *nxt_conf_json_print_string(u_char *p, nxt_conf_value_t *value);
-static size_t nxt_conf_json_array_length(nxt_conf_value_t *value,
+static size_t nxt_conf_json_string_length(const nxt_conf_value_t *value);
+static u_char *nxt_conf_json_print_string(u_char *p,
+    const nxt_conf_value_t *value);
+static size_t nxt_conf_json_array_length(const nxt_conf_value_t *value,
     nxt_conf_json_pretty_t *pretty);
-static u_char *nxt_conf_json_print_array(u_char *p, nxt_conf_value_t *value,
+static u_char *nxt_conf_json_print_array(u_char *p,
+    const nxt_conf_value_t *value, nxt_conf_json_pretty_t *pretty);
+static size_t nxt_conf_json_object_length(const nxt_conf_value_t *value,
     nxt_conf_json_pretty_t *pretty);
-static size_t nxt_conf_json_object_length(nxt_conf_value_t *value,
-    nxt_conf_json_pretty_t *pretty);
-static u_char *nxt_conf_json_print_object(u_char *p, nxt_conf_value_t *value,
-    nxt_conf_json_pretty_t *pretty);
+static u_char *nxt_conf_json_print_object(u_char *p,
+    const nxt_conf_value_t *value, nxt_conf_json_pretty_t *pretty);
 
 static size_t nxt_conf_json_escape_length(u_char *p, size_t size);
 static u_char *nxt_conf_json_escape(u_char *dst, u_char *src, size_t size);
@@ -162,21 +163,22 @@ nxt_conf_json_indentation(u_char *p, uint32_t level)
 
 
 void
-nxt_conf_get_string(nxt_conf_value_t *value, nxt_str_t *str)
+nxt_conf_get_string(const nxt_conf_value_t *value, nxt_str_t *str)
 {
     if (value->type == NXT_CONF_VALUE_SHORT_STRING) {
         str->length = value->u.str.length;
-        str->start = value->u.str.start;
+        str->start = (u_char *) value->u.str.start;
 
     } else {
         str->length = value->u.string.length;
-        str->start = value->u.string.start;
+        str->start = (u_char *) value->u.string.start;
     }
 }
 
 
 nxt_str_t *
-nxt_conf_get_string_dup(nxt_conf_value_t *value, nxt_mp_t *mp, nxt_str_t *str)
+nxt_conf_get_string_dup(const nxt_conf_value_t *value, nxt_mp_t *mp,
+    nxt_str_t *str)
 {
     nxt_str_t  s;
 
@@ -2233,7 +2235,8 @@ nxt_conf_json_parse_error(nxt_conf_json_error_t *error, u_char *pos,
 
 
 size_t
-nxt_conf_json_length(nxt_conf_value_t *value, nxt_conf_json_pretty_t *pretty)
+nxt_conf_json_length(const nxt_conf_value_t *value,
+    nxt_conf_json_pretty_t *pretty)
 {
     switch (value->type) {
 
@@ -2265,7 +2268,7 @@ nxt_conf_json_length(nxt_conf_value_t *value, nxt_conf_json_pretty_t *pretty)
 
 
 u_char *
-nxt_conf_json_print(u_char *p, nxt_conf_value_t *value,
+nxt_conf_json_print(u_char *p, const nxt_conf_value_t *value,
     nxt_conf_json_pretty_t *pretty)
 {
     switch (value->type) {
@@ -2299,7 +2302,7 @@ nxt_conf_json_print(u_char *p, nxt_conf_value_t *value,
 
 
 static size_t
-nxt_conf_json_string_length(nxt_conf_value_t *value)
+nxt_conf_json_string_length(const nxt_conf_value_t *value)
 {
     nxt_str_t  str;
 
@@ -2310,7 +2313,7 @@ nxt_conf_json_string_length(nxt_conf_value_t *value)
 
 
 static u_char *
-nxt_conf_json_print_string(u_char *p, nxt_conf_value_t *value)
+nxt_conf_json_print_string(u_char *p, const nxt_conf_value_t *value)
 {
     nxt_str_t  str;
 
@@ -2327,7 +2330,7 @@ nxt_conf_json_print_string(u_char *p, nxt_conf_value_t *value)
 
 
 static size_t
-nxt_conf_json_array_length(nxt_conf_value_t *value,
+nxt_conf_json_array_length(const nxt_conf_value_t *value,
     nxt_conf_json_pretty_t *pretty)
 {
     size_t            len;
@@ -2369,7 +2372,7 @@ nxt_conf_json_array_length(nxt_conf_value_t *value,
 
 
 static u_char *
-nxt_conf_json_print_array(u_char *p, nxt_conf_value_t *value,
+nxt_conf_json_print_array(u_char *p, const nxt_conf_value_t *value,
     nxt_conf_json_pretty_t *pretty)
 {
     nxt_uint_t        n;
@@ -2421,7 +2424,7 @@ nxt_conf_json_print_array(u_char *p, nxt_conf_value_t *value,
 
 
 static size_t
-nxt_conf_json_object_length(nxt_conf_value_t *value,
+nxt_conf_json_object_length(const nxt_conf_value_t *value,
     nxt_conf_json_pretty_t *pretty)
 {
     size_t                    len;
@@ -2465,7 +2468,7 @@ nxt_conf_json_object_length(nxt_conf_value_t *value,
 
 
 static u_char *
-nxt_conf_json_print_object(u_char *p, nxt_conf_value_t *value,
+nxt_conf_json_print_object(u_char *p, const nxt_conf_value_t *value,
     nxt_conf_json_pretty_t *pretty)
 {
     nxt_uint_t                n;
--- a/src/nxt_conf.c
+++ b/src/nxt_conf.c
@@ -123,12 +123,12 @@ static u_char *nxt_conf_json_parse_number(nxt_mp_t *mp, nxt_conf_value_t *value,
 static void nxt_conf_json_parse_error(nxt_conf_json_error_t *error, u_char *pos,
     const char *detail);
 
-static nxt_int_t nxt_conf_copy_value(nxt_mp_t *mp, nxt_conf_op_t *op,
-    nxt_conf_value_t *dst, nxt_conf_value_t *src);
-static nxt_int_t nxt_conf_copy_array(nxt_mp_t *mp, nxt_conf_op_t *op,
-    nxt_conf_value_t *dst, nxt_conf_value_t *src);
-static nxt_int_t nxt_conf_copy_object(nxt_mp_t *mp, nxt_conf_op_t *op,
-    nxt_conf_value_t *dst, nxt_conf_value_t *src);
+static nxt_int_t nxt_conf_copy_value(nxt_mp_t *mp, const nxt_conf_op_t *op,
+    nxt_conf_value_t *dst, const nxt_conf_value_t *src);
+static nxt_int_t nxt_conf_copy_array(nxt_mp_t *mp, const nxt_conf_op_t *op,
+    nxt_conf_value_t *dst, const nxt_conf_value_t *src);
+static nxt_int_t nxt_conf_copy_object(nxt_mp_t *mp, const nxt_conf_op_t *op,
+    nxt_conf_value_t *dst, const nxt_conf_value_t *src);
 
 static size_t nxt_conf_json_string_length(nxt_conf_value_t *value);
 static u_char *nxt_conf_json_print_string(u_char *p, nxt_conf_value_t *value);
@@ -186,7 +186,7 @@ nxt_conf_get_string_dup(nxt_conf_value_t *value, nxt_mp_t *mp, nxt_str_t *str)
 
 
 void
-nxt_conf_set_string(nxt_conf_value_t *value, nxt_str_t *str)
+nxt_conf_set_string(nxt_conf_value_t *value, const nxt_str_t *str)
 {
     if (str->length > NXT_CONF_MAX_SHORT_STRING) {
         value->type = NXT_CONF_VALUE_STRING;
@@ -245,7 +245,7 @@ nxt_conf_get_boolean(nxt_conf_value_t *value)
 
 
 nxt_uint_t
-nxt_conf_object_members_count(nxt_conf_value_t *value)
+nxt_conf_object_members_count(const nxt_conf_value_t *value)
 {
     return value->u.object->count;
 }
@@ -276,7 +276,7 @@ nxt_conf_create_object(nxt_mp_t *mp, nxt_uint_t count)
 
 
 void
-nxt_conf_set_member(nxt_conf_value_t *object, nxt_str_t *name,
+nxt_conf_set_member(nxt_conf_value_t *object, const nxt_str_t *name,
     const nxt_conf_value_t *value, uint32_t index)
 {
     nxt_conf_object_member_t  *member;
@@ -290,8 +290,8 @@ nxt_conf_set_member(nxt_conf_value_t *object, nxt_str_t *name,
 
 
 nxt_int_t
-nxt_conf_set_member_dup(nxt_conf_value_t *object, nxt_mp_t *mp, nxt_str_t *name,
-    nxt_conf_value_t *value, uint32_t index)
+nxt_conf_set_member_dup(nxt_conf_value_t *object, nxt_mp_t *mp,
+    const nxt_str_t *name, const nxt_conf_value_t *value, uint32_t index)
 {
     nxt_conf_object_member_t  *member;
 
@@ -304,8 +304,8 @@ nxt_conf_set_member_dup(nxt_conf_value_t *object, nxt_mp_t *mp, nxt_str_t *name,
 
 
 void
-nxt_conf_set_member_string(nxt_conf_value_t *object, nxt_str_t *name,
-    nxt_str_t *value, uint32_t index)
+nxt_conf_set_member_string(nxt_conf_value_t *object, const nxt_str_t *name,
+    const nxt_str_t *value, uint32_t index)
 {
     nxt_conf_object_member_t  *member;
 
@@ -319,7 +319,7 @@ nxt_conf_set_member_string(nxt_conf_value_t *object, nxt_str_t *name,
 
 nxt_int_t
 nxt_conf_set_member_string_dup(nxt_conf_value_t *object, nxt_mp_t *mp,
-    nxt_str_t *name, nxt_str_t *value, uint32_t index)
+    const nxt_str_t *name, const nxt_str_t *value, uint32_t index)
 {
     nxt_conf_object_member_t  *member;
 
@@ -332,7 +332,7 @@ nxt_conf_set_member_string_dup(nxt_conf_value_t *object, nxt_mp_t *mp,
 
 
 void
-nxt_conf_set_member_integer(nxt_conf_value_t *object, nxt_str_t *name,
+nxt_conf_set_member_integer(nxt_conf_value_t *object, const nxt_str_t *name,
     int64_t value, uint32_t index)
 {
     u_char                    *p, *end;
@@ -353,7 +353,7 @@ nxt_conf_set_member_integer(nxt_conf_value_t *object, nxt_str_t *name,
 
 
 void
-nxt_conf_set_member_null(nxt_conf_value_t *object, nxt_str_t *name,
+nxt_conf_set_member_null(nxt_conf_value_t *object, const nxt_str_t *name,
     uint32_t index)
 {
     nxt_conf_object_member_t  *member;
@@ -400,7 +400,7 @@ nxt_conf_set_element(nxt_conf_value_t *array, nxt_uint_t index,
 
 nxt_int_t
 nxt_conf_set_element_string_dup(nxt_conf_value_t *array, nxt_mp_t *mp,
-    nxt_uint_t index, nxt_str_t *value)
+    nxt_uint_t index, const nxt_str_t *value)
 {
     nxt_conf_value_t  *element;
 
@@ -411,21 +411,21 @@ nxt_conf_set_element_string_dup(nxt_conf_value_t *array, nxt_mp_t *mp,
 
 
 nxt_uint_t
-nxt_conf_array_elements_count(nxt_conf_value_t *value)
+nxt_conf_array_elements_count(const nxt_conf_value_t *value)
 {
     return value->u.array->count;
 }
 
 
 nxt_uint_t
-nxt_conf_array_elements_count_or_1(nxt_conf_value_t *value)
+nxt_conf_array_elements_count_or_1(const nxt_conf_value_t *value)
 {
     return (value->type == NXT_CONF_VALUE_ARRAY) ? value->u.array->count : 1;
 }
 
 
 nxt_uint_t
-nxt_conf_type(nxt_conf_value_t *value)
+nxt_conf_type(const nxt_conf_value_t *value)
 {
     switch (value->type) {
 
@@ -459,7 +459,7 @@ nxt_conf_type(nxt_conf_value_t *value)
 
 
 nxt_conf_value_t *
-nxt_conf_get_path(nxt_conf_value_t *value, nxt_str_t *path)
+nxt_conf_get_path(nxt_conf_value_t *value, const nxt_str_t *path)
 {
     nxt_str_t              token;
     nxt_int_t              ret, index;
@@ -550,7 +550,7 @@ nxt_conf_path_next_token(nxt_conf_path_parse_t *parse, nxt_str_t *token)
 
 
 nxt_conf_value_t *
-nxt_conf_get_object_member(nxt_conf_value_t *value, nxt_str_t *name,
+nxt_conf_get_object_member(const nxt_conf_value_t *value, const nxt_str_t *name,
     uint32_t *index)
 {
     nxt_str_t                 str;
@@ -584,8 +584,8 @@ nxt_conf_get_object_member(nxt_conf_value_t *value, nxt_str_t *name,
 
 
 nxt_int_t
-nxt_conf_map_object(nxt_mp_t *mp, nxt_conf_value_t *value, nxt_conf_map_t *map,
-    nxt_uint_t n, void *data)
+nxt_conf_map_object(nxt_mp_t *mp, const nxt_conf_value_t *value,
+    const nxt_conf_map_t *map, nxt_uint_t n, void *data)
 {
     double            num;
     nxt_str_t         str, *s;
@@ -736,7 +736,7 @@ nxt_conf_map_object(nxt_mp_t *mp, nxt_conf_value_t *value, nxt_conf_map_t *map,
 
 
 nxt_conf_value_t *
-nxt_conf_next_object_member(nxt_conf_value_t *value, nxt_str_t *name,
+nxt_conf_next_object_member(const nxt_conf_value_t *value, nxt_str_t *name,
     uint32_t *next)
 {
     uint32_t                  n;
@@ -764,7 +764,7 @@ nxt_conf_next_object_member(nxt_conf_value_t *value, nxt_str_t *name,
 
 
 nxt_conf_value_t *
-nxt_conf_get_array_element(nxt_conf_value_t *value, uint32_t index)
+nxt_conf_get_array_element(const nxt_conf_value_t *value, uint32_t index)
 {
     nxt_conf_array_t  *array;
 
@@ -802,7 +802,7 @@ nxt_conf_get_array_element_or_itself(nxt_conf_value_t *value, uint32_t index)
 
 
 void
-nxt_conf_array_qsort(nxt_conf_value_t *value,
+nxt_conf_array_qsort(const nxt_conf_value_t *value,
     int (*compare)(const void *, const void *))
 {
     nxt_conf_array_t  *array;
@@ -818,8 +818,9 @@ nxt_conf_array_qsort(nxt_conf_value_t *value,
 
 
 nxt_conf_op_ret_t
-nxt_conf_op_compile(nxt_mp_t *mp, nxt_conf_op_t **ops, nxt_conf_value_t *root,
-    nxt_str_t *path, nxt_conf_value_t *value, nxt_bool_t add)
+nxt_conf_op_compile(nxt_mp_t *mp, nxt_conf_op_t **ops,
+    const nxt_conf_value_t *root, const nxt_str_t *path,
+    nxt_conf_value_t *value, nxt_bool_t add)
 {
     nxt_str_t                 token;
     nxt_int_t                 ret, index;
@@ -956,7 +957,7 @@ nxt_conf_op_compile(nxt_mp_t *mp, nxt_conf_op_t **ops, nxt_conf_value_t *root,
 
 
 nxt_conf_value_t *
-nxt_conf_clone(nxt_mp_t *mp, nxt_conf_op_t *op, nxt_conf_value_t *value)
+nxt_conf_clone(nxt_mp_t *mp, nxt_conf_op_t *op, const nxt_conf_value_t *value)
 {
     nxt_int_t         rc;
     nxt_conf_value_t  *copy;
@@ -977,8 +978,8 @@ nxt_conf_clone(nxt_mp_t *mp, nxt_conf_op_t *op, nxt_conf_value_t *value)
 
 
 static nxt_int_t
-nxt_conf_copy_value(nxt_mp_t *mp, nxt_conf_op_t *op, nxt_conf_value_t *dst,
-    nxt_conf_value_t *src)
+nxt_conf_copy_value(nxt_mp_t *mp, const nxt_conf_op_t *op,
+    nxt_conf_value_t *dst, const nxt_conf_value_t *src)
 {
     if (op != NULL
         && src->type != NXT_CONF_VALUE_ARRAY
@@ -1020,8 +1021,8 @@ nxt_conf_copy_value(nxt_mp_t *mp, nxt_conf_op_t *op, nxt_conf_value_t *dst,
 
 
 static nxt_int_t
-nxt_conf_copy_array(nxt_mp_t *mp, nxt_conf_op_t *op, nxt_conf_value_t *dst,
-    nxt_conf_value_t *src)
+nxt_conf_copy_array(nxt_mp_t *mp, const nxt_conf_op_t *op,
+    nxt_conf_value_t *dst, const nxt_conf_value_t *src)
 {
     size_t            size;
     nxt_int_t         rc;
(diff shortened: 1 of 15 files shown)


src/nxt_conf.c
```diff
@@ -459,7 +463,7 @@ nxt_conf_type(nxt_conf_value_t *value)
 
 
 nxt_conf_value_t *
-nxt_conf_get_path(nxt_conf_value_t *value, nxt_str_t *path)
+nxt_conf_get_path(const nxt_conf_value_t *value, const nxt_str_t *path)
```
hongzhidao, line 466: If we touch it, I feel it's like this:
```
const nxt_conf_value_t *
nxt_conf_get_path(const nxt_conf_value_t *value, const nxt_str_t *path)
```
Since both the value and the return value are read-only.

src/nxt_conf.c
```diff
@@ -162,21 +163,24 @@ nxt_conf_json_indentation(u_char *p, uint32_t level)
 
 
 void
-nxt_conf_get_string(nxt_conf_value_t *value, nxt_str_t *str)
+nxt_conf_get_string(const nxt_conf_value_t *value, nxt_str_t *str)
 {
+    nxt_conf_value_t *v = (nxt_conf_value_t *) value;
```
hongzhidao, line 168: The `value` is not updated inside the function, why do we need to use the temporary `v`?

src/nxt_conf.c
```diff
@@ -783,12 +787,13 @@ nxt_conf_get_array_element(nxt_conf_value_t *value, uint32_t index)
...
 nxt_conf_value_t *
-nxt_conf_get_array_element_or_itself(nxt_conf_value_t *value, uint32_t index)
+nxt_conf_get_array_element_or_itself(const nxt_conf_value_t *value,
+    uint32_t index)
 {
     nxt_conf_array_t  *array;
 
     if (value->type != NXT_CONF_VALUE_ARRAY) {
-        return (index == 0) ? value : NULL;
+        return (index == 0) ? (nxt_conf_value_t *) value : NULL;
```
hongzhidao, line 796: The same as above, should the return type be with `const`?

src/nxt_conf.c
ac000: Ah, let me check...

ac000: OK, so that's going to be a much larger intrusive change. You have to start making function variables `const` then you need to start making various structure members `const`, or you introduce a tonne of casting...

Either way this would definitely be a separate commit.

src/nxt_conf.c
ac000: I'll check, but again this is likely to result in a much larger intrusive change.

src/nxt_conf.c
hongzhidao: Yep, it will introduce much change. Or stay them the same.

hongzhidao: Btw, others like just adding `const` look good. They can be released first.

src/nxt_conf.c
ac000: Otherwise we get

```
... error: initialization discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
```

Which I _think_ is [this](https://port70.net/~nsz/c/c11/n1570.html#6.7.3p12)

```
pi = &cs.mem;           //   violates type constraints for =
```

hongzhidao: Take a look at this first.
```
static nxt_int_t
nxt_php_set_target(nxt_task_t *task, nxt_php_target_t *target,
    nxt_conf_value_t *conf)
{
    ...
    nxt_conf_value_t  *value;

    ...

    value = nxt_conf_get_object_member(conf, &root_str, NULL);

    nxt_conf_get_string(value, &str);
```
I found the function isn't changed, but if we changed the `nxt_conf_get_string()`, did you miss touching where it's called?

ac000: I don't think the fact `value` starts off non-const before being passed into a function that takes it as a const is a problem. Indeed the compiler is perfectly happy with it and it's a perfectly reasonable thing to do, you may want to make some changes to it locally or in some other function, but _not_ some other function... E.g

```c
nxt_conf_value_t  *value;

value = ...

change_value_in_some_way(value);    // value passed in as non-const

...

dont_change_value(value);    // value passed in as const
```

True, it may be possible that in the case you showed value could be made const, but such changes were out of scope for what I was wanting to achieve  with these patches.

src/nxt_conf.c
ac000: I can certainly drop _this_ change for now, it doesn't prevent my initial goal with these patches...

src/nxt_conf.c
hongzhidao: > If yes, there is the same issue as above.
Sorry if I misunderstand you. It's not an issue. I mean the new way introduced a bit of confusion.
Let's return the function.
```
void
nxt_conf_get_string(nxt_conf_value_t *value, nxt_str_t *str)
{
    if (value->type == NXT_CONF_VALUE_SHORT_STRING) {
        str->length = value->u.str.length;
        str->start = value->u.str.start;

    } else {
        str->length = value->u.string.length;
        str->start = value->u.string.start;
    }
}
```
It's to get the string from the conf value.
The key is if `str` is changeable. I feel it's read-only, if yes, it should have its own allocation, for example:
```
void
nxt_conf_get_string(nxt_conf_value_t *value, nxt_str_t *str)
{
    ...
    if (value->type == NXT_CONF_VALUE_SHORT_STRING) {
        length = value->u.str.length;
        start = value->u.str.start;

    } else {
        length = value->u.string.length;
        start = value->u.string.start;
    }

    str->start = malloc(length); /* mp should be used */
    ...
}
```

So if we make `value` const, it seems `str` is also const. For example:
```
void
nxt_conf_get_string(const nxt_conf_value_t *value, const nxt_str_t *str)
```

src/nxt_conf.c
ac000: I can also drop this...

src/nxt_conf.c
ac000: Hmm, wait, hang on.

I'm maybe not understanding what you're getting at at, but it won't make any difference if value is const or not before calling `nxt_conf_get_string()`, you'd still get the compiler warning...

hongzhidao: 1. The following is a bit confusing. Reviewers need to understand it's because of the compiler warning.
```
nxt_conf_value_t *v = (nxt_conf_value_t *) value;
```
2. It seems it's unnecessary to add the above to avoid warning, and it's better not to add something like it.
3. If `value` is const, `str` is also const, since they are read-only and str gets part from `value`.

src/nxt_conf.c
ac000: Dropped.

src/nxt_conf.c
ac000: Dropped.

src/nxt_conf.c
ac000: If `str` is `const`, then we can't set any of its members. E.g quick hack to demonstrate

```c
void                                                                            
nxt_conf_get_string(const nxt_conf_value_t *value, const nxt_str_t *str)        
{                                                                                                                                               
    if (value->type == NXT_CONF_VALUE_SHORT_STRING) {                           
        str->length = value->u.str.length;                                      
        str->start = value->u.str.start;                                        
                                                                                
    } else {                                                                    
        str->length = value->u.string.length;                                   
        str->start = value->u.string.start;                                     
    }                                                                           
} 
```

```
  CC     build/src/nxt_conf.o
src/nxt_conf.c: In function ‘nxt_conf_get_string’:
src/nxt_conf.c:169:21: error: assignment of member ‘length’ in read-only object
  169 |         str->length = value->u.str.length;
      |                     ^
src/nxt_conf.c:170:20: error: assignment of member ‘start’ in read-only object
  170 |         str->start = value->u.str.start;
      |                    ^
src/nxt_conf.c:173:21: error: assignment of member ‘length’ in read-only object
  173 |         str->length = value->u.string.length;
      |                     ^
src/nxt_conf.c:174:20: error: assignment of member ‘start’ in read-only object
  174 |         str->start = value->u.string.start;
      |                    ^
```

Of course you'd hit the exact same thing with the `str->start = malloc(length);`

This is the point of making things const after all...

The only way I can immediately think of to make value const and not have the cast is by using memcpy(3)...

hongzhidao: > If str is const, then we can't set any of its members.

I didn't test it, it's to show the idea. Anyway, if we touch the function, it's important to check if `str` is read-only.

hongzhidao: Will check it tomorrow, not sure if my idea is correct.
The same for the two changes you dropped, they might make sense.

hongzhidao: Btw, you could try with this.
```
void
nxt_conf_get_string(const nxt_conf_value_t *value, nxt_str_t *str)
{
    if (value->type == NXT_CONF_VALUE_SHORT_STRING) {
        str->length = value->u.str.length;
        str->start = (u_char *) value->u.str.start;

    } else {
        str->length = value->u.string.length;
        str->start = value->u.string.start;
    }
}
```

ac000: > I didn't test it, it's to show the idea. Anyway, if we touch the function, it's important to check if str is read-only.

It _can't_ be read-only as we _need_ to modify it...

Indeed, your suggestion works.

For some bizarre reason I thought it was failing on the `start->length = `, which didn't make sense to me, no wonder!

So you're happy with `(u_char *)` casts?

As the dropped changes, I'm happy to do them in a separate commit anyway and am not even 100% conviced by the `nxt_conf_get_path()` one...

ac000:
Added the `(u_char *)` casts in `nxt_conf_get_string()`

```
$ git range-diff 996e7c78...0a8ef824
1:  996e7c78 ! 1:  0a8ef824 configuration: Constify more pointers
    @@ Commit message
         While it takes a value parameter that is never modified, simply making
         it const results in
     
    -      warning: initialization discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
    +      CC     build/src/nxt_conf.o
    +    src/nxt_conf.c: In function ‘nxt_conf_get_string’:
    +    src/nxt_conf.c:170:20: error: assignment discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
    +      170 |         str->start = value->u.str.start;
    +          |                    ^
     
         due to the assignment operator. Making value const will allow for
         numerous other constification and seeing as we are not modifying it,
         seems worthwhile.
     
    -    So to get around the warning we simply take a non-const copy of value.
    +    We can get around the warning by casting ->u.{str,string}.start
     
    @@ src/nxt_conf.c: nxt_conf_json_indentation(u_char *p, uint32_t level)
     -nxt_conf_get_string(nxt_conf_value_t *value, nxt_str_t *str)
     +nxt_conf_get_string(const nxt_conf_value_t *value, nxt_str_t *str)
      {
    -+    nxt_conf_value_t *v = (nxt_conf_value_t *) value;
    -+
          if (value->type == NXT_CONF_VALUE_SHORT_STRING) {
    --        str->length = value->u.str.length;
    +         str->length = value->u.str.length;
     -        str->start = value->u.str.start;
    -+        str->length = v->u.str.length;
    -+        str->start = v->u.str.start;
    ++        str->start = (u_char *) value->u.str.start;
      
          } else {
    --        str->length = value->u.string.length;
    +         str->length = value->u.string.length;
     -        str->start = value->u.string.start;
    -+        str->length = v->u.string.length;
    -+        str->start = v->u.string.start;
    ++        str->start = (u_char *) value->u.string.start;
          }
      }
      
```

ac000:
Fix some wrapping/indentation

```
$ git range-diff 0a8ef824...b4d2153a
1:  0a8ef824 ! 1:  b4d2153a configuration: Constify more pointers
    @@ src/nxt_conf.h: nxt_conf_value_t *nxt_conf_json_parse(nxt_mp_t *mp, u_char *star
      
     -NXT_EXPORT void nxt_conf_get_string(nxt_conf_value_t *value, nxt_str_t *str);
     -NXT_EXPORT nxt_str_t *nxt_conf_get_string_dup(nxt_conf_value_t *value,
    -+NXT_EXPORT void nxt_conf_get_string(const nxt_conf_value_t *value, nxt_str_t *str);
    ++NXT_EXPORT void nxt_conf_get_string(const nxt_conf_value_t *value,
    ++    nxt_str_t *str);
     +NXT_EXPORT nxt_str_t *nxt_conf_get_string_dup(const nxt_conf_value_t *value,
          nxt_mp_t *mp, nxt_str_t *str);
      NXT_EXPORT void nxt_conf_set_string(nxt_conf_value_t *value,
```

src/nxt_conf.c
ac000: > ```
> nxt_conf_value_t *
> nxt_conf_get_path(const nxt_conf_value_t *value, nxt_str_t *path)
> ```
>
> I feel the declaration is correct. And can this work without adding v?
> ```
> @@ -510,7 +510,7 @@ nxt_conf_get_path(nxt_conf_value_t *value, nxt_str_t *path)
>
>    } while (parse.last == 0);
>
> -    return value;
> +    return (nxt_conf_value_t *) value;
>  }
> ```

Yes, that is actually what I had originally...

```diff
@@ -459,7 +459,7 @@ nxt_conf_type(nxt_conf_value_t *value)
 
 
 nxt_conf_value_t *
-nxt_conf_get_path(nxt_conf_value_t *value, nxt_str_t *path)
+nxt_conf_get_path(const nxt_conf_value_t *value, const nxt_str_t *path)
 {
     nxt_str_t              token;
     nxt_int_t              ret, index;
@@ -510,7 +510,7 @@ nxt_conf_get_path(nxt_conf_value_t *value, nxt_str_t *path)
 
     } while (parse.last == 0);
 
-    return value;
+    return (nxt_conf_value_t *) value;
 }
 
 
```

The bit I'm not 100% sure about is, that yes, while it clearly works, taking a const then returning _it_ as a non-const perhaps has a slight _smell_ to it... need to mull it over...

callahad:
> He ghosted me after he got some value for free.

Hi @alejandro-colomar, I have no idea what you're talking about. I received no further communication from you and assumed you were occupied with other things. :-/ Sounds like something has been dropped, but I don't know where. Regardless, I'm sincerely sorry that you feel like I've ghosted you. Not my intent; let's cure that.

> mainly due to incompetence

I must ask you to refrain from disparaging folks here. #795 hasn't been touched in well over a year, significantly before my time. We need to thoroughly re-triage, but righting this ship takes time. We'll get there.

callahad:
Update: Corporate Outlook was flagging Alex's replies to me as Spam. 🤦‍♂️

Let's keep this review focused on this patch.

alejandro-colomar:
> Update: Corporate Outlook was flagging Alex's replies to me as Spam. 🤦‍♂️ I've followed up out-of-band.
> 
> Let's keep this review focused on this patch.

I'm glad to see that it wasn't your fault.  I'm sorry about the assumptions I made.  If you want, I can drop my comment, or just leave it for historical reasons.

alejandro-colomar:
Hi Andrew!

The thing with `const`ifying `nxt_str_t` is that you're constifying the pointers within the structure, not the pointees.  There's no way in C to constify the pointee.  You'd need to use a distinct structure type.

Have a lovely night!
Alex

ac000:
> Hi Andrew!

Hi Alex,

> This is placebo.

Heh, _if_ the below was really happening then it's worse than that!
 
> ```
> Nacked-by: Alejandro Colomar <alx@kernel.org>
> ```
> 
> ```
> alx@debian:~/tmp/c$ cat const.c 
> #include <stdio.h>
> #include <string.h>
> 
> struct s {
> 	int len;
> 	char *s;
> };
> 
> static const struct s s = {3, "foo"};
> 
> int
> main(void)
> {
> 	memcpy(s.s, "bar", 3);
> 	putchar(s.s[0]);
> }
> alx@debian:~/tmp/c$ cc -Wall -Wextra const.c 
> alx@debian:~/tmp/c$ clang -Weverything const.c 
> const.c:6:8: warning: padding struct 'struct s' with 4 bytes to align 's' [-Wpadded]
>         char *s;
>               ^
> 1 warning generated.
> alx@debian:~/tmp/c$ ./a.out 
> Segmentation fault
> ```

Indeed, I _am_ actually aware that modifying something defined as const
behind its back is _undefined behaviour_...

I've re-checked where I've added `const`s and I don't see anywhere we
try to modify them indirectly.

However your demo program segfaults for _other_ reasons. I.e it still
segfaults if you remove the `static const`

It segfaults on the `memcpy(3)` due to I _think_ trying to overwrite the
string literal "foo".

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000000000401055 in main () at const.c:13
13              memcpy(s.s, "bar", 3);
```

valgrind shows

```
==129333== Process terminating with default action of signal 11 (SIGSEGV): dumping core
==129333==  Bad permissions for mapped region at address 0x402010
==129333==    at 0x401050: main (const.c:13)
==129333== 
```

However, I think I know what you were trying to demonstrate...

ac000:
> The thing with constifying nxt_str_t is that you're constifying the
> pointer within the structure, not the pointee. There's no way in C to
> constify the pointee. You'd need to use a distinct structure type.

Yes, I'm aware.

> So, your patch protects against pointing to a different string
> (s->start = "foo";) but not against overwriting the string
> (memcpy(s->start, "foo", 3);).

Yes, but also no, as in it's not so much about _preventing_ it from
being modified, you can always access it through a non-const alias.

As we know, C simply has _no_ concept of `const` meaning that the
underlying memory is immutable.

What this is _really_ about is type checking. It's about informing the
compiler of your intentions.

The compiler can then warn you if you deviate from that (to a point,
obviously if you're determined you can get around it, but then you're
definitely shooting yourself in the foot).

So as Linus says, anything that can be `const` should be, to give the
compiler more opportunity to help you. Of course it's best to think about
this upfront because trying to do this retroactively is a minefield.

So that's what this patch set is for, not really about _preventing_
things from being modified, but about enabling more compiler warnings
if you try to do something unintentional.

alejandro-colomar:
> > Hi Andrew!
> 
> Hi Alex,
> 
> > This is placebo.
> 
> Heh, _if_ the below was really happening then it's worse than that!

...

> > ```
> > alx@debian:~/tmp/c$ cat const.c 
> > #include <stdio.h>
> > #include <string.h>
> > 
> > struct s {
> > 	int len;
> > 	char *s;
> > };
> > 
> > static const struct s s = {3, "foo"};
> > 
> > int
> > main(void)
> > {
> > 	memcpy(s.s, "bar", 3);
> > 	putchar(s.s[0]);
> > }
> > alx@debian:~/tmp/c$ cc -Wall -Wextra const.c 
> > alx@debian:~/tmp/c$ clang -Weverything const.c 
> > const.c:6:8: warning: padding struct 'struct s' with 4 bytes to align 's' [-Wpadded]
> >         char *s;
> >               ^
> > 1 warning generated.
> > alx@debian:~/tmp/c$ ./a.out 
> > Segmentation fault
> > ```
> 
> Indeed, I _am_ actually aware that modifying something defined as const behind its back is _undefined behaviour_...
> 
> I've re-checked where I've added `const`s and I don't see anywhere we try to modify them indirectly.

No, you don't modify the string literals.

> However your demo program segfaults for _other_ reasons. I.e it still segfaults if you remove the `static const`
> 
> It segfaults on the `memcpy(3)` due to I _think_ trying to overwrite the string literal "foo".

Yep.

> 
> ```
> Program terminated with signal SIGSEGV, Segmentation fault.
> #0  0x0000000000401055 in main () at const.c:13
> 13              memcpy(s.s, "bar", 3);
> ```
> 
> valgrind shows
> 
> ```
> ==129333== Process terminating with default action of signal 11 (SIGSEGV): dumping core
> ==129333==  Bad permissions for mapped region at address 0x402010
> ==129333==    at 0x401050: main (const.c:13)
> ==129333== 
> ```
> 
> However, I think I know what you were trying to demonstrate...

Yep, I assume you know it.  :)

Just for others that read this: I was trying to demonstrate that this use of `const` does not protect against modifying the strings.

```
if you try to write to them then the compiler will yell at you!
```

So, it won't really yell at you.  Yeah, if you modify the pointer to point to something else, or the length (admittedly, one will usually modify the length when modifying the string), it'll yell at you.  And yeah, that's already an improvement.  But only a mild improvement, not really protecting against a simple modification of the string, which is why I called it a placebo.

ac000:
> Just to remove doubt: I was trying to demonstrate that this use of
> `const` does not protect against modifying the strings (which is the
> main reason one would want to use it, after all).

I think I covered this is in my other reply. But C simply no such thing.
`const` is _really_ about type checking _not_ preventing things being modified.

I didn't _really_ understand what `const` was actually for until I read [this](https://yarchive.net/comp/const.html).

alejandro-colomar:
> > The thing with constifying nxt_str_t is that you're constifying the
> > pointer within the structure, not the pointee. There's no way in C to
> > constify the pointee. You'd need to use a distinct structure type.
> 
> Yes, I'm aware.
> 
> > So, your patch protects against pointing to a different string
> > (s->start = "foo";) but not against overwriting the string
> > (memcpy(s->start, "foo", 3);).
> 
> Yes, but also no, as in it's not so much about _preventing_ it from being modified, you can always access it through a non-const alias.
> 
> As we know, C simply has _no_ concept of `const` meaning that the underlying memory is immutable.
> 
> What this is _really_ about is type checking. It's about informing the compiler of your intentions.
> 
> The compiler can then warn you if you deviate from that (to a point, obviously if you're determined you can get around it, but then you're definitely shooting yourself in the foot).

My point is that you don't need to use casts or other workarounds to tell the compiler that you're determined to shoot yourself in the foot in this case.  In my example program, I enabled all warnings, used const, and still shoot myself in the foot.  `nxt_str_t` is necessarily unsafe, as it is now.  const can't make it safe.

> 
> So as Linus says, anything that can be `const` should be, to give the compiler more opportunity to help you. Of course it's best to think about this upfront because trying to do this retroactively is a minefield.
> 
> So that's what this patch set is for, not really about _preventing_ things from being modified, but about enabling more compiler warnings if you try to do something unintentional.

Hmmm, I agree that adding const where possible is good, but I think this case needs a little bit more thought, because one may think that by adding const the compiler will warn when doing something stupid, and then it will be a bad surprise when one realizes that the compiler won't really warn.

So, while I agree to adding this const, I think this is just protecting half of the problem.

ac000:
> So, while I agree to adding this const, I think this is just
> protecting half of the problem.

Right, my impetus was to fix all the weird `static nxt_str_t`'s, the
initial patches allowed for that, the last patch was just some other low
hanging fruit in the vicinity.

We of course could go much further, would have been much better if
this was though about upfront, but here we are. I tried doing some other
const stuff and very quickly you do down a rabbit hole af adding consts
(and casts) throughout the code base...

alejandro-colomar:
Here's a smaller example showing what's going on:

```c
alx@debian:~/tmp/c$ cat const.c 
#include <stdio.h>
#include <string.h>

int
main(void)
{
	char *const s = "foo";

	memcpy(s, "bar", 3);
	// s = "baz";  /* error: assignment of read-only variable ‘s’ */
	putchar(s[0]);
}
alx@debian:~/tmp/c$ cc -Wall -Wextra const.c 
alx@debian:~/tmp/c$ ./a.out 
Segmentation fault
```

This shows how having const in the pointer does not produce any warning (yeah, in the `=` line it does, but).  Mainly because const is discarded when taking an rvalue (so when passing to any function), since we actually take a copy of the (pointer) value, not the (pointer) variable itself.  So, there's really little type safety in this example program.

alejandro-colomar:
ISO C relevant point:
<http://port70.net/~nsz/c/c11/n1570.html#6.3.2.1p2>

ac000:
I think that's still wrong... "foo" is still going to be in read-only memory... You'd still get the segfault without the const. I think you'd need to make it point to some malloc'd memory for it to _"work"_...

alejandro-colomar:
> I think that's still wrong... "foo" is still going to be in read-only memory... You'd still get the segfault without the const. I think you'd need to make it point to some malloc'd memory for it to _"work"_...

Oh yeah, the segfault is on purpose.  I want to show that I'm "accidentally" shooting myself in the foot, and the compiler is not yelling at me, because this use of `const` really provides no type safety.

alejandro-colomar:
If you prefer it without a segfault, here's some example modifying a malloc'd string:

```c
alx@debian:~/tmp/c$ cat const.c 
#include <stdio.h>
#include <string.h>

int
main(void)
{
	char *const  s = strdup("foo");

	printf("%s\n", strcpy(s, "bar"));
}
alx@debian:~/tmp/c$ cc -Wall -Wextra const.c 
alx@debian:~/tmp/c$ ./a.out 
bar
```

alejandro-colomar:
Relevant links:

-  <https://stackoverflow.com/a/1143321/6872717>
-  <https://stackoverflow.com/questions/21476869/constant-pointer-vs-pointer-to-constant>

ac000:
> If you prefer it without a segfault, here's some example modifying a malloc'd string:

Well, it's just that it wasn't because of any `const` thing that it was segfaulting...

> ```c
> alx@debian:~/tmp/c$ cat const.c 
> #include <stdio.h>
> #include <string.h>
> 
> int
> main(void)
> {
> 	char *const  s = strdup("foo");
> 
> 	printf("%s\n", strcpy(s, "bar"));
> }
> alx@debian:~/tmp/c$ cc -Wall -Wextra const.c 
> alx@debian:~/tmp/c$ ./a.out 
> bar
> ```
> 
> No type-safety violations here, according to the compiler, despite the use of `const`.

Sure, it's not a silver bullet...

alejandro-colomar:
> Of course if you define s in probably the more common way (albeit with a different meaning) of
> 
> ```c
> const char *s = strdup("foo");
> ```
> 
> then you do get the compiler warning

Indeed.  But that's *not* what you're doing in this patch.  What you're doing is apply `const` to the pointer, not the pointee.

```c
alx@debian:~/tmp/c$ cat const.c 
#include <stdio.h>

#define type_str(x)                                                           \
(                                                                             \
	_Generic(x,                                                           \
		char *:        "'char *'",                                    \
		const char *:  "'const char *'",                              \
		default:       "other"                                        \
	)                                                                     \
)

struct str_t {
	long  len;
	char  *start;
};

struct rstr_t {
	long        len;
	const char  *start;
};

int
main(void)
{
	static const struct str_t  sc = {3, "foo"};
	static struct str_t        s = {3, "foo"};
	const struct str_t         c = {3, "foo"};
	struct str_t               x = {3, "foo"};

	printf("sc: %s\n", type_str(sc.start));
	printf("s: %s\n", type_str(s.start));
	printf("c: %s\n", type_str(c.start));
	printf("x: %s\n", type_str(x.start));

	struct rstr_t              r = {3, "foo"};

	printf("r: %s\n", type_str(r.start));
}
alx@debian:~/tmp/c$ cc -Wall -Wextra const.c 
alx@debian:~/tmp/c$ ./a.out 
sc: 'char *'
s: 'char *'
c: 'char *'
x: 'char *'
r: 'const char *'
```
    
> 
> ```
> const.c: In function ‘main’:
> const.c:9:31: warning: passing argument 1 of ‘strcpy’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
>     9 |         printf("%s\n", strcpy(s, "bar"));
>       |                               ^
> In file included from const.c:2:
> /usr/include/string.h:141:39: note: expected ‘char * restrict’ but argument is of type ‘const char *’
>   141 | extern char *strcpy (char *__restrict __dest, const char *__restrict __src)
>       |                      ~~~~~~~~~~~~~~~~~^~~~~~
> ```
> 
> Yes, it still modifies s, but at least you're warned about it...

alejandro-colomar:
Or put a different way:

```c
const struct a { char         *x; } A;
struct b       { char *const  x; }  B;
struct c       { const char   *y; } C;
```

The following two are equivalent: `A.x` and `B.x` are both `char *const` (which easily degrades to `char *` after undergoing lvalue conversion (see ISO C reference a few comments above).
`C.x` is `const char *`, which is the safe thing; that `const` is enforced by the compiler.

ac000:
Not in that exact case, no, but in the general case where you pass a const pointer into a function then pass it non-const into another.

alejandro-colomar:
> Not in that exact case, no, but in the general case where you pass a const pointer into a function then pass it non-const into another.

A const pointer isn't safe at all.

```c
alx@debian:~/tmp/c$ cat const.c 
#include <string.h>

void f(char * p);

void g(char *const q)
{
	strcpy(q, "foo");
}
alx@debian:~/tmp/c$ cc -Wall -Wextra const.c -S
alx@debian:~/tmp/c$ 
```

What is safe is a const pointee (aka, pointer to const).

Your calls having `const nxt_str_t *s` protect against passing to other functions accepting `nxt_str_t *s`, but you can easily change the prototype of those to use `const nxt_str_t *s`, and there won't be any warnings at all.

I suggest you do the following:

Apply a patch on top of your suggested patch, modifying a string inside a function that you modified to supposedly add warnings against that.  You can for example modify the first byte of a string to be `/`.  And then show me any warnings you get.  I bet there won't be any.  So, please show what warnings this patch is adding, with real examples.

ac000:
I think you're arguing for arguments sake now, my _final_ reply to you is this...

1) Those nxt_str_t's should not be just statics, that makes no sense. static const does and matches others in the code.
2) Using unit idioms

Current way

```c
#include <stdio.h>
#include <stddef.h>

typedef struct {
        size_t  length;
        char    *start;
} nxt_str_t;

#define nxt_length(s)           (sizeof(s) - 1)
#define nxt_string(str)         { nxt_length(str), (char *)str }

#define nxt_str_set(str, text)                                                \
        do {                                                                  \
                (str)->length = nxt_length(text);                             \
                (str)->start = (char *)text;                                  \
        } while (0)

int main(void)
{
        static nxt_str_t success_str = nxt_string("success");

        printf("%s\n", success_str.start);

        nxt_str_set(&success_str, "test");
        printf("%s\n", success_str.start);

        return 0;
}
```

```
$ gcc -Wall -Wextra -O2 -g -o s2 s2.c
$ ./s2
success
test
```

With

```c
static const nxt_str_t success_str = nxt_string("success");
```

```
$ gcc -Wall -Wextra -O2 -g -o s2 s2.c
s2.c: In function ‘main’:
s2.c:14:31: error: assignment of member ‘length’ in read-only object
   14 |                 (str)->length = nxt_length(text);                             \
      |                               ^
s2.c:24:9: note: in expansion of macro ‘nxt_str_set’
   24 |         nxt_str_set(&success_str, "test");
      |         ^~~~~~~~~~~
s2.c:15:30: error: assignment of member ‘start’ in read-only object
   15 |                 (str)->start = (char *)text;                                  \
      |                              ^
s2.c:24:9: note: in expansion of macro ‘nxt_str_set’
   24 |         nxt_str_set(&success_str, "test");
```

I'm sure there's ways of defeating it, but this is definitely an improvement.

Anyway it accomplished my goal, so job done...

alejandro-colomar:
> I think you're arguing for arguments sake now,

Not really.

>     1. Those nxt_str_t's should not be just statics, that makes no sense. static const does and matches others in the code.

Okay.

>     2. Using unit idioms
> 
> 
> Current way
> 
> ```c
> #include <stdio.h>
> #include <stddef.h>
> 
> typedef struct {
>         size_t  length;
>         char    *start;
> } nxt_str_t;
> 
> #define nxt_length(s)           (sizeof(s) - 1)
> #define nxt_string(str)         { nxt_length(str), (char *)str }
> 
> #define nxt_str_set(str, text)                                                \
>         do {                                                                  \
>                 (str)->length = nxt_length(text);                             \
>                 (str)->start = (char *)text;                                  \
>         } while (0)
> 
> int main(void)
> {
>         static nxt_str_t success_str = nxt_string("success");
> 
>         printf("%s\n", success_str.start);
> 
>         nxt_str_set(&success_str, "test");
>         printf("%s\n", success_str.start);
> 
>         return 0;
> }
> ```
> 
> ```
> $ gcc -Wall -Wextra -O2 -g -o s2 s2.c
> $ ./s2
> success
> test
> ```
> 
> With
> 
> ```c
> static const nxt_str_t success_str = nxt_string("success");
> ```
> 
> ```
> $ gcc -Wall -Wextra -O2 -g -o s2 s2.c
> s2.c: In function ‘main’:
> s2.c:14:31: error: assignment of member ‘length’ in read-only object
>    14 |                 (str)->length = nxt_length(text);                             \
>       |                               ^
> s2.c:24:9: note: in expansion of macro ‘nxt_str_set’
>    24 |         nxt_str_set(&success_str, "test");
>       |         ^~~~~~~~~~~
> s2.c:15:30: error: assignment of member ‘start’ in read-only object
>    15 |                 (str)->start = (char *)text;                                  \
>       |                              ^
> s2.c:24:9: note: in expansion of macro ‘nxt_str_set’
>    24 |         nxt_str_set(&success_str, "test");
>       |         ^~~~~~~~~~~
> ```

Hmmm, yeah, that's a valid concern.  This patch is insufficient, but still good.

> I'm sure there's ways of defeating it, but this is definitely an improvement.

Agreed.

ac000:
`const`'d a couple of `static struct`'s

```
$ git range-diff 76874b93...06644f92
1:  3faf1f04 ! 1:  33547a50 Constify a bunch of static local variables
    @@ src/nxt_conf_validation.c: nxt_conf_vldt_if(nxt_conf_validation_t *vldt, nxt_con
      
          if (nxt_conf_type(value) != NXT_CONF_STRING) {
              return nxt_conf_vldt_error(vldt, "The \"if\" must be a string");
    +@@ src/nxt_conf_validation.c: nxt_conf_vldt_action(nxt_conf_validation_t *vldt, nxt_conf_value_t *value,
    +     nxt_conf_value_t        *action;
    +     nxt_conf_vldt_object_t  *members;
    + 
    +-    static struct {
    ++    static const struct {
    +         nxt_str_t               name;
    +         nxt_conf_vldt_object_t  *members;
    + 
     @@ src/nxt_conf_validation.c: nxt_conf_vldt_pass(nxt_conf_validation_t *vldt, nxt_conf_value_t *value,
          nxt_int_t  ret;
          nxt_str_t  segments[3];
    @@ src/nxt_conf_validation.c: nxt_conf_vldt_app(nxt_conf_validation_t *vldt, nxt_st
     -    static nxt_str_t  type_str = nxt_string("type");
     +    static const nxt_str_t  type_str = nxt_string("type");
      
    -     static struct {
    +-    static struct {
    ++    static const struct {
              nxt_conf_vldt_handler_t  validator;
    +         nxt_conf_vldt_object_t   *members;
    + 
     @@ src/nxt_conf_validation.c: nxt_conf_vldt_php(nxt_conf_validation_t *vldt, nxt_conf_value_t *value,
      {
          nxt_conf_value_t  *targets;
2:  bd28f4fd = 2:  591ea805 php: Constify some local static variables
3:  b4d2153a = 3:  402c83c0 configuration: Constify more pointers
4:  76874b93 = 4:  06644f92 Tighten up some string arrays
```

ac000:
@hongzhidao 

I'm going to leave `nxt_conf_get_path()` (and the other one I can't remember) as they are for now...

I'm not planning any more additions to this particular pull-request.

hongzhidao:
> @hongzhidao
> 
> I'm going to leave `nxt_conf_get_path()` (and the other one I can't remember) as they are for now...
> 
> I'm not planning any more additions to this particular pull-request.

Makes sense. Feel free to ship it, LGTM.

ac000:
> > @hongzhidao
> > I'm going to leave `nxt_conf_get_path()` (and the other one I can't remember) as they are for now...
> > I'm not planning any more additions to this particular pull-request.
> 
> Makes sense. Feel free to ship it, LGTM.

Thanks! @hongzhidao 

Although I can't until you _actually_ **approve** it! ;)

While I'm here, do I have your permission to add

```
Reviewed-by: Zhidao HONG <z.hong@f5.com>
```

to each commit?

hongzhidao:
Sure, you are welcome to add it.

ac000:
Rebase with master

```
$ git range-diff 06644f92...cd9c4e35
 -:  -------- >  1:  678c0568 Fixes: 64934e59f ("HTTP: Introduce quoted target marker in HTTP parsing") Reviewed-by: Andrew Clayton <a.clayton@nginx.com>
 -:  -------- >  2:  a48fbc03 Add additional information to the README
 -:  -------- >  3:  d7ce3569 Elaborate on docker image differences
 -:  -------- >  4:  237a26aa wasm-wc: Bump the rustls crate from 0.21.10 to 0.21.11
 -:  -------- >  5:  3fbca6ca Fix some trailing whitespace and long lines in the README
 1:  b7e95efa =  6:  269459a8 configuration: Constify numerous pointers
 2:  33547a50 =  7:  5ed112ae Constify a bunch of static local variables
 3:  591ea805 =  8:  719c6d7a php: Constify some local static variables
 4:  402c83c0 =  9:  13cedde3 configuration: Constify more pointers
 5:  06644f92 = 10:  cd9c4e35 Tighten up some string arrays
```

ac000:
Add Zhidao's `Reviewed-by:`'s

```
$ git range-diff cd9c4e35...b26c119f
1:  269459a8 ! 1:  e5bc299d configuration: Constify numerous pointers
    @@ Commit message
         are currently defined with 'static' storage class and turn them into
         'static const', which will be done in a subsequent patch.
     
    +    Reviewed-by: Zhidao HONG <z.hong@f5.com>
     
      ## src/nxt_conf.c ##
2:  5ed112ae ! 2:  8f861cf4 Constify a bunch of static local variables
    @@ Commit message
     
         This handles core code conversion.
     
    +    Reviewed-by: Zhidao HONG <z.hong@f5.com>
     
3:  719c6d7a ! 3:  33c978cc php: Constify some local static variables
    @@ Commit message
         Not sure why static, as they were being treated more like string
         literals, let's actually make them constants (qualifier wise).
     
    +    Reviewed-by: Zhidao HONG <z.hong@f5.com>
     
      ## src/nxt_php_sapi.c ##
4:  13cedde3 ! 4:  31cec908 configuration: Constify more pointers
    @@ Commit message
     
         We can get around the warning by casting ->u.{str,string}.start
     
    +    Reviewed-by: Zhidao HONG <z.hong@f5.com>
     
      ## src/nxt_conf.c ##
5:  cd9c4e35 ! 5:  b26c119f Tighten up some string arrays
    @@ Commit message
     
         This is the normal way of declaring such things.
     
    +    Reviewed-by: Zhidao HONG <z.hong@f5.com>
     
      ## src/nxt_controller.c ##
```

ac000:
@hongzhidao 

I've re-requested your review as it can be hard to find the 'Approve button` again...

callahad:
I'll hit the approve button since it's late for Zhidao and from his comments he's clearly happy with this change.

callahad (approved):

ac000:
Thanks Dan!

PR #1225, opened 2024-04-19
Replicate (mostly) buildbot workflow for commits into master

oxpa:
Hello,

This pull request utilizes  nginx self hosted runners to build unit and it's modules on a variety of distros and two archs (for the lack of others for now).

The proposed CI creates 252 jobs which can be a bit excessive and may require some optimisations: some jobs don't really do what they should do just skipping test. It's mostly true for python, java and php where each distro has a number of version available and I was too lazy to get a proper configuration for what plays where.

I also replicated configuration management system by creating `setup.sh` which installs packages from a number of lists. And I created those per-OS list. The list is a bit excessive and replicates current buildbot workers. So there is some room for optimisations there as well.

The existing CI is left active for pull requests as it provides relatively fast way of checking code for obvious problems.


.github/workflows/ci-pull.yml
```diff
@@ -1,10 +1,8 @@
 name: ci
 
 on:
+  workflow_dispatch:
   pull_request:
-  push:
-    branches:
-      - master
```
callahad, line 7: I think we'd still want our PR checks to run on master to catch and obvious failures (e.g., PR CI passes separately for branch A and B independently, but not for the result of merging both into master)

oxpa: PR checks are more or less a subset of what this proposed CI does. We can still run it for a commit but it should not add any value

callahad: Limited value for us, but external forks won't have access to our self-hosted runners, and I'd want them to have sufficiently functional CI out-of-the-box.

oxpa: Not quite sure I fully understand the concern but I re-enabled the workflow for commits into master branch.

.github/scripts/setup.sh
```diff
@@ -0,0 +1,260 @@
+#/bin/sh -x
+
+DIR=`dirname $0`
+DISTRO="$1"; shift;
+
+# package management helpers
+PACKAGE_MANAGER_INSTALL_yum="yum --setopt=skip_missing_names_on_install=False install -y"
+PACKAGE_MANAGER_REMOVE_yum="yum erase -y"
```
ac000, line 8: Once we drop CentOS 7 (EOL in a couple of months IIRC) we should replace yum with dnf (it replaced yum several years ago), I don't even have yum on my Fedora 39 boxes.

.github/workflows/ci-commit.yml
```diff
@@ -0,0 +1,345 @@
...
+      fail-fast: false
+      matrix:
+        os: 
+          - ubuntu-22.04
+          - ubuntu-23.10
+          - debian-11
+          - debian-12
+          - alpine-3.19
+          - rhel-9
+          - amazonlinux-2023
```
ac000, line 23: There are more package lists than what's listed here, is that intentional?

ac000:
Some observations

1) Does it make sense to still rename ci.yml?, seeing as it isn't limited to pull-requests.
2) The last patch should just be folded into the first or
3) Do we even need those commits (first & last) now?

.github/scripts/setup.sh
callahad: Looks like RedHat rebrands `dnf` to `yum` in RHEL 8: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/managing-software-packages_configuring-basic-system-settings

> Upstream documentation identifies the technology as DNF and the tool is referred to as DNF in the upstream.

callahad:
I agree with Andrew's comments above; LGTM once those are addressed.

.github/scripts/setup.sh
oxpa: We can fix this later, I guess. 
As per fedora, in a default rawhide container `yum` is still a symlink to `dnf`, so I don't expect any problems with earlier releases either.
```
:~$ docker run --rm -it fedora:rawhide /usr/bin/yum --version
4.19.0
  Installed: dnf-0:4.19.0-1.fc40.noarch at Mon Feb 19 08:10:37 2024
  Built    : Fedora Project at Thu Feb  8 16:33:42 2024

  Installed: rpm-0:4.19.1.1-1.fc40.x86_64 at Mon Feb 19 08:10:37 2024
  Built    : Fedora Project at Wed Feb  7 15:55:53 2024
```

.github/workflows/ci-commit.yml
oxpa: Yes and no: there is `alpine-3.18` list which I failed to use because there is no alpine 3.18 image for self hosted runners. So this one can be removed.
At the same time, having general lists (`debian`, `alpine`, `ubuntu`, `rhel`) is intentional: one can install a package for all versions of an OS family.

oxpa:
@ac000 
1. I'm happy to rename files further. See no reason to change their names but happy to rename it whatever you prefer.
2. I expect the purll request to be squash-merged. See no reason to keep this history of commits
3. Same as above: there should be a single commit in the repository.

.github/scripts/setup.sh
ac000: > Looks like RedHat rebrands `dnf` to `yum` in RHEL 8: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/managing-software-packages_configuring-basic-system-settings
> 
> > Upstream documentation identifies the technology as DNF and the tool is referred to as DNF in the upstream.

Well, it's a _version_ of yum that is compatible with dnf (and older yum).

ac000: > We can fix this later, I guess. As per fedora, in a default rawhide container `yum` is still a symlink to `dnf`, so I don't expect any problems with earlier releases either.
> 
> ```
> :~$ docker run --rm -it fedora:rawhide /usr/bin/yum --version
> 4.19.0
>   Installed: dnf-0:4.19.0-1.fc40.noarch at Mon Feb 19 08:10:37 2024
>   Built    : Fedora Project at Thu Feb  8 16:33:42 2024
> 
>   Installed: rpm-0:4.19.1.1-1.fc40.x86_64 at Mon Feb 19 08:10:37 2024
>   Built    : Fedora Project at Wed Feb  7 15:55:53 2024
> ```

Maybe I removed it...

```
$ yum
bash: yum: command not found
$ rpm -q yum
package yum is not installed
```

.github/workflows/ci-commit.yml
ac000: rhel-8 also seems to be in the lists but not the workflow file...

.github/workflows/ci-pull.yml
```diff
@@ -1,10 +1,8 @@
 name: ci
 
 on:
+  workflow_dispatch:
```
ac000: Do we still need this change? Things _seem_ to work without it...

oxpa: This allows you to run a workflow manually on any branch. IMO it's useful and doesn't break anything.

ac000:
>     2. I expect the purll request to be squash-merged. See no reason to keep this history of commits

_If_ you keep the ci.yml change, then that should at least be a separate commit (with an explanation of why it's needed).

.github/workflows/ci-pull.yml
arbourd: Allows us to kick off builds without actually running a change, manually. Can be really useful!

ac000: OK, thanks for the explanation.

Then yeah, that should be done as a separate commit (it's not related to the other changes).

arbourd (approved):
This looks great! Assuming it works tho, as the job wasn't able to kick off within the PR, obviously.

If there are too many jobs, the long-running self hosted environments could run all tests together at once, or at least, matrix against every module-lang rather than module-lang-version. But that can always be optimized later.

I'm happy to help with some of those optimizations if needed, as well.

oxpa:
@arbourd Running too may tests is my main concern, really. I failed to find an easy way of pinning os-lang-version triplet while this will reduce the amount of jobs significantly. I'm happy to adopt any solution (maybe just hardcode it in?).
For now, I'm going to split the workflow into two parts, one per arch to avoid "no more than 256 jobs" limit.
We can also test multiple modules at once (say, go, php, java) on a single run. But it's more complicated with multiple versions of a single module (java-11, java-17, java-21). I'm happy to hear any proposals here as well.

thresheek:
Is it a such a good idea to create a job for every "$OS-$ARCH-$LANG"?  I think if we're to reuse the buildbot approach, we should be fine with "$OS-$ARCH" and then run steps for different $LANGs... this will save us quite some overhead in spinning up the machines.

.github/workflows/ci-commit.yml
```diff
@@ -0,0 +1,345 @@
+name: ci
```
ac000: This needs a more descriptive name...

.github/workflows/ci-commit.yml
```diff
@@ -0,0 +1,345 @@
...
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
```
ac000: I do sort of question this, as checking once things have hit master is a bit late...

.github/workflows/ci-pull.yml
ac000: The file renaming part of this commit should be dropped.

.github/workflows/ci-selfhosted-or-smth-like-that.yml
ac000, line 1: Can we just name it correctly from the start?

.github/workflows/ci.yml
ac000, line 1: I think this should be dropped. I.e this file should never have changed its name...

.github/workflows/ci-selfhosted.yml
ac000, line 1: Again, can we name this correctly from the start?

.github/workflows/ci-commit.yml
thresheek: We can't have self-hosted runners in a PR environment.  Also, all of this should move to a separate repo with added protections on who can edit and call the CI steps on the workflow.

thresheek:
I've come up with a slightly (much) easier workflow, not too dissimilar from what we currently use for njs (https://github.com/nginx/njs/blob/master/.github/workflows/ci.yml and https://github.com/nginx/ci-self-hosted/blob/main/.github/workflows/njs-buildbot.yml). 

I'll follow up with a PR for that soon-ish.

callahad:
Thanks for the heads up; closing in lieu of the new approach.

PR #1230, opened 2024-04-23
fs: Use branchless code

alejandro-colomar:
That branch was to avoid an infinite loop on the slash.  However, we can achieve the same by using a +1 to make sure we advance at least 1 byte in each iteration.

Cc: @ac000

### DIFF
--- a/src/nxt_fs.c
+++ b/src/nxt_fs.c
@@ -23,11 +23,7 @@ nxt_fs_mkdir_all(const u_char *dir, mode_t mode)
     start = (char *) dir;
 
     while (*start != '\0') {
-        if (*start == '/') {
-            *dst++ = *start++;
-        }
-
-        end = strchr(start, '/');
+        end = strchr(start + 1, '/');
         if (end == NULL) {
             end = ((char *)dir + dirlen);
         }


alejandro-colomar:
Not tested.

alejandro-colomar:
> On the face of it change looks good.
> 
> As an aside I created a new git/github workflow document (Dan's OK'd it, just hasn't been _publicly_ released yet) anyway the point is in it I recommend to use lower case for the topic(s), some look weird capitalised and some simply shouldn't be. It also matches other projects...

Okay; both styles are similarly wide spread.  In the Linux man-pages, Michael's rule was that commit messages, when removing the prefix, should be correct English sentences (except that we omit the trailing period), but of course it's arbitrary to omit the period but not omit the capital, so I guess also omitting the capital is fine.

I'll adapt the patch.

alejandro-colomar:
v1b changes:

-  lowcase

```
$ git range-diff master gh/mkdirp mkdirp 
1:  77340b39 ! 1:  88ddba5e Fs: Use branchless code
    @@ Metadata
     
      ## Commit message ##
    -    Fs: Use branchless code
    +    Fs: use branchless code
     
         That branch was to avoid an infinite loop on the slash.  However, we can
         achieve the same by using a +1 to make sure we advance at least 1 byte
```

ac000:
Heh, you've mis-understood.

It's the **topics(s)** that should be lower case, I.e

```
fs: Use branchless code
```

ac000:
Tested with cgroups and isolation/rootfs (hopefully the pytests also tested it!)

```
Tested-by: Andrew Clayton <a.clayton@nginx.com>
```

alejandro-colomar:
> Heh, you've mis-understood.
> 
> It's the **topic(s)** that should be lower case, I.e
> 
> ```
> fs: Use branchless code
> ```

Ahh, yup!  That sounds much better!!  I never understood why one would sentence-case a prefix :)

alejandro-colomar:
> Tested with cgroups and isolation/rootfs (hopefully the pytests also tested it!)
> 
> ```
> Tested-by: Andrew Clayton <a.clayton@nginx.com>
> ```

Lovely!  Thanks.

alejandro-colomar:
v1c changes:

-  case
-  tested

```
$ git range-diff master gh/mkdirp mkdirp 
1:  88ddba5e ! 1:  163309e0 Fs: use branchless code
    @@ Metadata
     
      ## Commit message ##
    -    Fs: use branchless code
    +    fs: Use branchless code
     
         That branch was to avoid an infinite loop on the slash.  However, we can
         achieve the same by using a +1 to make sure we advance at least 1 byte
         in each iteration.
     
    +    Tested-by: Andrew Clayton <a.clayton@nginx.com>
     
      ## src/nxt_fs.c ##
```

alejandro-colomar:
Closing in favour of <https://github.com/nginx/unit/pull/1235>

PR #1237, opened 2024-04-29
No need to explicitly set -fno-strict-aliasing in the wasm module now

ac000:
Since commit 0b5223e1c ("Disable strict-aliasing in clang by default") we explicitly always build with -fno-strict-aliasing so there's no need to set it independently in auto/modules/wasm


hongzhidao (approved):

ac000:
Rebased with master

PR #1252, opened 2024-05-07
Packages: added Ubuntu 24.04 "noble" support.

en-jschuetze:
To make it possible to use the official nginx unit packages on the latest LTS from ubuntu, we need to have this packages build.

I searched through the repo and found this Makefile for building the packages. But afterwards we need to change also https://github.com/nginx/unit-docs/edit/master/source/installation.rst and see if the files really apply and are available at 
https://packages.nginx.org/unit/ubuntu/dists/

### DIFF
--- a/pkg/deb/Makefile
+++ b/pkg/deb/Makefile
@@ -19,6 +19,23 @@ BUILD_DEPENDS = $(BUILD_DEPENDS_unit)
 
 MODULES=
 
+# Ubuntu 24.04
+ifeq ($(CODENAME),noble)
+include Makefile.php
+include Makefile.python311
+include Makefile.python312
+include Makefile.go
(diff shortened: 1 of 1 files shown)


thresheek:
I think it needs more changes to make it buildable on 24.04, e.g. FORTIFY_SOURCE needs a dance similar to what we implemented in nginx: https://hg.nginx.org/pkg-oss/rev/ce74bf99bf45

Also, I think the list of included Makefiles needs to be adapted to reflect the reality - e.g. there is no Python 3.11 shipped with 24.04; same for Java 19 and 20 as far as I can tell.

thresheek:
A correct version is submitted at https://github.com/nginx/unit/pull/1287

PR #1256, opened 2024-05-09
Add unitctl build and release CI

arbourd:
Adds a GitHub Actions workflow that builds and releases `unitctl` binaries when a tag prefixed with `unitctl/` is pushed.

Binaries are built on pull-requests that change any files within `tools/unitctl`, on `master` branch pushes and when `unitctl/` prefixed tags are pushed.

Screenshot 2024-05-09 at 2 30 03 PM

### DIFF
--- a/.github/workflows/unitctl.yml
+++ b/.github/workflows/unitctl.yml
@@ -1,0 +1,132 @@
+
+name: unitctl
+
+on:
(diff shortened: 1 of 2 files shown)


arbourd:
Accidentally pushed to wrong remote, brb

PR #1257, opened 2024-05-09
Add unitctl build and release CI

arbourd:
Adds a GitHub Actions workflow that builds and releases unitctl binaries when a tag prefixed with `unitctl/` is pushed.

Binaries are built on pull-requests that change any files within `tools/unitctl`, on `master` branch pushes and when `unitctl/` prefixed tags are pushed.


avahahn (approved):
This is more or less perfect in my opinion. I will remove the extra newline from the Readme and then take over incorporating feedback on this!

danielledeleo (approved):
I tested the macOS artifacts locally and they appear to be good.

callahad:
On the `unitctl/x.y.z` tag namespace... Do we want to have distinct versioning for unitctl, or do we want to bind its version to Unit releases?

ac000:
To echo Dan's sentiments.

We have tags depicting Unit releases. Do we want to pollute that with *other* tags?.

Is there a *very* good reason not to just tie it off Unit releases?.

I.e I expect to see (and I don't know if we have tooling that expects the same)

```
$ git describe
1.32.0-72-g5d1ce5c4
```

The only thing you *could* do is use a *lightweight* tag, but that's not really correct as they aren't supposed to be distributed (although we do use them for the packaging tags), but I think you need a *really* good reason for it to have its own tags.

avahahn:
I am happy to tie it to unit releases. Currently testing a patch to do just that.

ac000:
What was the rush?

This should have been committed as a single commit.

We don't commit things in a known broken/incorrect/intermediary state immediately followed by fix up commits...

tools/unitctl/unit-openapi/README.md
```diff
@@ -409,3 +409,4 @@ cargo doc --open
 ## Author
 
 unit-owner@nginx.org
+
```
ac000, line 412: What is it with this file!?

I thought we'd left that trailing newline in and then ignored it via a git attribute?

.github/workflows/unitctl.yml
```diff
@@ -0,0 +1,132 @@
...
+    paths:
+      - tools/unitctl/**
+  push:
+    branches:
+      - master
+    tags:
+      - unitctl/[0-9]+.[0-9]+.[0-9]+
+
+permissions:
+  contents: write
```
ac000, line 15: What exactly is given write permissions to?

ac000 (commented):
```
 trigger unitctl CI on version tags of existing format

Signed-off-by: Ava Hahn <a.hahn@f5.com>
```

Subjects (other than any prefix) should really be capitalised.

.github/workflows/unitctl.yml
```diff
@@ -0,0 +1,132 @@
...
+        with:
+          tool: cross
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          prefix-key: rust-${{ matrix.build }}
+          workspaces: ./tools/unitctl -> target
+          save-if: ${{ github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/unitctl/') }}
+
+      - name: Configure linux arm depedencies
```
ac000, line 63: s/depedencies/dependencies/

.github/workflows/unitctl.yml
```diff
@@ -0,0 +1,132 @@
...
+          pre-build = [
+              "dpkg --add-architecture \$CROSS_DEB_ARCH",
+              "apt-get update && apt-get install --assume-yes libssl-dev:\$CROSS_DEB_ARCH"
+          ]
+          EOF
+
+          cat Cross.toml
+          echo "CARGO=cross" >> $GITHUB_ENV
+
+      - name: Install macOS depedencies
```
ac000, line 77: s/depedencies/dependencies/

.github/workflows/unitctl.yml
arbourd: Allows the job to create or update the release. It's opt-in now.

> Work with the contents of the repository. For example, contents: read permits an action to list the commits, and contents: write allows the action to create a release. For more information, see "[Permissions required for GitHub Apps](https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents)."

tools/unitctl/unit-openapi/README.md
arbourd: That was actually my auto-ensure-trailing-newline func that comes with VS Code, but I actually had it in source to test the actual build (as it needs to be triggered by changes in that folder).

.github/workflows/unitctl.yml
```diff
@@ -0,0 +1,132 @@
+
+name: unitctl
+
+on:
+  pull_request:
+    paths:
+      - tools/unitctl/**
```
ac000, line 7: Does this actually need to run on pull requests?

For releases it will surely only be once a unit release tag is pushed to the repository?

.github/workflows/unitctl.yml
ac000: So it doesn't have write access to the repository itself?

.github/workflows/unitctl.yml
arbourd: I don't why you wouldn't want it to? It won't kick off the release but it will attempt to compile `unitctl`.

If / when there are tests, they can also be added as a step that runs on PRs.

ac000: Fair enough.

.github/workflows/unitctl.yml
arbourd: It allows the action to write to any of these endpoints: https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents

.github/workflows/unitctl.yml
```diff
@@ -9,7 +9,7 @@ on:
     branches:
       - master
     tags:
-      - unitctl/[0-9]+.[0-9]+.[0-9]+
+      - '[0-9]+.[0-9]+.[0-9]+'
```
ac000, line 12: I'm not sure exactly what regex syntax this is using but for example with PCRE that would also match the packaging tags (_if_ it matches against lightweight tags) e.g

```
$ echo "1.32.1-1" | grep -P '[0-9]+.[0-9]+.[0-9]+'
1.32.1-1
```

with PCRE at least you'd want `^[0-9]+\.[0-9]+\.[0-9]+$`

E.g
```
$ echo "1.32.1-1" | grep -P '^[0-9]+\.[0-9]+\.[0-9]+$'
$
```
We also want to _match_ the `.`'s, not any character... (though looking at the GH docs is _looks_ like `.` doesn't have that meaning here...)

(in our case the `^` is technically superfluous, but doesn't hurt and at this point we'd technically only need to check if the first digit is [1-9])

.github/workflows/unitctl.yml
ac000: Wow!

Is there anyway to give more fine-grained access?

.github/workflows/unitctl.yml
avahahn: Its not real regex, its Github's special subset of regex. I did test using a more robust regex:
```
([0-9]+\.[0-9]+\.[0-9]+)(-[0-9]+)?
```
But found during testing that this wasnt actually regex that this was parsing.

PR #1258, opened 2024-05-10
Add unit testing to Unitctl CI workflow

avahahn:
* fix a few misspellings in unitctl CI workflow
* add unit testing job
* disable one test 
* add additional triggers for CI

About the disabled test: It tests that the CLI can run a local copy of the unitd binary. This would require a custom image to run in CI. If we find it necessary, I am happy to work on it as an additional changeset.

### DIFF
--- a/.github/workflows/unitctl.yml
+++ b/.github/workflows/unitctl.yml
@@ -10,11 +10,67 @@ on:
       - master
     tags:
       - '[0-9]+.[0-9]+.[0-9]+'
+  workflow_dispatch:
+    inputs:
+      version:
+        type: string
+        description: "Semver tag"
+        required: true
 
 permissions:
   contents: write
 
 jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        working-directory: tools/unitctl
+    env:
+      MAKE: make
+      CARGO: cargo
+      VERSION:
+      SHORT_VERSION:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - build: linux-x86_64
+            os: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+          - build: macos-aarch64
+            os: macos-latest
+            target: aarch64-apple-darwin
+    steps:
+      - uses: actions/checkout@v4
+
+      - run: rustup update stable
+      - run: rustup target add ${{ matrix.target }}
+
+      - name: Install cross
+        if: matrix.target == 'aarch64-unknown-linux-gnu'
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cross
+
+      - name: Install macOS depedencies
+        if: startsWith(matrix.os, 'macos')
+        run: |
+          brew install make gnu-sed grep gawk
+          echo "MAKE=gmake" >> $GITHUB_ENV
+
+      - uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 21
+
+      - run: ${{ env.MAKE }} list-targets
+
+      - name: Generate openapi
+        run: ${{ env.MAKE }} openapi-generate
+      - name: Test ${{ matrix.os }}
+        run: ${{ env.MAKE }} test
+
   build:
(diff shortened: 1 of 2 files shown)


tools/unitctl/unit-openapi/README.md
```diff
@@ -409,4 +409,3 @@ cargo doc --open
 ## Author
 
 unit-owner@nginx.org
-
```
ac000, line 412: No.

Leave the blank line in there otherwise we're going to just be ping-ponging on this file in every PR. That's why we introduced the .`gitattributes` file in that directory.

ac000 (changes requested):

.github/workflows/unitctl.yml
```diff
@@ -15,6 +15,50 @@ permissions:
...
+  test:
+    runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        working-directory: tools/unitctl
+    env:
+      MAKE: make
+      CARGO: cargo
+      VERSION:
+      SHORT_VERSION:
```
arbourd, line 27: If we aren't doing cross we shouldn't need these, correct? The VERSION stuff is for releasing either way.

.github/workflows/unitctl.yml
```diff
@@ -15,6 +15,50 @@ permissions:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - build: linux-x86_64
+            os: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+          - build: macos-x86_64
+            os: macos-latest
+            target: x86_64-apple-darwin
```
arbourd, line 37: `macOS-latest` is actually `aarch64` now by default. Would it make sense to just build that? Might be faster?

By default a macOS arm machine can build and execute either arm64 or amd64 thanks to [Rosetta](https://en.wikipedia.org/wiki/Rosetta_(software)#Rosetta_2).

arbourd (approved):
Dope!

.github/workflows/unitctl.yml
avahahn: Fair enough! I think it absolutely would be reasonable!

.github/workflows/unitctl.yml
avahahn: we are using cross!

.github/workflows/unitctl.yml
```diff
@@ -10,6 +10,13 @@ on:
       - master
     tags:
       - '[0-9]+.[0-9]+.[0-9]+'
+      - '[0-9]+.[0-9]+.[0-9]+-[0-9]+'
```
ac000, line 13: I don't know if we need it for *both* of these.

The `-1` is a packaging thing and ideally will go away in the future...

ac000:
Are the various unitctl test failures an issue?

I guess they should either be fixed or skipped...

avahahn:
Here is what a test release looks like:
https://github.com/avahahn/unit/releases/tag/1.22.44

avahahn:
> Are the various unitctl test failures an issue?
> 
> I guess they should either be fixed or skipped...

All tests now passing.

.github/workflows/unitctl.yml
```diff
@@ -32,15 +38,21 @@ jobs:
           - build: linux-x86_64
             os: ubuntu-latest
             target: x86_64-unknown-linux-gnu
-          - build: macos-x86_64
+          - build: macos-aarch64
             os: macos-latest
-            target: x86_64-apple-darwin
+            target: aarch64-apple-darwin
```
ac000, line 43: Doesn't this change belong in the previous commit?

avahahn: Personally, I think it should all be squashed into one commit. 
I also am not sure how to add existing in-tree changes into a commit that is not HEAD

ac000 (approved):

.github/workflows/unitctl.yml
avahahn: Okay! I can make sure to have presentation in order before making future merge requests.

Issue #1234, opened 2024-04-24
Bug: Proxy validation


Pavlusha311245:
### Nginx Unit version: 1.32.1

Validation error when setting action proxy. http://host.docker.internal:7600 fails verification

### Config inside my docker container:

I tried to use this inside a container to communicate with the host, but there is no way to simply load it :)

```json
{
  "listeners": {
    "*:80": {
      "pass": "routes/test"
    },
  },
  "routes": {
    "test": [
      {
        "action": {
          "proxy": "http://host.docker.internal:7600"
        }
      }
    ],
  }
}
```

### Error:
```json
{
        "error": "Invalid configuration.",
        "detail": "The \"proxy\" address is invalid \"host.docker.internal:7600\""
}
```

andrey-zelenkov:
Hi @Pavlusha311245,

Unit doesn't support DNS resolving at the moment. More information here: https://github.com/nginx/unit/issues/779

Pavlusha311245:
> Hi @Pavlusha311245,
> 
> Unit doesn't support DNS resolving at the moment. More information here: #779

Have Unit same problem with upstreams?

andrey-zelenkov:
> Have Unit same problem with upstreams?

Yes, no resolving anywhere.

Pavlusha311245:
@andrey-zelenkov I found new situation. Why I'm get an error while using http://172.24.0.3 instead of http://172.24.0.3:80?

andrey-zelenkov:
There is currently no default port specified.

Issue #779, opened 2022-11-01
Unable to proxy to docker container - invalid configuration (invalid is address)


JanMikes:
Hi, referred to the documentation https://unit.nginx.org/configuration/#proxying, i should be able to proxy request to any valid hostname, but dont know how to do it for docker container, defined and run in the same docker-compose file.

What im trying to do is to proxy all http requests from `localhost:8080/.well-known/mercure` to the `mercure` service, but i get 400 response from unit when running put request to config:
```log
/usr/local/bin/docker-entrypoint.sh: Looking for configuration snippets in /docker-entrypoint.d/...
/usr/local/bin/docker-entrypoint.sh: Applying configuration /docker-entrypoint.d/config.json
2022/11/01 21:02:15 [error] 12#12 invalid address "mercure"
/usr/local/bin/docker-entrypoint.sh: Error: HTTP response status code is '400'
{
	"error": "Invalid configuration.",
	"detail": "The \"proxy\" address is invalid \"mercure\""
}
```

docker-compose.yml (simplified):
```
version: "3"

services:
    web:
        image: unit
        ports:
            - "8080:8080"

    mercure:
        image: dunglas/mercure
```

config.json (simplified):
```json
{
    "listeners": {
        "*:8080": {
            "pass": "routes"
        }
    },

    "routes": [
        {
            "match": {
                "uri": "/.well-known/mercure"
            },
            "action": {
                "proxy": "http://mercure"
            }
        }
    ]
}

```

For `"proxy": "http://mercure"` i tried different values like `mercure`, `mercure:80`, `http://mercure:80/` and so on, but nothing worked, the result was the same.

I was able to get it working with traefik, but my goal was to not add additional reverse proxy and having Unit used for that.

If i can provide any further details that would help troubleshooting, please let me know.

Thank you!

tippexs:
@JanMikes we are currently working on this. At the moment we do not support any DNS or name bases addresses they need to be resolved. Only IPv4/6 and Unix domain sockets are valid inputs. Will flag this as bug and let you know as soon as a fix will be out.

JanMikes:
Thank you for such fast reply, i will be looking forward.

I can confirm that proxying to static IP works fine.

travisbell:
[...]

PR #1261, opened 2024-05-13
Proxy HTTP: Added standard port for inet and inet6

Pavlusha311245:
Added a standard port for IPv4 and IPv6 addresses when using proxy and HTTP, eliminating the need to specify an additional port 80. This is defined by the RFC standards for HTTP. Currently Nginx Unit only uses HTTP connections for proxying, version 1.1.

https://datatracker.ietf.org/doc/html/rfc2616#page-12
> HTTP communication usually takes place over TCP/IP connections. The default port is TCP 80

### DIFF
--- a/src/nxt_sockaddr.c
+++ b/src/nxt_sockaddr.c
@@ -667,7 +667,7 @@ nxt_sockaddr_inet6_parse(nxt_mp_t *mp, nxt_str_t *addr)
         p = NULL;
     }
 
-    port = 0;
+    port = 80;
 
     if (p != NULL) {
         length = (start + length) - p;
@@ -742,7 +742,7 @@ nxt_sockaddr_inet_parse(nxt_mp_t *mp, nxt_str_t *addr)
         }
     }
 
-    port = 0;
+    port = 80;
 
     if (p != NULL) {
         p++;


hongzhidao:
Hi @Pavlusha311245,
Thanks for the patches.
The `nxt_sockaddr.c` was supposed to parse IP:PORT address. It's a generic function but not depend on http stuff.
Proxy belongs to HTTP, and it makes sense to support the format you suggested as much as possible.
But the way that touched `nxt_sockaddr.c` is incorrect.
I feel the logic would be something like this:
```
sa = nxt_sockaddr_parse(&str);

if (sa == NULL) {
    result = getaddrinfo(...);
    /* process result */
}
```
https://www.man7.org/linux/man-pages/man3/getaddrinfo.3.html
But it would introduce another question, it looks like it's better to support domain names as well.
And it may resolve multiple addresses, this topic was discussed.
https://github.com/nginx/unit/issues/779
https://github.com/nginx/unit/issues/1234

Pavlusha311245:
@hongzhidao  Thanks for the review

In the current implementation, this is a quick fix that will allow you to at least allocate a default port and it is strictly aimed at only inet implementations excluding changes to unix at this stage. I think that we can start working with the proxy part of the web server :)

Regarding domains and https support, I think we should start a separate discussion and discuss further actions

hongzhidao:
I guess the current implementation may introduce issues like this:
```
{
    "listeners": {
        "127.0.0.1": {
             "pass": "routes"
        }
    }
}
```
It's an invalid configuration, but it's allowed with the patch.

Pavlusha311245:
Yes it is, but it is still HTTP, which works on port 80 by default. The ability to add and use another port has not gone away. And also does not interfere with the operation of unix sockets

hongzhidao:
I feel `nxt_sockaddr.c` has nothing to do with the scheme. The change should happen in `nxt_http_proxy.c` or another place that resolves the format `http://IP`. And these need to be based on the decision of whether or not to make the requirement. Will discuss it with the team.
Personally, I feel the requirement makes sense. Just to support the format `http://ip-address` for the proxing.

Pavlusha311245:
Then I will wait for your decision with the team. I think I can move this part only to the proxy section

hongzhidao:
> I think I can move this part only to the proxy section

Indeed.

Btw, the other idea I can think of right now is to do it like this in proxy module.
```
addr_str = concat(host_str, ':', default_port_str);
sa = sockaddr_parse(addr_str);
```

ac000:
To echo @hongzhidao 

I'm not against the idea, but I think the location is wrong. This should really happen in the proxy code.

hongzhidao:
Hi @Pavlusha311245,
> Then I will wait for your decision with the team. I think I can move this part only to the proxy section

We think it's a good idea to support the feature, welcome to continue.
Let me know if there is anything I can help.

Issue #445, opened 2020-07-20
411 Content-Length Required when requesting Transfer-Enconding Chunked


robjr:
Hi guys, I'm facing this issue where unit returns 411 content-length required when the request doesn't specify the content-length, but it actually specifies Transfer-Encoding: Chunked. 

```
    const formData = new FormData();
    formData.append('file', request.get('any.com/img.png'));
    await axios.post('svc.com/media', formData, { headers: formData.getHeaders() });
```
It is streaming a file from one end to another and it works locally, but whenever it goes through unit, it returns 411. Is there any configuration to enable transfer-encoding?

VBart:
Unit doesn't support chunked transfer encoding in requests for the moment. So, the only options right now is to specify the Content-Length.

mafin:
Good question - is there any configuration to change up `Transfer-Encoding`?

VBart:
@mafin It needs to be changed on the client.  There's no way to control client's request `Transfer-Encoding` from the server side.

saper:
Ouch, we just ran into this issue with [Axway API Gateway](https://support.axway.com/kb/176289/language/en) ...

saper:
> Hi @saper thanks for reaching out? What do you mean with that? Can you explain that? Thanks Timo

We started using unit for some REST services and now someone put Axway API manager in front of them - the result - unit always sends 411 Length Required, because API manager ALWAYS transforms requests with Content-Length into chunked encoding. 

We also could not easily tell unit to dump all requests and responses received, hard to trace the down (debug mode prints lots of syscall and thread info but no actual requests).

lcrilly:
Thanks for the extra context - that's really helpful. In this case, I suspect the Axway behavior is due to it's transformation features for things like JSON/XML conversion. It's a shame that it does the same even when no transformation occurs.

We strive to be RFC compliant and update the issue title to reflect that. For now, this is on the backlog, but no ETA just yet.

saper:
Why "Ice-Box"? Is something unclear about it?

I was thinking, isn't websocket code already doing something similar to chunked encoding?

ac000:
RFC 7230 has been obsoleted by RFCs [9110](https://datatracker.ietf.org/doc/html/rfc9110) & [9112](https://datatracker.ietf.org/doc/html/rfc9112)

lcrilly:
[...]

PR #1262, opened 2024-05-14
Support chunked request bodies

hongzhidao:
A draft patch for the MVP of this feature.

Internally, I transferred the chunked request to content-length request since the application process can't handle chunked request now.

### Test
conf.json
```
{
    "listeners": {
        "*:8080": {
            "pass": "routes"
        }
    },
    "routes": [
        {
            "action": {
                "pass": "applications/app"
            }
        }
    ],
    "applications": {
        "app": {
            "type": "python",
            "path": "/www",
            "module": "wsgi"
        }
    }
}
```
/www/wsgi.py
```
def application(environ, start_response):
    content_length = int(environ.get('CONTENT_LENGTH', 0))
    body = bytes(environ['wsgi.input'].read(content_length))
    start_response(
        '200', [],
    )
    return [body]
```
```
> curl -X POST -d 'hello' -H "Transfer-Encoding: chunked" http://127.0.0.1:8080 -v
Note: Unnecessary use of -X or --request, POST is already inferred.
* Rebuilt URL to: http://127.0.0.1:8080/
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8080 (#0)
> POST / HTTP/1.1
> Host: 127.0.0.1:8080
> User-Agent: curl/7.61.1
> Accept: */*
> Transfer-Encoding: chunked
> Content-Type: application/x-www-form-urlencoded
>
> 5
* upload completely sent off: 12 out of 5 bytes
< HTTP/1.1 200 OK
< Server: Unit/1.33.0
< Date: Wed, 29 May 2024 09:04:27 GMT
< Transfer-Encoding: chunked
<
* Connection #0 to host 127.0.0.1 left intact
hello
```

### DIFF
--- a/src/nxt_h1proto.c
+++ b/src/nxt_h1proto.c
@@ -839,7 +839,12 @@ nxt_h1p_transfer_encoding(void *ctx, nxt_http_field_t *field, uintptr_t data)
     if (field->value_length == 7
         && memcmp(field->value, "chunked", 7) == 0)
     {
+        if (r->chunked_field != NULL) {
+            return NXT_HTTP_BAD_REQUEST;
+        }
+
         te = NXT_HTTP_TE_CHUNKED;
+        r->chunked_field = field;
 
     } else {
         te = NXT_HTTP_TE_UNSUPPORTED;
@@ -856,14 +861,16 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
 {
     size_t             size, body_length, body_buffer_size, body_rest;
     ssize_t            res;
-    nxt_buf_t          *in, *b;
+    nxt_buf_t          *in, *b, *out, *chunk;
     nxt_conn_t         *c;
     nxt_h1proto_t      *h1p;
+    nxt_socket_conf_t  *skcf;
     nxt_http_status_t  status;
 
     static const nxt_str_t tmp_name_pattern = nxt_string("/req-XXXXXXXX");
 
     h1p = r->proto.h1;
+    skcf = r->conf->socket_conf;
 
     nxt_debug(task, "h1p request body read %O te:%d",
               r->content_length_n, h1p->transfer_encoding);
@@ -871,8 +878,19 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
     switch (h1p->transfer_encoding) {
 
     case NXT_HTTP_TE_CHUNKED:
-        status = NXT_HTTP_LENGTH_REQUIRED;
-        goto error;
+        if (!skcf->chunked_transform) {
+            status = NXT_HTTP_LENGTH_REQUIRED;
+            goto error;
+        }
+
+        if (r->content_length != NULL || !nxt_h1p_is_http11(h1p)) {
+            status = NXT_HTTP_BAD_REQUEST;
+            goto error;
+        }
+
+        r->chunked = 1;
+        h1p->chunked_parse.mem_pool = r->mem_pool;
+        break;
 
     case NXT_HTTP_TE_UNSUPPORTED:
         status = NXT_HTTP_NOT_IMPLEMENTED;
@@ -883,19 +901,20 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
         break;
     }
 
-    if (r->content_length_n == -1 || r->content_length_n == 0) {
+    if (!r->chunked &&
+        (r->content_length_n == -1 || r->content_length_n == 0))
+    {
         goto ready;
     }
 
     body_length = (size_t) r->content_length_n;
 
-    body_buffer_size = nxt_min(r->conf->socket_conf->body_buffer_size,
-                               body_length);
+    body_buffer_size = nxt_min(skcf->body_buffer_size, body_length);
 
     if (body_length > body_buffer_size) {
         nxt_str_t  *tmp_path, tmp_name;
 
-        tmp_path = &r->conf->socket_conf->body_temp_path;
+        tmp_path = &skcf->body_temp_path;
 
         tmp_name.length = tmp_path->length + tmp_name_pattern.length;
 
@@ -946,31 +965,69 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
 
     r->body = b;
 
-    body_rest = body_length;
+    body_rest = r->chunked ? 1 : body_length;
 
     in = h1p->conn->read;
 
     size = nxt_buf_mem_used_size(&in->mem);
 
     if (size != 0) {
-        size = nxt_min(size, body_length);
-
         if (nxt_buf_is_file(b)) {
-            res = nxt_fd_write(b->file->fd, in->mem.pos, size);
-            if (nxt_slow_path(res < (ssize_t) size)) {
-                status = NXT_HTTP_INTERNAL_SERVER_ERROR;
-                goto error;
-            }
+            if (r->chunked) {
+                out = nxt_http_chunk_parse(task, &h1p->chunked_parse, in);
 
-            b->file_end += size;
+                if (h1p->chunked_parse.error) {
+                    status = NXT_HTTP_INTERNAL_SERVER_ERROR;
+                    goto error;
+                }
+
+                if (h1p->chunked_parse.chunk_error) {
+                    status = NXT_HTTP_BAD_REQUEST;
+                    goto error;
+                }
+
+                for (chunk = out; chunk != NULL; chunk = chunk->next) {
+                    size = nxt_buf_mem_used_size(&chunk->mem);
+
+                    res = nxt_fd_write(b->file->fd, chunk->mem.pos, size);
+                    if (nxt_slow_path(res < (ssize_t) size)) {
+                        status = NXT_HTTP_INTERNAL_SERVER_ERROR;
+                        goto error;
+                    }
+
+                    b->file_end += size;
+
+                    if ((size_t) b->file_end > skcf->max_body_size) {
+                        status = NXT_HTTP_PAYLOAD_TOO_LARGE;
+                        goto error;
+                    }
+                }
+
+                if (h1p->chunked_parse.last) {
+                    body_rest = 0;
+                }
+
+            } else {
+                size = nxt_min(size, body_length);
+                res = nxt_fd_write(b->file->fd, in->mem.pos, size);
+                if (nxt_slow_path(res < (ssize_t) size)) {
+                    status = NXT_HTTP_INTERNAL_SERVER_ERROR;
+                    goto error;
+                }
+
+                b->file_end += size;
+
+                in->mem.pos += size;
+                body_rest -= size;
+            }
 
         } else {
             size = nxt_min(body_buffer_size, size);
             b->mem.free = nxt_cpymem(b->mem.free, in->mem.pos, size);
-        }
 
-        in->mem.pos += size;
-        body_rest -= size;
+            in->mem.pos += size;
+            body_rest -= size;
+        }
     }
 
     nxt_debug(task, "h1p body rest: %uz", body_rest);
@@ -1028,9 +1085,10 @@ nxt_h1p_conn_request_body_read(nxt_task_t *task, void *obj, void *data)
 {
     size_t              size, body_rest;
     ssize_t             res;
-    nxt_buf_t           *b;
+    nxt_buf_t           *b, *out, *chunk;
     nxt_conn_t          *c;
     nxt_h1proto_t       *h1p;
+    nxt_socket_conf_t   *skcf;
     nxt_http_request_t  *r;
     nxt_event_engine_t  *engine;
 
@@ -1040,38 +1098,77 @@ nxt_h1p_conn_request_body_read(nxt_task_t *task, void *obj, void *data)
     nxt_debug(task, "h1p conn request body read");
 
     r = h1p->request;
+    skcf = r->conf->socket_conf;
 
     engine = task->thread->engine;
 
     b = c->read;
 
     if (nxt_buf_is_file(b)) {
-        body_rest = b->file->size - b->file_end;
 
-        size = nxt_buf_mem_used_size(&b->mem);
-        size = nxt_min(size, body_rest);
+        if (r->chunked) {
+            body_rest = 1;
 
-        res = nxt_fd_write(b->file->fd, b->mem.pos, size);
-        if (nxt_slow_path(res < (ssize_t) size)) {
-            nxt_h1p_request_error(task, h1p, r);
-            return;
-        }
+            out = nxt_http_chunk_parse(task, &h1p->chunked_parse, b);
 
-        b->file_end += size;
-        body_rest -= res;
-
-        b->mem.pos += size;
-
-        if (b->mem.pos == b->mem.free) {
-            if (body_rest >= (size_t) nxt_buf_mem_size(&b->mem)) {
-                b->mem.free = b->mem.start;
-
-            } else {
-                /* This required to avoid reading next request. */
-                b->mem.free = b->mem.end - body_rest;
+            if (h1p->chunked_parse.error) {
+                nxt_h1p_request_error(task, h1p, r);
+                return;
             }
 
-            b->mem.pos = b->mem.free;
+            if (h1p->chunked_parse.chunk_error) {
+                nxt_http_request_error(task, r, NXT_HTTP_BAD_REQUEST);
+                return;
+            }
+
+            for (chunk = out; chunk != NULL; chunk = chunk->next) {
+                size = nxt_buf_mem_used_size(&chunk->mem);
+                res = nxt_fd_write(b->file->fd, chunk->mem.pos, size);
+                if (nxt_slow_path(res < (ssize_t) size)) {
+                    nxt_h1p_request_error(task, h1p, r);
+                    return;
+                }
+
+                b->file_end += size;
+
+                if ((size_t) b->file_end > skcf->max_body_size) {
+                    nxt_h1p_request_error(task, h1p, r);
+                    return;
+                }
+            }
+
+            if (h1p->chunked_parse.last) {
+                body_rest = 0;
+            }
+
+        } else {
+            body_rest = b->file->size - b->file_end;
+
+            size = nxt_buf_mem_used_size(&b->mem);
+            size = nxt_min(size, body_rest);
+
+            res = nxt_fd_write(b->file->fd, b->mem.pos, size);
+            if (nxt_slow_path(res < (ssize_t) size)) {
+                nxt_h1p_request_error(task, h1p, r);
+                return;
+            }
+
+            b->file_end += size;
+            body_rest -= res;
+
+            b->mem.pos += size;
+
+            if (b->mem.pos == b->mem.free) {
+                if (body_rest >= (size_t) nxt_buf_mem_size(&b->mem)) {
+                    b->mem.free = b->mem.start;
+
+                } else {
+                    /* This required to avoid reading next request. */
+                    b->mem.free = b->mem.end - body_rest;
+                }
+
+                b->mem.pos = b->mem.free;
+            }
         }
 
     } else {
--- a/src/nxt_h1proto.c
+++ b/src/nxt_h1proto.c
@@ -856,7 +856,6 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
 {
     size_t             size, body_length, body_buffer_size, body_rest;
     ssize_t            res;
-    nxt_str_t          *tmp_path, tmp_name;
     nxt_buf_t          *in, *b;
     nxt_conn_t         *c;
     nxt_h1proto_t      *h1p;
@@ -894,6 +893,8 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
                                body_length);
 
     if (body_length > body_buffer_size) {
+        nxt_str_t  *tmp_path, tmp_name;
+
         tmp_path = &r->conf->socket_conf->body_temp_path;
 
         tmp_name.length = tmp_path->length + tmp_name_pattern.length;
@@ -901,23 +902,11 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
         b = nxt_buf_file_alloc(r->mem_pool,
                                body_buffer_size + sizeof(nxt_file_t)
                                + tmp_name.length + 1, 0);
+        if (nxt_slow_path(b == NULL)) {
+            status = NXT_HTTP_INTERNAL_SERVER_ERROR;
+            goto error;
+        }
 
-    } else {
-        /* This initialization required for CentOS 6, gcc 4.4.7. */
-        tmp_path = NULL;
-        tmp_name.length = 0;
-
-        b = nxt_buf_mem_alloc(r->mem_pool, body_buffer_size, 0);
-    }
-
-    if (nxt_slow_path(b == NULL)) {
-        status = NXT_HTTP_INTERNAL_SERVER_ERROR;
-        goto error;
-    }
-
-    r->body = b;
-
-    if (body_length > body_buffer_size) {
         tmp_name.start = nxt_pointer_to(b->mem.start, sizeof(nxt_file_t));
 
         memcpy(tmp_name.start, tmp_path->start, tmp_path->length);
@@ -946,7 +935,16 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
                   &tmp_name, b->file->fd);
 
         unlink((char *) tmp_name.start);
+
+    } else {
+        b = nxt_buf_mem_alloc(r->mem_pool, body_buffer_size, 0);
+        if (nxt_slow_path(b == NULL)) {
+            status = NXT_HTTP_INTERNAL_SERVER_ERROR;
+            goto error;
+        }
     }
+
+    r->body = b;
 
     body_rest = body_length;
 
--- a/src/nxt_http_parse.h
+++ b/src/nxt_http_parse.h
@@ -96,11 +96,8 @@ struct nxt_http_field_s {
 
 
 typedef struct {
-    u_char                    *pos;
     nxt_mp_t                  *mem_pool;
-
     uint64_t                  chunk_size;
-
     uint8_t                   state;
     uint8_t                   last;         /* 1 bit */
     uint8_t                   chunk_error;  /* 1 bit */
--- a/src/nxt_http_request.c
+++ b/src/nxt_http_request.c
@@ -540,14 +540,57 @@ static const nxt_http_request_state_t  nxt_http_request_body_state
 };
 
 
+static nxt_int_t
+nxt_http_request_chunked_transform(nxt_http_request_t *r)
+{
+    size_t            size;
+    u_char            *p, *end;
+    nxt_http_field_t  *f;
+
+    r->chunked_field->skip = 1;
+
+    size = r->body->file_end;
+
+    f = nxt_list_zero_add(r->fields);
+    if (nxt_slow_path(f == NULL)) {
+        return NXT_ERROR;
+    }
+
+    nxt_http_field_name_set(f, "Content-Length");
+
+    p = nxt_mp_nget(r->mem_pool, NXT_OFF_T_LEN);
+    if (nxt_slow_path(p == NULL)) {
+        return NXT_ERROR;
+    }
+
+    f->value = p;
+    end = nxt_sprintf(p, p + NXT_OFF_T_LEN, "%uz", size);
+    f->value_length = end - p;
+
+    r->content_length = f;
+    r->content_length_n = size;
+
+    return NXT_OK;
+}
+
+
 static void
 nxt_http_request_ready(nxt_task_t *task, void *obj, void *data)
 {
+    nxt_int_t           ret;
     nxt_http_action_t   *action;
     nxt_http_request_t  *r;
 
     r = obj;
     action = r->conf->socket_conf->action;
+
+    if (r->chunked) {
+        ret = nxt_http_request_chunked_transform(r);
+        if (nxt_slow_path(ret != NXT_OK)) {
+            nxt_http_request_error(task, r, NXT_HTTP_INTERNAL_SERVER_ERROR);
+            return;
+        }
+    }
 
     nxt_http_request_action(task, r, action);
 }
--- a/src/nxt_conf_validation.c
+++ b/src/nxt_conf_validation.c
@@ -368,6 +368,9 @@ static nxt_conf_vldt_object_t  nxt_conf_vldt_http_members[] = {
     }, {
         .name       = nxt_string("server_version"),
         .type       = NXT_CONF_VLDT_BOOLEAN,
+    }, {
+        .name       = nxt_string("chunked_transform"),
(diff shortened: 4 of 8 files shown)


src/nxt_h1proto.c
```diff
@@ -884,6 +882,34 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
...
+        in = h1p->conn->read;
+        size = nxt_buf_mem_used_size(&in->mem);
+        if (size == 0) {
+            status = NXT_HTTP_NOT_IMPLEMENTED;
+            goto error;
+        }
+
+        nxt_buf_t *out = nxt_http_chunk_parse(task, &h1p->chunked_parse, in);
+
+        if (h1p->chunked_parse.chunk_error ||
```
avahahn, line 896: Is this conditional meant to be formatted like this?

src/nxt_h1proto.c
```diff
@@ -884,6 +882,34 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
         break;
     }
 
+    if (r->chunked) {
+        h1p->chunked_parse.mem_pool = r->mem_pool;
+        in = h1p->conn->read;
+        size = nxt_buf_mem_used_size(&in->mem);
+        if (size == 0) {
+            status = NXT_HTTP_NOT_IMPLEMENTED;
+            goto error;
```
avahahn, line 891: I notice in `nxt_h1p_peer_header_read_done` that the case where `nxt_buf_mem_used_size` returns non-zero goes to the following logic:
```c
        r->state->ready_handler(task, r, peer);
        return;
```

Does that mean we should goto ready instead?

avahahn (commented):
One question:
Does this result in the response being buffered in NGINX Unit, or are the chunks immediately streamed to the peer?

src/nxt_h1proto.c
ac000: No, it looks simply like a snafu...

src/nxt_h1proto.c
```diff
@@ -884,6 +882,34 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
...
+;
+            goto error;
+        }
+
+        if (h1p->chunked_parse.last) {
+            out->next = nxt_buf_sync_alloc(r->mem_pool, NXT_BUF_SYNC_LAST);
+            out = nxt_h1p_chunk_create(task, r, out);
+            r->body = out;
+            r->state->ready_handler(task, r,
+NULL);
```
ac000, line 908: ... same as here ...

But this is clearly just a PoC/work in progress...

hongzhidao: Yep, just for PoC.

avahahn:
fixes: #445

avahahn:
Here's a patch that fixes the formatting nits:
```diff
diff --git a/src/nxt_h1proto.c b/src/nxt_h1proto.c
index 6550efb9..16f6dcaa 100644
--- a/src/nxt_h1proto.c
+++ b/src/nxt_h1proto.c
@@ -893,10 +893,8 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
 
         nxt_buf_t *out = nxt_http_chunk_parse(task, &h1p->chunked_parse, in);
 
-        if (h1p->chunked_parse.chunk_error ||
- h1p->chunked_parse.error) {
-            status = NXT_HTTP_NOT_IMPLEMENTED
-;
+        if (h1p->chunked_parse.chunk_error || h1p->chunked_parse.error) {
+            status = NXT_HTTP_NOT_IMPLEMENTED;
             goto error;
         }
 
@@ -904,8 +902,7 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
             out->next = nxt_buf_sync_alloc(r->mem_pool, NXT_BUF_SYNC_LAST);
             out = nxt_h1p_chunk_create(task, r, out);
             r->body = out;
-            r->state->ready_handler(task, r,
-NULL);
+            r->state->ready_handler(task, r, NULL);
             return;
         }
     }
```
Can we get this PR converted to a non-draft PR? I think we should push this in as a solution for chunked transfer encoded requests.

hongzhidao:
> Can we get this PR converted to a non-draft PR?

Does non-draft PR mean it's ready to review? And it will run with GH actions for tests?

ac000:
> > Can we get this PR converted to a non-draft PR?
> 
> Does non-draft PR mean it's ready to review? And it will run with GH actions for tests?

I generally regard a non-draft PR as reviewable (whether I've been assigned to review or not), sometimes I'll do a rough review of a draft PR depending on circumstances...

Draft and non-draft PR's both get run through the GH actions, of which you can generally ignore the Ruby 3.2 failures....

hongzhidao:
Good to know, thanks.
Btw, the way to support the chunked body should be different from the patch. The final solution is not clear yet for me.

callahad:
> Btw, the way to support the chunked body should be different from the patch. The final solution is not clear yet for me.

Can you say more about how this approach doesn't feel quite right to you?

We need to land something ASAP, and should act as though the deadline to have a mergeable / testable branch is the end of this week. 

Is the current approach _good enough_ for that, or do you think this approach is so flawed that you would block merging?

ac000:
> We need to land something ASAP, and should act as though the deadline

Actually, I don't think we do, not in the main Unit repository anyway...

> to have a mergeable / testable branch is the end of this week.

There is nothing preventing you from having a branch in whoevers
personal clone for testing purposes or even for actual use until a
_correct_ solution in merged upstream...

hongzhidao:
> Can you say more about how this approach doesn't feel quite right to you?

Sure. Let me list some of the challenges.

 1. About the pipeline process.
    Unit supports pipeline requests. For any request, it doesn't handle the request after it. Since for content-length body, we know how much to read based on the content-length size from the request header. But for chunked body, the size isn't known until reading the data in the request body and parsed successfully for each chunk.
    So to solve the problem, we need to check if we need to change the design of limiting not to read or process the next request for the current request. I'd prefer to keep the design, but it looks like chunked body might break it.

2. About the body for size that is larger than the size of the buffer.
  In Unit, if the request body is larger than the buffer which is 8k size, the body will be dumped into a temporary file. For content-length request, after reading request header, we've known if the body will be saved into a memory buffer or a file buffer. But for chunked body, we have to check it while reading and parsing the body.
 To write a patch to be ready to review, I feel we need to change the implementation of content-length body reading and make it general. NGINX does it this way.

3. Ideally, we should send the chunked body to application process for the chunked request.
   Unfortunately, the application process only process content-length body. It's not a small change to support it. The imperfect way is to transfer the chunked body to content-length body before sending to application process, and it means we need to receive the entire body in this way.

4. The buffer type for the request body.
   For the size that is less than buffer size, Unit uses a memory buffer to save it. But if the size is larger than buffer size, Unit uses a file buffer instead. To note this point since this design has a lot to do with the challenge.

> Is the current approach good enough for that, or do you think this approach is so flawed that you would block merging?

The patch I posted last week is for the demo asked Thursday,  and it didn't solve the above challenges.

> We have an internal opportunity with a constraint that requires landing functional (but not perfect) support for Transfer-Encoding: chunked in master this calendar month.

I'll try my best to provide a patch that can work with application processes, but I'm afraid it's not ready to be reviewed or merged into upstream. It needs to take more development time.

hongzhidao:
> One small other nit, perhaps you could reword the subject to something like
> Support chunked request bodies

Good suggestion, let me know if `body` needs to be replaced with `bodies`.

> I see this is still marked as a draft

Yep, it's for the demo.
> We need to land something ASAP, and should act as though the deadline to have a mergeable / testable branch is the end of this week.

> Internally, I transferred the chunked request to content-length request since the application process can't handle chunked request now.

The design should be rewritten to make more sense. But it's not clear yet. Will do it after the demo.

src/nxt_h1proto.c
```diff
@@ -872,52 +873,39 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
...
         break;
     }
 
-    if (r->content_length_n == -1 || r->content_length_n == 0) {
+    if ((r->content_length_n <= 0) && !r->chunked) {
         goto ready;
     }
 
-    body_length = (size_t) r->content_length_n;
+    body_length = (ssize_t) r->content_length_n;
```
ac000, line 894: This cast shouldn't be needed.

body_length is a `ssize_t` and  r->content_length_n is a `nxt_off_t` aka `off_t` which is defined as signed by [POSIX](https://pubs.opengroup.org/onlinepubs/009696899/basedefs/sys/types.h.html)

So both types are `signed`, the only difference being on 32bit, ssize_t is 4 bytes whereas off_t is still 8 bytes due to _FILE_OFFSET_BITS=64

src/nxt_http_request.c
```diff
@@ -549,6 +549,34 @@ nxt_http_request_ready(nxt_task_t *task, void *obj, void *data)
...
+        nxt_http_field_t *f;
+        f = nxt_list_zero_add(r->fields);
+        if (nxt_slow_path(f == NULL)) {
+            /* error */
+            return;
+        }
+
+        nxt_http_field_name_set(f, "Content-Length");
+
+        u_char *p, *end;
```
ac000, line 565: Can we please declare these at the top of the `if ()` scope?, i.e

```c
if (r->chunked) {
    u_char  *p, *end;

...
```

(I'd love for us to be able to compile with `-Wdeclaration-after-statement`)

ac000:
> > One small other nit, perhaps you could reword the subject to something like
> > Support chunked request bodies
> 
> Good suggestion, let me know if `body` needs to be replaced with `bodies`.

Yes, I think the plural `bodes` makes more sense here...

> > I see this is still marked as a draft
> 
> Yep, it's for the demo.
> 
> > We need to land something ASAP, and should act as though the deadline to have a mergeable / testable branch is the end of this week.
> 
> > Internally, I transferred the chunked request to content-length request since the application process can't handle chunked request now.
> 
> The design should be rewritten to make more sense. But it's not clear yet. Will do it after the demo.

Sure...

src/nxt_http_request.c
```diff
@@ -549,6 +549,34 @@ nxt_http_request_ready(nxt_task_t *task, void *obj, void *data)
...
+        p = nxt_mp_nget(r->mem_pool, NXT_OFF_T_LEN);
+        if (nxt_slow_path(p == NULL)) {
+            /* error */
+            return;
+        }
+
+        f->value = p;
+        end = nxt_sprintf(p, p + NXT_OFF_T_LEN, "%O", size);
```
andrey-zelenkov, line 573: Buildbot found following error when trying to compile with clang-ast on Debian 11:

```
src/nxt_http_request.c:573:55: error: bad argument for %O arg: size_t, resolved as: unsigned long
        end = nxt_sprintf(p, p + NXT_OFF_T_LEN, "%O", size);
```

ac000: Forgot about that thing. We should look at moving that over to GH actions...

`size` is  a `size_t`

so

```
 *    %[0][width][u][x|X]z      ssize_t/size_t                                  
```

I guess it should be `%uz` instead...

src/nxt_http_request.c
hongzhidao: Yep, such declares should be at the top.
I guess the code will be rewrite with a good design, now it's for the running demo.

src/nxt_http_request.c
hongzhidao: Fixed, thanks.

hongzhidao: Fixed, thanks.

hongzhidao:
Hi,
> Ideally, we should send the chunked body to application process for the chunked request.
Unfortunately, the application process only process content-length body. It's not a small change to support it. The imperfect way is to transfer the chunked body to content-length body before sending to application process, and it means we need to receive the entire body in this way.

I'm trying to get the application process to parse chunked request body.

andrey-zelenkov (commented):
Hi @hongzhidao, some error handling improvements suggested. Full diff:

```
diff --git a/src/nxt_h1proto.c b/src/nxt_h1proto.c
--- a/src/nxt_h1proto.c
+++ b/src/nxt_h1proto.c
@@ -829,16 +829,27 @@ nxt_h1p_websocket_version(void *ctx, nxt
 static nxt_int_t
 nxt_h1p_transfer_encoding(void *ctx, nxt_http_field_t *field, uintptr_t data)
 {
+    nxt_h1proto_t       *h1p;
     nxt_http_te_t       te;
     nxt_http_request_t  *r;
 
     r = ctx;
+    h1p = r->proto.h1;
+
+    if (nxt_slow_path(r->content_length != NULL) || !nxt_h1p_is_http11(h1p)) {
+        return NXT_HTTP_BAD_REQUEST;
+    }
+
     field->skip = 1;
     field->hopbyhop = 1;
 
     if (field->value_length == 7
         && memcmp(field->value, "chunked", 7) == 0)
     {
+        if (nxt_slow_path(r->proto.h1->transfer_encoding == NXT_HTTP_TE_CHUNKED)) {
+            return NXT_HTTP_BAD_REQUEST;
+        }
+
         te = NXT_HTTP_TE_CHUNKED;
 
     } else {
@@ -856,12 +867,13 @@ nxt_h1p_transfer_encoding(void *ctx, nxt
 static void
 nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
 {
-    size_t         size, body_buffer_size, body_rest;
-    ssize_t        res, body_length;
-    nxt_str_t      *tmp_path, tmp_name;
-    nxt_buf_t      *in, *b, *out, *chunk;
-    nxt_conn_t     *c;
-    nxt_h1proto_t  *h1p;
+    size_t             size, body_buffer_size, body_rest;
+    ssize_t            res, body_length;
+    nxt_str_t          *tmp_path, tmp_name;
+    nxt_buf_t          *in, *b, *out, *chunk;
+    nxt_conn_t         *c;
+    nxt_h1proto_t      *h1p;
+    nxt_http_status_t  status;
 
     static const nxt_str_t tmp_name_pattern = nxt_string("/req-XXXXXXXX");
 
@@ -878,9 +890,8 @@ nxt_h1p_request_body_read(nxt_task_t *ta
         break;
 
     case NXT_HTTP_TE_UNSUPPORTED:
-        h1p->keepalive = 0;
-        nxt_http_request_error(task, r, NXT_HTTP_NOT_IMPLEMENTED);
-        return;
+        status = NXT_HTTP_NOT_IMPLEMENTED;
+        goto error;
 
     default:
     case NXT_HTTP_TE_NONE:
@@ -903,6 +914,7 @@ nxt_h1p_request_body_read(nxt_task_t *ta
                                body_buffer_size + sizeof(nxt_file_t)
                                + tmp_name.length + 1, 0);
         if (nxt_slow_path(b == NULL)) {
+            status = NXT_HTTP_INTERNAL_SERVER_ERROR;
             goto error;
         }
 
@@ -926,6 +938,7 @@ nxt_h1p_request_body_read(nxt_task_t *ta
         if (nxt_slow_path(b->file->fd == -1)) {
             nxt_alert(task, "mkstemp(%s) failed %E", tmp_name.start, nxt_errno);
 
+            status = NXT_HTTP_INTERNAL_SERVER_ERROR;
             goto error;
         }
 
@@ -941,6 +954,7 @@ nxt_h1p_request_body_read(nxt_task_t *ta
 
         b = nxt_buf_mem_alloc(r->mem_pool, body_buffer_size, 0);
         if (nxt_slow_path(b == NULL)) {
+            status = NXT_HTTP_INTERNAL_SERVER_ERROR;
             goto error;
         }
     }
@@ -965,6 +979,9 @@ nxt_h1p_request_body_read(nxt_task_t *ta
                 out = nxt_http_chunk_parse(task, &h1p->chunked_parse, in);
 
                 if (h1p->chunked_parse.chunk_error || h1p->chunked_parse.error) {
+                    nxt_debug(task, "h1p request body chunk parse error");
+
+                    status = NXT_HTTP_BAD_REQUEST;
                     goto error;
                 }
 
@@ -972,6 +989,7 @@ nxt_h1p_request_body_read(nxt_task_t *ta
                     size = nxt_buf_mem_used_size(&chunk->mem);
                     res = nxt_fd_write(b->file->fd, chunk->mem.pos, size);
                     if (nxt_slow_path(res < (ssize_t) size)) {
+                        status = NXT_HTTP_INTERNAL_SERVER_ERROR;
                         goto error;
                     }
 
@@ -985,6 +1003,7 @@ nxt_h1p_request_body_read(nxt_task_t *ta
             } else {
                 res = nxt_fd_write(b->file->fd, in->mem.pos, size);
                 if (nxt_slow_path(res < (ssize_t) size)) {
+                    status = NXT_HTTP_INTERNAL_SERVER_ERROR;
                     goto error;
                 }
 
@@ -1035,7 +1054,7 @@ error:
 
     h1p->keepalive = 0;
 
-    nxt_http_request_error(task, r, NXT_HTTP_INTERNAL_SERVER_ERROR);
+    nxt_http_request_error(task, r, status);
 }
 
 
diff --git a/src/nxt_http_request.c b/src/nxt_http_request.c
--- a/src/nxt_http_request.c
+++ b/src/nxt_http_request.c
@@ -220,6 +220,10 @@ nxt_http_request_content_length(void *ct
 
     r = ctx;
 
+    if (nxt_slow_path(r->transfer_encoding != NULL)) {
+        return NXT_HTTP_BAD_REQUEST;
+    }
+
     if (nxt_fast_path(r->content_length == NULL)) {
         r->content_length = field;
 
 
```

Feel free to edit/reject parts or whole diff, as you choose.

hongzhidao:
Hi all,
welcome to review, but the `max_body_size` check for the chunked body is not supported yet.
I'd suggest checking with @andrey-zelenkov's tests to understand how it is used.
https://github.com/nginx/unit/pull/1307

src/nxt_h1proto.c
```diff
@@ -847,20 +847,21 @@ nxt_h1p_transfer_encoding(void *ctx, nxt_http_field_t *field, uintptr_t data)
 
     r->proto.h1->transfer_encoding = te;
 
+    r->transfer_encoding = field;
+
```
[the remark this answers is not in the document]
hongzhidao: Good point.

src/nxt_h1proto.c
```diff
@@ -872,52 +873,39 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
...
-        status = NXT_HTTP_INTERNAL_SERVER_ERROR;
-        goto error;
-    }
-
-    r->body = b;
-
-    if (body_length > body_buffer_size) {
         tmp_name.start = nxt_pointer_to(b->mem.start, sizeof(nxt_file_t));
 
         memcpy(tmp_name.start, tmp_path->start, tmp_path->length);
```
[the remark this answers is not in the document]
hongzhidao: I didn't touch it, but I'm fine with using `nxt_memcpy` instead`.

src/nxt_h1proto.c
```diff
@@ -1007,7 +1035,7 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
 
     h1p->keepalive = 0;
 
-    nxt_http_request_error(task, r, status);
+    nxt_http_request_error(task, r, NXT_HTTP_INTERNAL_SERVER_ERROR);
```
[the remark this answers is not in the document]
hongzhidao: Makes sense.

src/nxt_h1proto.c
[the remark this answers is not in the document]
hongzhidao: Ok, I reworked the checking about transfer encoding stuff.

src/nxt_h1proto.c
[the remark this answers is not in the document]
ac000: I don't think it's worth changing it for changes sake... it's a straight wrapper for memcpy(3)

ac000:
Ignore the Fedora Rawhide failures, Rawhide seems broken at the moment...

src/nxt_http_parse.h
```diff
@@ -96,11 +96,8 @@ struct nxt_http_field_s {
 
 
 typedef struct {
-    u_char                    *pos;
     nxt_mp_t                  *mem_pool;
-
     uint64_t                  chunk_size;
-
```
ac000, line 103: I would normally suggest such a change (removing the blank lines) be done in a separate commit, but in this case I think it's OK...

ac000:
Hi @hongzhidao 

You can add my

```
```

to

```
http: Move chunked buffer pos pointer while parsing
```

src/nxt_h1proto.c
```diff
@@ -946,8 +934,20 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
                   &tmp_name, b->file->fd);
 
         unlink((char *) tmp_name.start);
+
+    } else {
+        tmp_path = NULL;
+        tmp_name.length = 0;
```
ac000, line 951: Hmm, there was comment that went along with this

```c
-        /* This initialization required for CentOS 6, gcc 4.4.7. */
-        tmp_path = NULL;
-        tmp_name.length = 0;
```

Is this still required?

The oldest version of GCC we _may_ currently support is 4.7.x for C11/GNU11

[4.6.x](https://gcc.gnu.org/onlinedocs/gcc-4.6.4/gcc/Standards.html#Standards) did have some preliminary support, but

> A fourth version of the C standard, known as C1X, is under development; GCC has limited preliminary support for parts of this standard, enabled with -std=c1x.

not via -std=c11/gnu11

[4.7.x](https://gcc.gnu.org/onlinedocs/gcc-4.7.4/gcc/Standards.html#Standards) was the first to introduce -std=c11/gnu11

> A fourth version of the C standard, known as C11, was published in 2011 as ISO/IEC 9899:2011. GCC has limited incomplete support for parts of this standard, enabled with -std=c11 or -std=iso9899:2011. (While in development, drafts of this standard version were referred to as C1X.)

Having said all that, I think the oldest version of GCC we actively use/used is 4.8.5 from CentOS 7 which is essentially EOL now...

hongzhidao:
> Hi @hongzhidao
> 
> You can add my
> 
> ```
> Reviewed-by: Andrew Clayton <a.clayton@nginx.com>
> ```
> 
> to
> 
> ```
> http: Move chunked buffer pos pointer while parsing
> ```

Does this mean the review on the commit has been done when the Reviewed-by tag is added?

src/nxt_h1proto.c
hongzhidao: > Is this still required?

I removed it. I feel it's needed in the past because of the logic like this:
```
if (a) {
    b = alloc1();
} else {
    /* This initialization required for CentOS 6, gcc 4.4.7. */
    b = alloc2()
}

if (b == NULL) {
    return NULL;
}

if (a) {
    use b;
} else {
    use b;
}
```

I merged the two condition branches, something like this:
```
if (a) {
    b = alloc1();
    if (b == NULL) {
        return NULL;
    }

    /* more logic will be added in the subsequent patch.
    use b;

} else {
    b = alloc2();
    if (b == NULL) {
        return NULL;
    }
}
```

But unfortunately, I can't reproduce the warning in the current code.
I feel it's impossible to cause such a warning with the new change.

ac000: For interest...

Current GCC, 14.x, at least doesn't complain if you remove those dummy initialisations, so it got smarter...

Also just for clarity, the problem was

```c
if (body_length > body_buffer_size) {
    tmp_path = ...
    tmp_name.length = ...
    b = ...
} else {
    b = ...
}

if (body_length > body_buffer_size) {
    tmp_name.start = ...
    ...
    b->mem.start += sizeof(nxt_file_t) + tmp_name.length + 1;
    ...
}
```

I guess older GCC at least couldn't match those two if () statements and so couldn't tell that those two variables _had_ to be already initialised by the time we enter the second if () statement.

After your refactoring this is a complete non-issue anyway as we no longer make any reference to those two variables after the initital if () statement.

In fact those two variable should be declared in the if () statement

```diff
diff --git ./src/nxt_h1proto.c ./src/nxt_h1proto.c
index bb8022d5..40e865af 100644
--- ./src/nxt_h1proto.c
+++ ./src/nxt_h1proto.c
@@ -861,7 +861,6 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
 {
     size_t             size, body_length, body_buffer_size, body_rest;
     ssize_t            res;
-    nxt_str_t          *tmp_path, tmp_name;
     nxt_buf_t          *in, *b, *out, *chunk;
     nxt_conn_t         *c;
     nxt_h1proto_t      *h1p;
@@ -903,6 +902,8 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
                                body_length);
 
     if (r->chunked || (body_length > body_buffer_size)) {
+        nxt_str_t  *tmp_path, tmp_name;
+
         tmp_path = &r->conf->socket_conf->body_temp_path;
 
         tmp_name.length = tmp_path->length + tmp_name_pattern.length;
@@ -945,9 +946,6 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
         unlink((char *) tmp_name.start);
 
     } else {
-        tmp_path = NULL;
-        tmp_name.length = 0;
-
         b = nxt_buf_mem_alloc(r->mem_pool, body_buffer_size, 0);
         if (nxt_slow_path(b == NULL)) {
             status = NXT_HTTP_INTERNAL_SERVER_ERROR;
```

ac000:
Why remove the link to the chunked issue?

ac000:
> > Hi @hongzhidao
> > You can add my
> > ```
> > Reviewed-by: Andrew Clayton <a.clayton@nginx.com>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to
> > ```
> > http: Move chunked buffer pos pointer while parsing
> > ```
> 
> Added, and I wonder if this means the review on the commit has been done when the Reviewed-by tag is asked to be added? If yes, I'll add the same tag on the other commits after they are reviewed.

It means I have reviewed that particular patch. I'll notify you the same for the others...

hongzhidao:
> Why remove the link to the chunked issue?\
> 
> **EDIT:** Hmm, I guess because this won't be enabled by default...

Sorry, which link do you mean?

src/nxt_h1proto.c
hongzhidao: Good to know, thanks.

> In fact those two variable should be declared in the if () statement

I'm fine with both the two ways, I'll keep it the same.

ac000: I don't think leaving it as is makes sense...

Previously those dummy initialisations were to stop old versions of GCC from complaining [-Wuninitialized].

There was also a comment explaining this.

Your code re-jigs things, keeping the (now totally unnecessary) dummy initialisations but removing the comment explaining why they are there.

So you can see we are now in an undesirable situation.

The options in increasing order of correctness...

1) Put the comment back.

However the comment will need updating as it no longer makes sense...

a) It talks about a compiler version that simply cannot be used to build Unit now. You would need to check if it compiles on CentOS 7 with gcc 4.8.5 with those dummy initialisations removed without complaining or not to know whether the comment is even required.

b) However (a) above is a completely pointless exercise as it won't complain now in any case because of the next point.

2) Simply remove those dummy initialisations. They serve absolutely _no_ purpose. Those two variables are _only_ used in that first if () branch.
 
3) Do (2) but also reduce the scope of those two variables to that if () branch as I showed in the above patch.

So really the options are (2) or (3) with (3) being the obviously _correct_ one.

hongzhidao: Makes sense. Updated with the way (3).

ac000: That's good, though I'm not sure why you did it in the last commit ("http: Support chunked request bodies")  and not this one ("http: Refactored nxt_h1p_request_body_read()")?

hongzhidao: Good catch, I did it in the wrong commit.
I have updated it, thanks.

ac000: Oh yeah, that patch looks much better now! :)

You can add my

```
```

to that as well.

src/nxt_h1proto.c
```diff
@@ -946,31 +964,72 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
...
+                    res = nxt_fd_write(b->file->fd, chunk->mem.pos, size);
+                    if (nxt_slow_path(res < (ssize_t) size)) {
+                        status = NXT_HTTP_INTERNAL_SERVER_ERROR;
+                        goto error;
+                    }
+
+                    b->file_end += size;
+
+                    if ((size_t) b->file_end
+                         > r->conf->socket_conf->max_body_size)
```
ac000, line 1000: I think the `>` should be aligned to be under the second `(` (i.e pull it back one character) as it's really a continuation of the opening `(`.

src/nxt_h1proto.c
```diff
@@ -946,31 +964,72 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
...
+
+                if (h1p->chunked_parse.last) {
+                    body_rest = 0;
+                }
+
+            } else {
+                size = nxt_min(size, body_length);
+                res = nxt_fd_write(b->file->fd, in->mem.pos, size);
+                if (nxt_slow_path(res < (ssize_t) size)) {
+                    status = NXT_HTTP_INTERNAL_SERVER_ERROR;
```
ac000, line 1015: I know this is what the previous code was doing, but this seems a little fragile, unless the write is happening with suspended signals? (same with the write above...)

If this is happening with signals blocked then the following can be ignored.

To quote the write(2) man-page

> If a write() is interrupted by a signal handler before  any  bytes  are
       written, then the call fails with the error EINTR; if it is interrupted
       after  at  least  one byte has been written, the call succeeds, and re‐
       turns the number of bytes written.

The usual thing is to do the write(2) in a loop and just continue the write again on EINTR (or until all bytes are written barring other errors)..

But this is not really in the scope of this patch and would likely be done as an enhancement to `nxt_fd_write()`

hongzhidao: Agree, "safe-write" can be an enhancement to nxt_fd_write().

src/nxt_h1proto.c
hongzhidao: I've never seen anything like this before, I'm fine with it.
But it reminds me to go the other way. Recheck it, please.

ac000: I simply meant this

```c
+                    if ((size_t) b->file_end
+                        > r->conf->socket_conf->max_body_size)
```

as if the cast wasn't there...

But you've changed it and it doesn't need splitting over two lines now anyway...

src/nxt_h1proto.c
```diff
@@ -946,31 +965,70 @@ nxt_h1p_request_body_read(nxt_task_t *task, nxt_http_request_t *r)
...
+
+                b->file_end += size;
+
+                in->mem.pos += size;
+                body_rest -= size;
+            }
 
         } else {
+            size = nxt_min(size, body_length);
             size = nxt_min(body_buffer_size, size);
```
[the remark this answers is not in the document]
hongzhidao: I feel this can be like this:
```
size = nxt_min(size, body_length);
size = nxt_min(body_buffer_size, size);
```
=>
```
size = nxt_min(body_buffer_size, size);
```
Since `body_buffer_size` can't be greater than `body_length`.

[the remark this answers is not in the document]
ac000: Even better!

ac000:
OK @hongzhidao good stuff!

I think you can add my

```
```

to that last commit now...

ac000 (approved):

PR #1268, opened 2024-05-17
tools/unitcli: Filter docker images by command

### LINKED ISSUE #1254: CLI: Filter Docker containers in a more clever way
See this PR comment: https://github.com/nginx/unit/pull/1241/files#r1594347912

As discussed with @avahahn this can likely be solved by filtering on Docker image labels, one of these, e.g. 
https://github.com/nginx/unit/blob/5d1ce5c4475f8d126b61650f3c678e3f94549f6b/pkg/docker/Dockerfile.minimal#L3-L9

javorszky:
Closes #1254 

Tested by launching a number of containers with the following expectations:

* unit in the name or organisation, but not running `unitd` as a command ✅ correctly filtered out
* images based on `unit:php8.3`, but entrypoints and commands customized to check for:
  * nothing before the `unitd` command (empty entrypoint) ✅ tested locally, works for me
  * nothing after the `unitd` command (no arguments) ❓ containers exit immediately, can't verify
  * nothing before or after the `unitd` command ❓ same issue as above, they exit immediately
* container with image name that has no mention of `unit` (`gabor:bla`), but running `unitd` as a command ✅ correctly shows up in instances


tools/unitctl/unitctl/src/cmd/instances.rs
```diff
@@ -51,7 +51,7 @@ pub(crate) async fn cmd(args: InstanceArgs) -> Result<(), UnitctlError> {
                         if !sock_path.is_dir() {
                           return Err(UnitctlError::ControlSocketError{
                                 kind: ControlSocketErrorKind::General,
-                                message: format!("user must specify a directory of UNIX socket directory"),
+                                message: "user must specify a directory of UNIX socket directory".to_string(),
```
javorszky, line 54: this is something RustRover picked up. I don't feel strongly about it, being new to the language, but the suggestion made sense ("don't use format without placeholders and arguments")

ac000:
Please prefix the commit subjects with `tools/unitctl: `.

javorszky:
> Please prefix the commit subjects with `tools/unitctl: `.

done

avahahn (commented):
This is great! I had previously imagined using container labels, but pattern matching on the command will support more types of containers anyways, so I think this is a great solution!

avahahn (approved):

PR #1270, opened 2024-05-17
Tests: initial "wasm-wasi-component" test

andrey-zelenkov:
Note: `libwasm_wasi_component.so` module should be place into `build/lib/unit/modules` and be discoverable by Unit.


ac000:
On Fri, 17 May 2024 09:45:56 -0700
andrey-zelenkov ***@***.***> wrote:

> -- File Changes --
> 
>     A test/test_wasm_component.py (18)
>     A test/unit/applications/lang/wasm_component.py (61)
>     M test/unit/check/discover_available.py (4)
>     A test/wasm_component/hello_world/.vscode/settings.json (10)

Committed by accident?

andrey-zelenkov:
> Committed by accident?

Yes, the patch needs to be polished, especially the app part.

ac000:
> Note: `libwasm_wasi_component.so` module should be place into `build/lib/unit/modules` and be discoverable by Unit.

Ah, I see what you mean... should be fixed by https://github.com/nginx/unit/pull/1348

avahahn:
If you would like to avoid checking in the generated bindings you can always omit them and use [wasm-tools](https://github.com/bytecodealliance/wasm-tools) to adapt a generic wasm app into a proxy component!

Here is an example:
https://github.com/avahahn/wasm-traffic-tool

ac000:
> If you would like to avoid checking in the generated bindings you can always omit them and use [wasm-tools](https://github.com/bytecodealliance/wasm-tools) to adapt a generic wasm app into a proxy component!
> 
> Here is an example: https://github.com/avahahn/wasm-traffic-tool

Hmm, it's going to be whichever is the lesser of two evils... (there may be a third evil also)

1) Include the generated bindings

2) Add a dependency on wasm-tools _and_ the proxy adaptor module.

Though the tests already add `cargo` and `cargo-component` as dependencies...

3) Generate the bindings on the fly, though no doubt adding another dependency....

ac000:
How often would these bindings need to be updated?

avahahn (approved):
This looks fantastic as is! 
@andrey-zelenkov was there more you wanted this to do? 
@ac000 Were we waiting on information about alternative test component approaches?

I am curious if I can just merge this in as is...

ac000:
I guess it's OK to merge now. We can deal with the bindings later if they become an issue...

Issue #1275, opened 2024-05-20
OSS-Fuzz: Moving the build upstream


pkillarjun:
Hi, 
summon @andrey-zelenkov  

I'm the same guy who did the integration of `unit` into `oss-fuzz`;
Currently, I'm planning to shift the fuzzing harness files from the OSS-Fuzz repo to the unit repo;

### oss-fuzz plan:
 * Move fuzzer files with `build.sh` if possible [like this](https://github.com/ethereum/go-ethereum/blob/master/oss-fuzz.sh);
 * Delete blot from the oss-fuzz repo;
 * Add `CIFuzz` [workflow](https://google.github.io/oss-fuzz/getting-started/continuous-integration);
 * Increase the coverage with a new fuzzer or by using the API properly.

### Additional information

`unit-fuzzing-pr.patch` is my translation in the build system for integrating fuzz builds.

Note that I'm a little skeptical of [make](https://github.com/nginx/unit/blob/master/auto/make) file patch. I need a little extra oversight from you on that.

#### unit-fuzzing-pr.patch

```patch
diff --git a/auto/help b/auto/help
index 8f7553f..2e866ac 100644
--- a/auto/help
+++ b/auto/help
@@ -54,6 +54,7 @@ cat << END
 
   --debug              enable debug logging
 
+  --fuzz=ENGINE        enable fuzz testing
 
   python OPTIONS       configure Python module
                        run "./configure python --help" to see available options
diff --git a/auto/make b/auto/make
index 2788b9f..13829a5 100644
--- a/auto/make
+++ b/auto/make
@@ -158,7 +158,7 @@ END
 
 # Object files.
 
-for nxt_src in $NXT_LIB_SRCS $NXT_TEST_SRCS $NXT_LIB_UNIT_SRCS \
+for nxt_src in $NXT_LIB_SRCS $NXT_TEST_SRCS $NXT_FUZZ_SRCS $NXT_LIB_UNIT_SRCS \
                src/test/nxt_unit_app_test.c \
                src/test/nxt_unit_websocket_chat.c \
                src/test/nxt_unit_websocket_echo.c
@@ -324,6 +324,61 @@ END
 fi
 
 
+if [ $NXT_FUZZ != NO ]; then
+
+    # Fuzz object files list.
+
+       $echo "NXT_FUZZ_OBJS = \\" >> $NXT_MAKEFILE
+
+       for nxt_src in $NXT_FUZZ_SRCS
+       do
[...]

PR #1291, opened 2024-05-25
fuzzing: Add fuzzing support and build script

pkillarjun:
Issue: https://github.com/nginx/unit/issues/1275

Fix: Null-dereference in [nxt_http_parse_fuzz](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52469)

Question: Where is the [formatter](https://nginx.org/en/docs/dev/development_guide.html#code_style)?

### FAQ

* `*_seed_corpus` https://llvm.org/docs/LibFuzzer.html#corpus
* `.dict` https://llvm.org/docs/LibFuzzer.html#dictionaries
* `build-fuzz.sh` To build a fuzz test locally.
* `oss-fuzz.sh`  For oss-fuzz infra, this is inspired from [ethereum](https://github.com/ethereum/go-ethereum/blob/master/oss-fuzz.sh).


pkillarjun:
You can review everything except `src/fuzz` for now. I still have some work to do in `src/fuzz`.

src/fuzz/nxt_basic_fuzz.c
```diff
@@ -0,0 +1,78 @@
+
```
ac000, line 1: No need for a leading blank line.

src/fuzz/nxt_basic_fuzz.c
```diff
@@ -0,0 +1,78 @@
...
+/*
+ * Copyright (C) NGINX, Inc.
+ */
+
+#include <nxt_main.h>
+
+
+#define kMinInputLength 2
+#define kMaxInputLength 128
```
ac000, line 10: #defines should really be all uppercase...

src/fuzz/nxt_basic_fuzz.c
```diff
@@ -0,0 +1,78 @@
...
+
+    return 0;
+}
+
+
+void
+nxt_base64_fuzz(u_char *data, size_t size)
+{
+    u_char  buf[256];
+    nxt_base64_decode(buf, data, size);
```
ac000, line 53: Blank line between the variable declarations and code please.

src/fuzz/nxt_basic_fuzz.c
```diff
@@ -0,0 +1,78 @@
...
+    nxt_time_parse(data, size);
+}
+
+
+void
+nxt_utf8_fuzz(const u_char *data, size_t size)
+{
+    const u_char  *in;
+    in = data;
+    nxt_utf8_decode(&in, data + size);
```
ac000, line 77: Same here...

src/fuzz/nxt_http_parse_fuzz.c
```diff
@@ -0,0 +1,106 @@
...
+
+#include <nxt_main.h>
+
+/* DO NOT TRY THIS AT HOME! */
+#include "nxt_controller.c"
+#include "nxt_h1proto.c"
+
+
+#define kMinInputLength 2
+#define kMaxInputLength 1024
```
ac000, line 14: Uppercase #defines...

src/fuzz/nxt_http_parse_fuzz.c
```diff
@@ -0,0 +1,106 @@
...
+
+
+int
+LLVMFuzzerInitialize(int *argc, char ***argv)
+{
+    nxt_lib_start("fuzzing", NULL, &environ);
+
+    if (nxt_slow_path(nxt_http_fields_hash(&nxt_controller_fields_hash,
+                               nxt_controller_request_fields,
+                               nxt_nitems(nxt_controller_request_fields)) != NXT_OK)) {
```
ac000, line 31: Hmm, that's a little unwieldy.

I think it would be better to do the call to `nxt_http_fields_hash()` separately first then check it's retunr value.

src/fuzz/nxt_http_parse_fuzz.c
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    if (nxt_slow_path(nxt_http_fields_hash(&nxt_h1p_fields_hash,
+                               nxt_h1p_fields, nxt_nitems(nxt_h1p_fields)) != NXT_OK)) {
+        return NXT_ERROR;
+    }
+
+    if (nxt_slow_path(nxt_http_fields_hash(&nxt_h1p_peer_fields_hash,
+                                nxt_h1p_peer_fields,
+                                nxt_nitems(nxt_h1p_peer_fields)) != NXT_OK)) {
```
ac000, line 42: Probably the same for these...

ac000:
> Question: Where is the [formatter](https://nginx.org/en/docs/dev/development_guide.html#code_style)?

There isn't one! (and I'm not sure it'd be able to handle Units idiosyncrasies)...

ac000:
I wonder if `src/fuzz` should really be something like 'fuzzers/oss-fuzz` ?

pkillarjun:
> I wonder if `src/fuzz` should really be something like 'fuzzers/oss-fuzz` ?

Not really. The `src/fuzz` fuzz tests are heavily inspired by the `src/test` unit tests, so let them be together.



Also, I will squash all these commits after this review.

ac000:
> > I wonder if `src/fuzz` should really be something like 'fuzzers/oss-fuzz` ?
> 
> Not really. The `src/fuzz` fuzz tests are heavily inspired by the `src/test` unit tests, so let them be together.

Most of the tests, what gets run by CI is under `./test`

There also seems to be a lot of binary files in there, which I don't think should be under `./src/`...

I guess we _may_ also want to support _other_ fuzzers; _trinity_, _american fuzzy lop_ etc..., so this whole thing should probably be named after what it is...

pkillarjun:
>Most of the tests, what gets run by CI is under ./test
>
> There also seems to be a lot of binary files in there, which I don't think should be under ./src/...

Ah, alright, I will change it to the `fuzzer` dir.

> I guess we may also want to support other fuzzers; trinity, american fuzzy lop etc..., so this whole thing should probably be named after what it is...

`LLVMFuzzerTestOneInput` can be compiled with `afl++` using [libAFLDriver.a](https://github.com/AFLplusplus/AFLplusplus/blob/stable/utils/aflpp_driver/README.md).

If you want to test it locally with afl++
```
cd $HOME

git clone https://github.com/AFLplusplus/AFLplusplus.git

pushd AFLplusplus/
make
popd

export CC="$HOME/Desktop/AFLplusplus/afl-clang-fast"
export CXX="$HOME/Desktop/AFLplusplus/afl-clang-fast++"
export CFLAGS="-g -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address"
export CXXFLAGS="-g -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address"
export LIB_FUZZING_ENGINE="$HOME/Desktop/AFLplusplus/libAFLDriver.a"

git clone -b fuzzing-update https://github.com/pkillarjun/unit.git

pushd unit/
./configure --no-regex --no-pcre2 --fuzz=$LIB_FUZZING_ENGINE
make fuzz -j$(nproc)
popd
```
And the same goes for `honggfuzz`.

Also `Trinity` and `syzkaller` are kernel fuzzers.

src/fuzz/build-fuzz.sh
```diff
@@ -0,0 +1,18 @@
...
+
+./configure --no-regex --no-pcre2 --fuzz=$LIB_FUZZING_ENGINE
+make fuzz -j$(nproc)
+
+mkdir build/fuzz_basic_seed
+mkdir build/fuzz_http_parse_seed
+mkdir build/fuzz_json_seed
+
+echo ""
+echo "Run: ./build/\${fuzzer} build/\${fuzzer}_parse_seed src/fuzz/\${fuzzer}_seed_corpus"
```
ac000, line 17: If I want to run the fuzz_basic test, then I do

```
$ ./build/fuzz_basic build/fuzz_basic_seed/ src/fuzz/fuzz_basic_seed_corpus/
```

So it seems the second argument help text should just be

```
build/\${fuzzer}_seed
```

ac000:
Is there a description somewhere of what the contents of the corpus files are? E.g what exactly does `./src/fuzz/fuzz_basic_seed_corpus/base64_0.bin` contain?

Also how do you analyse the generated `crash` files?

src/fuzz/build-fuzz.sh
pkillarjun: Thanks for pointing it out.

pkillarjun:
> Is there a description somewhere of what the contents of the corpus files are? E.g what exactly does ./src/fuzz/fuzz_basic_seed_corpus/base64_0.bin contain?

The input used in the [nxt_base64_test.c](https://github.com/nginx/unit/blob/ea5c41b8056997ed40138020272b5159271f1b87/src/test/nxt_base64_test.c#L20) test files for `nxt_base64_decode`.

`nxt_conf_json_parse_str` in [nxt_clone_test.c](https://github.com/nginx/unit/blob/ea5c41b8056997ed40138020272b5159271f1b87/src/test/nxt_clone_test.c#L45)

Also, I hijacked the `nxt_http_parse_request` in [nxt_http_parse_test.c](https://github.com/nginx/unit/blob/ea5c41b8056997ed40138020272b5159271f1b87/src/test/nxt_http_parse_test.c) so it can give me seed files.

pkillarjun:
> Also how do you analyse the generated crash files?

./build/fuzz_${fuzzer} ./crash-${hash}

ASAN gives a really nice report.

ac000:
A couple of general questions...

1) This has pretty limited coverage of Unit, are there plans to increase this?
2) ... which sort of leads onto this next question. Who is intended to maintain this going forward?

pkillarjun:
> This has pretty limited coverage of Unit, are there plans to increase this?

Yes, currently I'm using `src/test` for my `fuzzing` tests. After that, I will try to find new tests in the source code.

> ... which sort of leads onto this next question. Who is intended to maintain this going forward?

I can only guarantee, build fixes for API changes or linker errors, and bugs in my harness code that is in the `fuzzing` directory.
I won't be touching or fixing bugs in your codebase, The reason for that is pretty clear and you know it.

Also, I will add you in `auto_ccs` in https://github.com/google/oss-fuzz/pull/12002 if you don't mind? I believe your email for this GitHub account is [andrew@digital-domain.net](mailto:andrew@digital-domain.net).

pkillarjun:
Self Note: Before merging, sift all the http_fuzz into different files for better hit coverage.

ac000:
> > This has pretty limited coverage of Unit, are there plans to increase this?
> 
> Yes, currently I'm using `src/test` for my `fuzzing` tests. After that, I will try to find new tests in the source code.

OK...

> > ... which sort of leads onto this next question. Who is intended to maintain this going forward?
> 
> I can only guarantee, build fixes for API changes or linker errors, and bugs in my harness code that is in the `fuzzing` directory. I won't be touching or fixing bugs in your codebase, The reason for that is pretty clear and you know it.

OK, that's what I figured...
 
> Also, I will add you in `auto_ccs` in [google/oss-fuzz#12002](https://github.com/google/oss-fuzz/pull/12002) if you don't mind? I believe your email for this GitHub account is [andrew@digital-domain.net](mailto:andrew@digital-domain.net).

Correct.

auto/make
```diff
@@ -324,6 +324,59 @@ END
...
+		\$(CFLAGS) $NXT_BUILD_DIR/fuzzing/nxt_json_fuzz.o \\
+		$NXT_BUILD_DIR/lib/$NXT_LIB_STATIC \\
+		$NXT_LD_OPT $NXT_LIBM $NXT_LIBS $NXT_LIB_AUX_LIBS \\
+		$NXT_FUZZ
+
+END
+
+fi
+
+
```
ac000, line 379: I think this should be in `auto/fuzzing`

configure
```diff
@@ -179,4 +180,5 @@ if [ $NXT_NJS != NO ]; then
 fi
 
 . auto/make
+. auto/fuzzing
```
pkillarjun, line 183: Not adding an if statement here for a good reason.

fuzzing/oss-fuzz.sh
```diff
@@ -0,0 +1,26 @@
...
+
+zip -r $OUT/fuzz_basic_seed_corpus.zip fuzz_basic_seed_corpus/
+zip -r $OUT/fuzz_http_controller_seed_corpus.zip  fuzz_http_controller_seed_corpus/
+zip -r $OUT/fuzz_http_h1p_seed_corpus.zip  fuzz_http_h1p_seed_corpus/ 
+zip -r $OUT/fuzz_http_h1p_peer_seed_corpus.zip  fuzz_http_h1p_peer_seed_corpus/ 
+zip -r $OUT/fuzz_json_seed_corpus.zip fuzz_json_seed_corpus/
+
+# Delete temporary directories.
+rm -r fuzz_http_controller_seed_corpus/ fuzz_http_h1p_seed_corpus/ fuzz_http_h1p_peer_seed_corpus/
+popd
```
pkillarjun, line 26: I don't like for loops much; they make things look ugly. Also, using a `temp` directory is necessary.

fuzzing/build-fuzz.sh
```diff
@@ -0,0 +1,20 @@
...
+export CXXFLAGS="-g -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address,undefined -fsanitize=fuzzer-no-link"
+export LIB_FUZZING_ENGINE="-fsanitize=fuzzer"
+
+./configure --no-regex --no-pcre2 --fuzz=$LIB_FUZZING_ENGINE
+make fuzz -j$(nproc)
+
+mkdir -p build/fuzz_basic_seed
+mkdir -p build/fuzz_http_controller_seed
+mkdir -p build/fuzz_http_h1p_seed
+mkdir -p build/fuzz_http_h1p_peer_seed
```
pkillarjun, line 15: Every HTTP parser deserves its own fuzzer separately, for better hit coverage.

fuzzing/nxt_basic_fuzz.c
```diff
@@ -0,0 +1,92 @@
...
+    u_char  *input;
+    u_char  buf[256];
+
+    /*
+     * Adding unnecessary NUL bytes to resolve buffer-overflow read and
+     * use-of-uninitialized-value in nxt_base64_decode nxt_string.c:833:21.
+     */
+
+    input = nxt_zalloc(size + 1);
+    nxt_memcpy(input, data, size);
```
pkillarjun, line 63: There is no other way around. I tried to fix it by myself, but I am unable to pass the tests.

ac000: I'd like to understand what exactly the problem is here...

pkillarjun: Alright, 

There is a bug in the `nxt_base64_decode` function at [nxt_string.c#L833](https://github.com/nginx/unit/blob/master/src/nxt_string.c#L833).

I already reported this to `"F5-SIRT"`. They said it's not a security vulnerability. Also, I'm aware that it's not a security bug. Usually, one or two bytes of over-read can't do much.

#### Reproduce:
Download [crash-fuzz_basic.txt](https://github.com/user-attachments/files/15612584/crash-fuzz_basic.txt).

```bash
git clone --depth=1 -b fuzzing-update https://github.com/pkillarjun/unit.git
cd unit
bash fuzzing/build-fuzz.sh
./build/fuzz_basic crash-fuzz_basic.txt # Can't upload .raw.
```

ASAN
```txt
=================================================================
==268367==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x502000000035 at pc 0x00000059f198 bp 0x7fffbfd3eb00 sp 0x7fffbfd3eaf8
READ of size 1 at 0x502000000035 thread T0
    #0 0x59f197 in nxt_base64_decode /home/oblivion/Desktop/unit/src/nxt_string.c:832:24
    #1 0x5454ed in nxt_base64_fuzz /home/oblivion/Desktop/unit/fuzzing/nxt_basic_fuzz.c:65:5
    #2 0x5454ed in LLVMFuzzerTestOneInput /home/oblivion/Desktop/unit/fuzzing/nxt_basic_fuzz.c:42:5
    #3 0x451a44 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (/home/oblivion/Desktop/unit/build/fuzz_basic+0x451a44) (BuildId: 34d7d355b8590bdd92fee196ddf89ef7e3735913)
    #4 0x43a89a in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) (/home/oblivion/Desktop/unit/build/fuzz_basic+0x43a89a) (BuildId: 34d7d355b8590bdd92fee196ddf89ef7e3735913)
    #5 0x4401e0 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) (/home/oblivion/Desktop/unit/build/fuzz_basic+0x4401e0) (BuildId: 34d7d355b8590bdd92fee196ddf89ef7e3735913)
    #6 0x46bf56 in main (/home/oblivion/Desktop/unit/build/fuzz_basic+0x46bf56) (BuildId: 34d7d355b8590bdd92fee196ddf89ef7e3735913)
    #7 0x7faaf8810087 in __libc_start_call_main (/lib64/libc.so.6+0x2a087) (BuildId: 4a92fcedbba6d6d2629ce066a2970017faa9995e)
    #8 0x7faaf881014a in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x2a14a) (BuildId: 4a92fcedbba6d6d2629ce066a2970017faa9995e)
    #9 0x434eb4 in _start (/home/oblivion/Desktop/unit/build/fuzz_basic+0x434eb4) (BuildId: 34d7d355b8590bdd92fee196ddf89ef7e3735913)
```

### Thought

After some debugging, I found out that this happened due to `src[1]` because there was not enough buffer available. So now the `NUL byte` becomes the wanted buffer, and `nxt_base64_decode` does not crash.

If i try to fix this over-read using `if (pad > 0 && end - src > 1)`, `nxt_base64_test` test dose not pass.

ac000: > Alright,
> 
> There is a bug in the `nxt_base64_decode` function at [nxt_string.c#L833](https://github.com/nginx/unit/blob/master/src/nxt_string.c#L833).

Line 832 no? (according to the asan report)

> I already reported this to `"F5-SIRT"`. They said it's not a security vulnerability. Also, I'm aware that it's not a security bug. Usually, one or two bytes of over-read can't do much.

Yes, I saw that, but couldn't see an obvious issue with the code. I also ran our `./src/test/nxt_base64_test.c` under valgrind and libasan and didn't see any issue.

When trying to debug it via `./fuzzing/nxt_basic_fuzz.c` from what I could tell it was failing when passing in an empty buffer but with a size of 2.

Is there any way to actually see what the inputs are when it fails? i.e what is it that's being passed into `nxt_base64_decode()` for `src`, `dst` and `size`?

(Other than putting in a printf(), which is where I'm seeing the empty buffer...)

> #### Reproduce:
> 
> Download [crash-fuzz_basic.txt](https://github.com/user-attachments/files/15612584/crash-fuzz_basic.txt).
> 
> ```shell
> git clone --depth=1 -b fuzzing-update https://github.com/pkillarjun/unit.git
> cd unit
> bash fuzzing/build-fuzz.sh
> ./build/fuzz_basic crash-fuzz_basic.txt # Can't upload .raw.
> ```

```
$ hexdump -C ~/Downloads/crash-fuzz_basic.txt
00000000  2c e5 00 00 00                                    |,....|
00000005
```

Which seems to be a comma followed by the non-printable extended ASCII character 229, then three nuls.

> ASAN
> 
> ```
> =================================================================
> ==268367==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x502000000035 at pc 0x00000059f198 bp 0x7fffbfd3eb00 sp 0x7fffbfd3eaf8
> READ of size 1 at 0x502000000035 thread T0
>     #0 0x59f197 in nxt_base64_decode /home/oblivion/Desktop/unit/src/nxt_string.c:832:24
> ```

```c
        v1 = decode[src[0]];                                                   
```

> 
> ### Thought
> 
> After some debugging, I found out that this happened due to `src[1]` because there was not enough buffer available. So now the `NUL byte` becomes the wanted buffer, and `nxt_base64_decode` does not crash.
> 
> If i try to fix this over-read using `if (pad > 0 && end - src > 1)`, `nxt_base64_test` test dose not pass.
> 
> For Example:
> 
>     * `openssl base64 -d <<< ""z9+Npe=="" | xxd -p`
>       Output : `cfdf8da5`.
> 
>     * What is should be from `tests[i].dec.start`
>       Output : `cfdf 8da5`
> 
>     * After Patch.
>       Output : `cfdf 8d`

So if I use `./src/test/nxt_base64_test.c` with a `src` of `,\E5\00\00\00`, what `size/length` was it using, 5? and run that under valgrind I should see the issue?

ac000: OK, a buffer with 13 chars fails, I suspect 17, 21, 25 ... will also fail... increments of 4 seems an issue... there is some mod 4 stuff going on in there...

ac000: With a sbuf of a given size and the second src@ & end@ are just before we do decode[src[0]] & decode[src[1]]

```
sbuf[17]
src@0x402220, end@0x402231
pad [3] length [17]
src@0x402230, end@0x402230

sbuf[18]
src@0x402220, end@0x402232
pad [2] length [18]
src@0x402230, end@0x402230

sbuf[19]
src@0x402220, end@0x402233
pad [1] length [19]
src@0x402230, end@0x402230

sbuf[20]
src@0x402220, end@0x402234
pad [0] length [20]
src@0x402234, end@0x402234
```

When pad ==3 (which it will be for any length >= 4 and an increment of 4) the src address is only one away from the end address.

Seeing if we need to special case the pad == 3 case inside the if (pad > 0)...

ac000: Here's a question.

Is it even possible for valid base64 to have 3 padding characters?

ac000: I did have a patch handling the pad == 3 case, but if I'd looked harder further up the function, I would have noticed that we indeed do treat a padding of 3 as invalid...

```c
    if (dst == NULL) {                                                          
        if (pad > 2) {                                                          
            return NXT_ERROR;                                                   
        } 
```
So, yeah as I suspected a padding of 3 would mean invalid base64.

`nxt_base64_decode()` is only used by the OpenSSL code. The configuration validation code calls it once with dst as `NULL` for checking the base64 stored in the config.

There is also a single call to `nxt_base64_decode(`) in `nxt_tls_ticket_keys()` (which depends on NXT_HAVE_OPENSSL_TLSEXT & SSL_CTRL_SET_TLSEXT_TICKET_KEY_CB) where it retrieves some base64 from the config.

But that looks to be checked by the afore mentioned configuration check in `nxt_conf_vldt_ticket_key_element()`

```c
    ret = nxt_base64_decode(NULL, key.start, key.length);                       
    if (ret == NXT_ERROR) {                                                     
        return nxt_conf_vldt_error(vldt, "Invalid Base64 format for the ticket "
                                   "key \"%V\".", &key);                        
    }
```

So there is no issue for us.

The correct way to use this API is to call `nxt_base64_decode()` with `dst = NULL` first to check the validity of the base64.

pkillarjun: > The correct way to use this API is to call nxt_base64_decode() with dst = NULL first to check the validity of the base64.

Ya, thanks for the help.

fuzzing/nxt_basic_fuzz.c
```diff
@@ -51,19 +51,19 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
...
     /*
-     * Adding unnecessary NUL bytes to resolve buffer-overflow read and
-     * use-of-uninitialized-value in nxt_base64_decode nxt_string.c:833:21.
+     * Do a validation check on base64 data before decoding.
      */
 
-    input = nxt_zalloc(size + 1);
-    nxt_memcpy(input, data, size);
+    ret = nxt_base64_decode(NULL, (u_char *)data, size);
+    if(ret == NXT_ERROR) {
```
ac000, line 61: Space between the `if` and opening `(` please.

fuzzing/nxt_basic_fuzz.c
```diff
@@ -51,19 +51,19 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 void
 nxt_base64_fuzz(const u_char *data, size_t size)
 {
-    u_char  *input;
-    u_char  buf[256];
+    ssize_t  ret;
+    u_char   buf[256];
```
ac000, line 55: In Unit we generally order out variables in 'Christmas tree' order of the variable type (then alphabetically).

So these two should be switched around...

pkillarjun: > Large arrays are put at the end of a declaration block. [code_style_variables](https://nginx.org/en/docs/dev/development_guide.html#code_style_variables)

Ohh, I thought arrays in general are put at the end of a declaration block.

fuzzing/nxt_basic_fuzz.c
pkillarjun: Missed it, that's why a formatter is necessary.

fuzzing/nxt_basic_fuzz.c
ac000: Hmm, we do deviate from the nginx coding style in some places and I do struggle to keep up with all the oddities we have.

Personally I'm not fussed in what order variables are declared. But here's some corroborating evidence...

```c
nxt_int_t
nxt_cgroup_proc_add(nxt_task_t *task, nxt_process_t *process)
{
    int        len;
    char       cgprocs[NXT_MAX_PATH_LEN];
    FILE       *fp;
    nxt_int_t  ret;
```

ac000:
So a lot has changed since your initial posting. Could you please rebase against current master and squash down to whatever commits you're planning on merging? and I'll take another pass over it...

pkillarjun:
> So a lot has changed since your initial posting. Could you please rebase against current master and squash down to whatever commits you're planning on merging? and I'll take another pass over it...

I think it would be better if this one is merged first: https://github.com/nginx/unit/pull/1295.

My current plan:
* Merge https://github.com/nginx/unit/pull/1295.
* Check if https://github.com/nginx/unit/pull/1291 works on OSS-Fuzz https://github.com/google/oss-fuzz/pull/12002.
* Rebase, squash, and get this merged.

pkillarjun:
> Or perhaps this is telling us that this needs splitting up into three sets of patches (all part of the same PR)...
> 
> 1. Fuzzing infrastructure
> 2. Fuzzers
> 3. Fixes ([http: fix use-of-uninitialized-value bug #1295](https://github.com/nginx/unit/pull/1295) would become part of this PR)

I tried to make things simple and easy. You can look at it now.

* [fuzzing: add fuzzing infrastructure in build system](https://github.com/nginx/unit/pull/1291/commits/6a0ffb454f61d77e1359736592d257b60c68776f) 
* [fuzzing: add fuzzing targets](https://github.com/nginx/unit/pull/1291/commits/1b4cdcd0633d6435731ff4015851933d1acbe888) 
* [fuzzing: add a fuzzing seed corpus and dictionary](https://github.com/nginx/unit/pull/1291/commits/1f5b0b1cff6d7f7492167ffff69e930c4a67a922) 

Also, let's not include https://github.com/nginx/unit/pull/1295 in this PR; there's already too much going on here.

ac000:
Hi @pkillarjun 

Thanks for the latest updates. It's looking more or less there now. Just some minor things...

1) There is a mixture of spaces and no spaces after casts in the fuzzers, my preference for new code is to have _no_ space, but I can fix that up before I merge, so don't worry about it.

2) The shell scripts are using bash, not a huge problem, are these intended to be run on non-Linux?, e.g the BSDs? where bash may well not be installed by default.

We generally use POSIX `sh` for our shell scripts for that reason...

3) Some white space [issues](https://github.com/nginx/unit/actions/runs/9477032955/job/26140445301?pr=1291).

Also the `.bin` files seem to be in DOS format (notice the `^M`s)

```diff
diff --git ./fuzzing/fuzz_http_seed_corpus/nxt_http_test_run_1.bin ./fuzzing/fuzz_http_seed_corpus/nxt_http_test_run_1.bin
new file mode 100644
index 00000000..2f6c6149
--- /dev/null
+++ ./fuzzing/fuzz_http_seed_corpus/nxt_http_test_run_1.bin
@@ -0,0 +1,2 @@
+GEt / HTTP/1.0^M
+^M
```

```console
$ file ./fuzzing/fuzz_http_seed_corpus/nxt_http_test_bench.bin
./fuzzing/fuzz_http_seed_corpus/nxt_http_test_bench.bin: ASCII text, with CRLF line terminators
```

Is that a requirement or can we convert them to Unix?


One last thing, this could maybe do with even just a brief README under fuzzing, with a quick start guide (which shell script to run...) and maybe some pre-requisites... it took me a while to figure out I needed to install the `compiler-rt` package... so on Fedora at least I guess you need the, clang, llvm and compiler-rt packages...

pkillarjun:
I believe this [little patch](https://github.com/nginx/unit/pull/1291/commits/022c86e7419afd7bbe41203766870a010a3f5999) addresse all your concerns.

> Also the .bin files seem to be in DOS format (notice the ^Ms)
> Is that a requirement or can we convert them to Unix?

Ah, `.bin` files should not be defined as anything; they are only for fuzzers to consume.
I think it would be better if we ignore all `.bin` files in `check-whitespace.yaml`

ac000:
> I believe this [little patch](https://github.com/nginx/unit/pull/1291/commits/022c86e7419afd7bbe41203766870a010a3f5999) addresse all your concerns.

Heh, OK, I can take it from here (split that up and apply the bits where required...)

> > Also the .bin files seem to be in DOS format (notice the ^Ms)
> > Is that a requirement or can we convert them to Unix?
> 
> Ah, `.bin` files should not be defined as anything; they are only for fuzzers to consume. I think it would be better if we ignore all `.bin` files in `check-whitespace.yaml`
> #### Note: This patch is generated using the OpenAI model `gpt-4o`.

Heh, if you want them to be kept in DOS format with trailing whietspace, there's no need to hack the whitespace checker!, we can use `.gitattributes` instead.

I'll add a patch for that.

Just to double check this particular one

```
fuzzing/fuzz_http_seed_corpus/nxt_http_test_run_12.bin:2: trailing whitespace.
+Host:example.com ^M
```

Is there meant to be a space character between the hostname and the `^M`?

Anyway I'll push up what I'm planning to commit for you to check over and if you can confirm about the above whitespace issue.

ac000:
- Whitespace fixes
- Added the README with some minor tweaks
- Added a `fuzzing/.gitattributes` file to ignore whitespace issues in the `.bin` files
- Added my r-b's and s-o-b's

Note I didn't change the shells seeing as this is meant to run under Linux. 

```
1:  6a0ffb45 ! 1:  a183d04d fuzzing: add fuzzing infrastructure in build system
    @@ Commit message
         fuzzing: add fuzzing infrastructure in build system
     
      ## auto/fuzzing (new) ##
     @@
    @@ fuzzing/oss-fuzz.sh (new)
     +
     +zip -r $OUT/fuzz_basic_seed_corpus.zip fuzz_basic_seed_corpus/
     +zip -r $OUT/fuzz_http_controller_seed_corpus.zip  fuzz_http_controller_seed_corpus/
    -+zip -r $OUT/fuzz_http_h1p_seed_corpus.zip  fuzz_http_h1p_seed_corpus/ 
    -+zip -r $OUT/fuzz_http_h1p_peer_seed_corpus.zip  fuzz_http_h1p_peer_seed_corpus/ 
    ++zip -r $OUT/fuzz_http_h1p_seed_corpus.zip  fuzz_http_h1p_seed_corpus/
    ++zip -r $OUT/fuzz_http_h1p_peer_seed_corpus.zip  fuzz_http_h1p_peer_seed_corpus/
     +zip -r $OUT/fuzz_json_seed_corpus.zip fuzz_json_seed_corpus/
     +
     +# Delete temporary directories.
2:  1b4cdcd0 ! 2:  9d017485 fuzzing: add fuzzing targets
    @@ Commit message
         fuzzing: add fuzzing targets
     
      ## fuzzing/nxt_basic_fuzz.c (new) ##
     @@
    @@ fuzzing/nxt_basic_fuzz.c (new)
     +    /*
     +     * Validate base64 data before decoding.
     +     */
    -+    ret = nxt_base64_decode(NULL, (u_char *) data, size);
    ++    ret = nxt_base64_decode(NULL, (u_char *)data, size);
     +    if (ret == NXT_ERROR) {
     +        return;
     +    }
     +
    -+    nxt_base64_decode(buf, (u_char *) data, size);
    ++    nxt_base64_decode(buf, (u_char *)data, size);
     +}
     +
     +
    @@ fuzzing/nxt_json_fuzz.c (new)
     +        return 0;
     +    }
     +
    -+    input.start = (u_char *) data;
    ++    input.start = (u_char *)data;
     +    input.length = size;
     +
     +    conf = nxt_conf_json_parse_str(mp, &input);
3:  1f5b0b1c ! 3:  ce538c49 fuzzing: add a fuzzing seed corpus and dictionary
    @@ Commit message
         fuzzing: add a fuzzing seed corpus and dictionary
     
      ## fuzzing/fuzz_basic_seed_corpus/base64_0.bin (new) ##
      Binary files /dev/null and fuzzing/fuzz_basic_seed_corpus/base64_0.bin differ
4:  022c86e7 < -:  -------- fuzzing: removed white space and spaces, updated build-fuzz.sh to use shell instead of bash, and added a readme
-:  -------- > 4:  b269e7a7 fuzzing: add a basic README
-:  -------- > 5:  3f44c4e9 fuzzing: Add a .gitattributes file
```

pkillarjun:
> Just to double check this particular one
> 
> ```
> fuzzing/fuzz_http_seed_corpus/nxt_http_test_run_12.bin:2: trailing whitespace.
> +Host:example.com ^M
> ```
> 
> Is there meant to be a space character between the hostname and the `^M`?

`nxt_http_test_run_12.bin` is an example of malformed data. I also ran [afl-cmin](https://github.com/AFLplusplus/AFLplusplus/blob/36db3428ab16156dd72196213d2a02a5eadaed11/docs/fuzzing_in_depth.md#b-making-the-input-corpus-unique) on `fuzz_http_seed_corpus`. All of these seed corpora have unique coverage.
My last thought always will be: don't change anything in .bin files at all.

ac000:
Rebased with master

```
$ git range-diff 3f44c4e9...35a572c2
 -:  -------- >  1:  a98acded ci: Add unit testing to unitctl CI workflow
 -:  -------- >  2:  44709bea .mailmap: Add an entry for Ava's GitHub address
 -:  -------- >  3:  30b39bd0 Add GitHub workflows for extra coverage
 -:  -------- >  4:  c30c2f5e Add unitctl quickstart to README.md
 -:  -------- >  5:  4e884d9e tools/unitctl: Replace matching image name to matching command
 -:  -------- >  6:  b91073e5 tools/unitctl: Replace format! with .to_string()
 -:  -------- >  7:  8fc16a77 Packaging: added missing build dependencies to Makefiles
 -:  -------- >  8:  f281207f Packaging: fix build-depends detection on debian-based systems
 -:  -------- >  9:  c38bcee1 contrib: be quiet on unpack
 -:  -------- > 10:  7b19a06c tstr: Constify the 'str' parameter to nxt_tstr_compile()
 -:  -------- > 11:  ea5c41b8 wasm: Add a missing 'const' qualifier in nxt_wasm_setup()
 -:  -------- > 12:  4fc50258 ci: Be more specific when to run the main Unit checks
 -:  -------- > 13:  e77a0c16 Tests: explicitly specify 'f' prefix to format string before printing
 -:  -------- > 14:  c9dced37 Tests: print unit.log on unsuccessful unmount
 -:  -------- > 15:  98983f3f Add a GitHub discussions badge to the README
 -:  -------- > 16:  a7e3686a Tools: improved error handling for unitc
 -:  -------- > 17:  d7ec30c4 ci: Limit when to run checks on pull-requests
 -:  -------- > 18:  04a24f61 http: fix use-of-uninitialized-value bug
 1:  a183d04d = 19:  965fc94e fuzzing: add fuzzing infrastructure in build system
 2:  9d017485 = 20:  a93d878e fuzzing: add fuzzing targets
 3:  ce538c49 = 21:  665353dc fuzzing: add a fuzzing seed corpus and dictionary
 4:  b269e7a7 = 22:  5b65134c fuzzing: add a basic README
 5:  3f44c4e9 = 23:  35a572c2 fuzzing: Add a .gitattributes file
```

ac000 (approved):

PR #1292, opened 2024-05-27
http: fix use-of-uninitialized-value bug

pkillarjun:
oss-fuzz Issue [68458](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=68458).


ac000:
Thanks for the patch.

The link above is giving me "Permission denied" but I take it it's the following call chain?

```
nxt_http_fields_hash()

        ret = nxt_lvlhsh_insert(hash, &lhq);                                    

nxt_lvlhsh_insert()

            return nxt_lvlhsh_bucket_insert(lhq, &lh->slot, key, -1);           

nxt_lvlhsh_bucket_insert()

        ret = nxt_lvlhsh_convert_bucket_to_level(lhq, slot, nlvl, bucket);      

nxt_lvlhsh_convert_bucket_to_level()

    lvl = proto->alloc(lhq->pool, size * (sizeof(void *)));                     
```

ac000:
Wondering why we don't hit this issue in practice...

In `src/nxt_http_parse.c::nxt_http_fields_hash()`

```c
lhc.proto,alloc
```

Is set via

```c
const nxt_lvlhsh_proto_t  nxt_http_fields_hash_proto  nxt_aligned(64) = {       
    NXT_LVLHSH_BUCKET_SIZE(64),                                                 
    { NXT_HTTP_FIELD_LVLHSH_SHIFT, 0, 0, 0, 0, 0, 0, 0 },                       
    nxt_http_field_hash_test,                                                   
    nxt_lvlhsh_alloc,                                                           
    nxt_lvlhsh_free,                                                            
}; 
...
lhq.proto = &nxt_http_fields_hash_proto;
```

Which results in something like

```gdb
(gdb) p *lhq.proto
$11 = {
  bucket_end = 12,
  bucket_size = 64,
  bucket_mask = 63,
  shift = "\005\000\000\000\000\000\000",
  test = 0x464350 <nxt_http_field_hash_test>,
  alloc = 0x40841a <nxt_lvlhsh_alloc>,
  free = 0x40843f <nxt_lvlhsh_free>
}
```

It looks like this step isn't happening in the fuzzing case, which I assume is `src/fuzz/nxt_http_parse_fuzz.c` ?

However I can't immediately see how this is being built....

pkillarjun:
> Wondering why we don't hit this issue in practice...

Because I didn't say `lhq->proto` isn't initialized; I said `lhq->pool` isn't initialized.

The `lhq->proto->alloc` call shows that there is a bug in `lhq->pool`  [nxt_lvlhsh.c#L292](https://github.com/nginx/unit/blob/master/src/nxt_lvlhsh.c#L292).

Also, my patch isn't new to this codebase; it's already been used.
* [nxt_unit.c#L6517](https://github.com/nginx/unit/blob/master/src/nxt_unit.c#L6517)
* [nxt_unit.c#L6553](https://github.com/nginx/unit/blob/master/src/nxt_unit.c#L6553)
* [nxt_unit.c#L6421](https://github.com/nginx/unit/blob/master/src/nxt_unit.c#L6421)
* [nxt_port_hash.c#L58](https://github.com/nginx/unit/blob/master/src/nxt_port_hash.c#L58)

pkillarjun:
scorched earth;

PR #1303, opened 2024-06-06
Tools: improved error handling for unitc

lcrilly:
This patch does a number of things to help when failing to apply a new
configuration.

* The error body from the Unit control API is displayed which can have
  useful troubleshooting information (when the version of curl supports it).

* When using the EDIT option, the temporary file with unapplied changes is
  preserved so that the user can edit it again without losing their work.

* Editing JavaScript modules no longer requires that module to have been
  enabled.

* Failure to apply edited JavaScript modules now rolls-back to the previous
  configuration instead of deleting the module.

### DIFF
--- a/tools/unitc
+++ b/tools/unitc
@@ -249,41 +249,58 @@ else
 	NOLOG=1
 fi
 
+# Set the base curl command after testing for newer features
+#
+$RPC_CMD curl --fail-with-body --version > /dev/null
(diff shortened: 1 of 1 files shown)


ac000 (approved):

callahad:
Looks reasonable to me. @ac000 can you handle merging please

ac000:
Rebased with master...

Issue #1317, opened 2024-06-14
CLI: subcommand for saving configuration


ava-affine:
`unitctl` should have an option to save the current running configuration, including NJS configuration and certificates.

Currently the closest analogue is for users to run `unitctl edit` and then to save the file, but this only works in interactive environments and even then does not include NJS configuration or certificates. Its also a weird path to take when all you want to do is save your configuration.

Consider the following:
`$ unitctl save golden-working-config.tar`

avahahn:
A couple user stories that this work pertains to:

- As a support person, devops engineer, or sysadmin responsible for applications deployed in one or more instances of NGINX Unit, I want to be able to save a backup of a perfectly working configuration.
- As an app developer or support person working on deploying an application with sufficiently complex configuration requirements under NGINX Unit, I want to be able to easily extract an existing configuration that I can work on.
- As an application developer I want to be able to save and distribute a working application server configuration along with my application.

Per @lcrilly's comments in https://github.com/nginx/unit/pull/1321 I think it is worth using `unitctl export` for the subcommand here.

ac000:
Can this be closed?

avahahn:
Not yet! We need to at least add support for js_modules

PR #1321, opened 2024-06-14
unitctl export subcommand

avahahn:
Currently implemented is a `unitctl` subcommand that saves Unitd configuration to a tarball. This is the first step in addressing #1317. 

Yet to be implemented is support for saving NJS modules or certificates. The code enclosed will be trivial to extend to support that later. Until then, #1317 should remain open.

### DIFF
--- a/tools/unitctl/README.md
+++ b/tools/unitctl/README.md
@@ -177,6 +177,20 @@ Imported /opt/unit/config/put.json -> /config
 Imported 3 files
 ```
 
+### Export configuration from a running Unit instance
+```
+$ unitctl export -f config.tar
+```
+
+Addtionally, standard out can be used:
+```
+$ unitctl export -f -
+$ unitctl export -f - | tar xf - config.json
+$ unitctl export -f - > config.tar
+```
+
+*Note:* The exported configuration omits certificates.
+
 ### Wait for socket to become available
 ```
 $ unitctl --wait-timeout-seconds=3 --wait-max-tries=4 import /opt/unit/config`
--- a/tools/unitctl/unitctl/src/cmd/save.rs
+++ b/tools/unitctl/unitctl/src/cmd/save.rs
@@ -1,0 +1,52 @@
+use crate::unitctl::UnitCtl;
+use crate::wait;
+use crate::UnitctlError;
+use crate::requests::send_empty_body_deserialize_response;
+use unit_client_rs::unit_client::UnitClient;
+use tar::{Builder, Header};
+use std::fs::File;
+use std::io::stdout;
+
+
+pub async fn cmd(
+    cli: &UnitCtl,
+    filename: &String
+) -> Result<(), UnitctlError> {
+    if !filename.ends_with(".tar") {
+        eprintln!("Warning: writing uncompressed tarball to {}", filename);
+    }
+
+    let control_socket = wait::wait_for_socket(cli).await?;
+    let client = UnitClient::new(control_socket);
+
+    let config_res = serde_json::to_string_pretty(
+        &send_empty_body_deserialize_response(&client, "GET", "/config").await?
+    );
+    if let Err(e) = config_res {
+        return Err(UnitctlError::DeserializationError{message: e.to_string()})
+    }
+
+    let current_config = config_res
+        .unwrap()
+        .into_bytes();
+
+    //let current_js_modules = send_empty_body_deserialize_response(&client, "GET", "/js_modules")
+    //    .await?;
+
+    let mut conf_header = Header::new_gnu();
+    conf_header.set_size(current_config.len() as u64);
+    conf_header.set_mode(0o644);
+    conf_header.set_cksum();
+
+    // builder has a different type depending on output
+    if filename == "-" {
+        let mut ar = Builder::new(stdout());
+        ar.append_data(&mut conf_header, "config.json", current_config.as_slice()).unwrap();
+    } else {
+        let file = File::create(filename).unwrap();
+        let mut ar = Builder::new(file);
+        ar.append_data(&mut conf_header, "config.json", current_config.as_slice()).unwrap();
+    }
+
+    Ok(())
+}
--- a/tools/unitctl/Cargo.lock
+++ b/tools/unitctl/Cargo.lock
@@ -485,23 +485,12 @@ checksum = "5443807d6dff69373d433ab9ef5378ad8df50ca6298caf15de6e52e24aaf54d5"
 
 [[package]]
 name = "errno"
-version = "0.3.1"
+version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4bcfec3a70f97c962c307b2d2c56e358cf1d00b558d74262b5f929ee8cc7e73a"
+checksum = "534c5cf6194dfab3db3242765c03bbe257cf92f22b38f6bc0c58d59108a820ba"
 dependencies = [
- "errno-dragonfly",
  "libc",
- "windows-sys 0.48.0",
-]
-
-[[package]]
-name = "errno-dragonfly"
-version = "0.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "aa68f1b12764fab894d2755d2518754e71b4fd80ecfb822714a1206c2aab39bf"
-dependencies = [
- "cc",
- "libc",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -509,6 +498,18 @@ name = "fastrand"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "25cbce373ec4653f1a01a31e8a5e5ec0c622dc27ff9c4e6606eefef5cbbed4a5"
+
+[[package]]
+name = "filetime"
+version = "0.2.23"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1ee447700ac8aa0b2f2bd7bc4462ad686ba06baa6727ac149a2d6277f0d240fd"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "redox_syscall",
+ "windows-sys 0.52.0",
+]
 
 [[package]]
 name = "fnv"
@@ -1009,9 +1010,9 @@ checksum = "830d08ce1d1d941e6b30645f1a0eb5643013d835ce3779a5fc208261dbe10f55"
 
 [[package]]
 name = "libc"
-version = "0.2.150"
+version = "0.2.155"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "89d92a4743f9a61002fae18374ed11e7973f530cb3a3255fb354818118b2203c"
+checksum = "97b3888a4aecf77e811145cadf6eef5901f4782c53886191b2f693f24761847c"
 
 [[package]]
 name = "libloading"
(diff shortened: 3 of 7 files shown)


tools/unitctl/unitctl/src/cmd/save.rs
```diff
@@ -0,0 +1,46 @@
...
+use std::fs::File;
+
+
+pub async fn cmd(
+    cli: &UnitCtl,
+    filename: &String
+) -> Result<(), UnitctlError> {
+    if !filename.ends_with(".tar") {
+        eprintln!("Warning: writing tarball to {}", filename);
+    }
```
ac000, line 16: Do we need to use a tarball?

If you just use `/` as the endpoint then you'll get everything, it's then very easy to use something like `jq(1)` to get just the bit you want.

It might also be useful to dump the config to stdout by default (if no filename is given) then you could do something like

```
$ unitctl save | jq '.certificiates'
```

Unless unitctl grew subcommands for just extracting certain bits of the config... it'd still be useful to be able to just dump to stdout for copy and pasting into bug reports etc...

Hmm, maybe `dump-cfg` or something would be a better name for this?

lcrilly: The problem with dumping `/` is that you cannot import it again. So I see the benefit of separating `/config`, `/certificates` and `/js_modules` into separate artefacts.

Agree with stdout by default. So you could still do
```
unitctl save | tar xf - certificates.json
```

lcrilly:
If `unitctl import` is how configuration gets _applied_ then shouldn't `unitctl export` be how configuration is retrieved?

Or perhaps that's already planned, and this is a deliberately different thing that is designed for the entire configuration? It's not clear to me what the overall goal of https://github.com/nginx/unit/issues/1317 is.

It would make it much easier to understand the intent of the PR if the commit message/comment had a sample shell snippet with an example `unitctl` interaction.

tools/unitctl/unitctl/src/cmd/save.rs
avahahn: I think the tarball is worthwhile here. Rarely will the User want to store a directory somewhere, and a directory is not something that can be distributed or shared without compression.

I also think it is an easy value add to not require the user to know how `jq` works.

avahahn:
> If unitctl import is how configuration gets applied then shouldn't unitctl export be how configuration is retrieved?

Sure, I am happy to rename the subcommand.

> It would make it much easier to understand the intent of the PR if the commit message/comment had a sample shell snippet with an example unitctl interaction.

off the top of my head:
```
$ unitctl export -f backup.tar
$ tar -xvf backup.tar
$ unitctl import backup
```

I am happy to record a small screencap of the above.

tools/unitctl/unitctl/src/cmd/save.rs
avahahn: stdout as an output target is now implemented:
```
unitctl export -f - | tar xf - config.json
```

avahahn:
Subcommand is now renamed to export

avahahn:
> The /certificates endpoint cannot be imported. Unit won't expose the secrets.

Sounds like you mean it can't be exported. That is unfortunate but I still believe this is worth implementing at least for configuration and js_modules. I will add a note to the Readme that certificates must be replicated separately.

tools/unitctl/unitctl/src/cmd/save.rs
ac000: Heh, think I'll stick to `jq </opt/unit/state/conf.json`

avahahn: You are welcome to. In fact you can use this code like this:
```
$ unitctl export -f - | tar xfO config.json | jq .....
```

avahahn:
> What do you think should happen if an exported configuration contains listeners with tls objects, and that configuration is sent to the import subcommand?

Currently import already handles pem files (and js_modules). This MR only saves the configuration JSON. 
Import will not directly read from a tar file. It is up to the user to extract the tarball and their certificates before importing. This provides a start that addresses at least the second two user stories from the linked ticket.

tools/unitctl/unitctl/src/cmd/save.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    let control_socket = wait::wait_for_socket(cli).await?;
+    let client = UnitClient::new(control_socket);
+
+    let config_res = serde_json::to_string(
+        &send_empty_body_deserialize_response(&client, "GET", "/config").await?
```
lcrilly, line 23: What's the reason for minifying the `export` content?

The Unit control socket applies formatting/prettification to the configuration which is significantly less readable when `export`ed.

```
$ curl --unix-socket /opt/homebrew/var/run/unit/control.sock _/config
{
	"routes": [
		{
			"action": {
				"share": "/Users/l.crilly/containers/unit/fallback/www$uri",
				"fallback": {
					"return": 400
				}
			}
		}
	],

	"listeners": {
		"*:9000": {
			"pass": "routes"
		}
	}
}
$ unitctl export -f - 2> /dev/null | tar xfO - config.json
{"routes":[{"action":{"share":"/Users/l.crilly/containers/unit/fallback/www$uri","fallback":{"return":400}}}],"listeners":{"*:9000":{"pass":"routes"}}}%
```

Why do we need to use serde here at all? Surely we can rely on the control API to return correctly formatted JSON?

tools/unitctl/unitctl/src/cmd/save.rs
```diff
@@ -0,0 +1,52 @@
...
+use std::fs::File;
+use std::io::stdout;
+
+
+pub async fn cmd(
+    cli: &UnitCtl,
+    filename: &String
+) -> Result<(), UnitctlError> {
+    if !filename.ends_with(".tar") {
+        eprintln!("Warning: writing tarball to {}", filename);
```
lcrilly, line 16: We already ask the user to type the name of the tar archive so why is it important to echo it back only in some circumstances?

tools/unitctl/unitctl/src/cmd/save.rs
avahahn: The CLI can return output in any number formats. I simply use this switch for colored output at my terminal. the output_format feature is not in the scope of this PR. Were you worried that the control API was not outputting proper JSON?

tools/unitctl/unitctl/src/cmd/save.rs
avahahn: In fact, if you look at the conditional you can see that there is one circumstance where I echo the filename back.
```rust
    if !filename.ends_with(".tar") {
        eprintln!("Warning: writing tarball to {}", filename);
    }
```
The conditional here, `!filename.ends_with(".tar")`, evaluates to true if and only if the filename supplied does not end in tar. The user may have entered a directory name or may have entered something like `.tar.bz` or `.tgz` expecting us to handle compression for them. 

You are asking why this is important? I have yet to work on a software project where the documentation was not overlooked and where the tooling wasn't misused by some users. Our goal here is not to prevent this behavior, but to offer the user all the tools they need to know what our command line tool is going to do with their input.

tools/unitctl/unitctl/src/cmd/save.rs
avahahn: Marking this conversation as resolved because the serialization and de-serialization features are not in scope of this PR.

ac000: So it can display the output as it received it with no processing?

I think your reply is slightly dismissive...

>  Were you worried that the control API was not outputting proper JSON?

I don't believe so, in fact quite the opposite.

The control API is returning properly formatted JSON which was them minified and made much harder to read by unitctl...

So AIUI, the question is. Why if the control API is giving us properly formatted JSON, is unitctl bothering to process it all? (rather than leave any further processing to the likes of jq(1))

callahad: Unresolving and throwing my hat in with the above.

Export is meant for archiving and exchanging Unit's state; pretty-printing makes manual inspection and editing significantly easier. Since the `/config` endpoint pretty prints itself, not having that in an export feels like a regression versus directly calling curl.

I see two solutions, both equally great:

1. Skip serde and write the raw body bytes from `GET /config`
2. Pretty-print when deserializing.

tools/unitctl/unitctl/src/cmd/save.rs
callahad: Feels like the error message could be better, but I can't think of an obvious Pareto improvement to it, so I'm inclined to regard it as _good enough._

Where my head bounces around is:

1. _Should_ we automagically handle common compressions?
2. Should the message be more explicit that it's writing an _uncompressed_ tarball?
3. Is it even possible to concisely explain _why_ we're warning the user without making everything more confusing?

The best I can come up with is: `"Warning: uncompressed tarball written to file without .tar extension: {}"`

Or maybe: `"Warning: tarball written to file with unexpected or missing extension: {}"`

Actually getting this right is _hard,_ so leaving it as it stands is probably fine for now. :)

tools/unitctl/README.md
```diff
@@ -157,6 +157,20 @@ Imported /opt/unit/config/put.json -> /config
...
+```
+
+Addtionally, standard out can be used:
+```
+$ unitctl export -f -
+$ unitctl export -f - | tar xf - config.json
+$ unitctl export -f - > config.tar
+```
+
+*Note:* The exported configuration will be devoid of any installed certificates. The Unit configuration API does not support fetching certificates or private keys once installed. User's will need to add relevant certificates and keys to an export.
```
callahad, line 172:
```suggestion
*Note:* The exported configuration omits certificates. 
```

...and then can we document importing the configuration (including the optional step of manually adding certs?)

It feels like there's a _lot_ more unitctl could do to ease cert management here, but that's best left for a future PR.

callahad (approved):
The functional code is good enough that I wouldn't block merging, but if possible I'd _really_ appreciate resolutions to the open conversations above.

tools/unitctl/unitctl/src/cmd/save.rs
avahahn: > Should we automagically handle common compressions?

I dont think so. Especially if the user needs to insert data into the tarball for certs.

> Should the message be more explicit that it's writing an uncompressed tarball?

Sounds good to me!

> Is it even possible to concisely explain why we're warning the user without making everything more confusing?

I dont know if that is necessary? I think leaving it as is notifies the user of the behavior of the procedure. That was my only goal with this eprint

callahad: Yeah, here's where I went in my head:

```console
$ unitctl export -f foo
Warning: writing tarball to foo

$ # Bad. What are you *warning* me about?
```

```console
$ unitctl export -f foo.tar.gz
Warning: writing tarball to foo.tar.gz

$ # Bad. Looks like you're warning me that you did what I wanted.
```

```console
$ unitctl export -f foo.tar.gz
Warning: writing uncompressed tarball to foo.tar.gz

$ # Good. Clear and precise, but more coincidental than smart.
```

```console
$ unitctl export -f foo
Warning: writing uncompressed tarball to foo

$ # Okay? Kind of clear, but feels like it could be better. Somehow.
```

tools/unitctl/unitctl/src/cmd/save.rs
avahahn: If you just want it pretty printed I am happy to do that. The output of the control api is deserialized which is why the formatting of the json is reset. (and why serde_json is used here).

It'll be significantly more efficient code wise to just re-add pretty printing than to implement custom request code just to have an option that doesnt deserialize the body.

tools/unitctl/README.md
avahahn: Sure. I can reword this accordingly. Certificates are imported with `unitctl import` in an example in the readme, but I agree that all the docs need to be elaborated on.

PR #1327, opened 2024-06-18
fuzzing: cifuzz workflow, fix oss-fuzz issues, updated dir name

pkillarjun:
Continuing the fuzzing integration: https://github.com/nginx/unit/pull/1291.

Adding [cifuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration) for checking build failures and PR fuzzing.
Using the main [example](https://google.github.io/oss-fuzz/getting-started/continuous-integration/#integrating-into-your-repository) provided by the OSS-Fuzz team, with an updated time of only 300 seconds.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,41 @@
+name: CIFuzz
+on:
+  pull_request:
+    paths:
+      - 'src/**'
+      - 'fuzzing/**'
+      - '.github/workflows/cifuzz.yml'
+
+permissions: {}
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'unit'
+        language: c
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'unit'
+        language: c
+        fuzz-seconds: 300
+        output-sarif: true
+    - name: Upload Crash
+      uses: actions/upload-artifact@v3
+      if: failure() && steps.build.outcome == 'success'
+      with:
+        name: artifacts
+        path: ./out/artifacts
+    - name: Upload Sarif
+      if: always() && steps.build.outcome == 'success'
+      uses: github/codeql-action/upload-sarif@v2
+      with:
+        # Path to SARIF file relative to the root of the repository
+        sarif_file: cifuzz-sarif/results.sarif
+        checkout_path: cifuzz-sarif
--- a/fuzzing/README.md
+++ b/fuzzing/README.md
@@ -43,11 +43,11 @@ $ mkdir -p build/fuzz_http_h1p_seed
 $ mkdir -p build/fuzz_http_h1p_peer_seed
 $ mkdir -p build/fuzz_json_seed
 
-$ ./build/fuzz_basic            build/fuzz_basic_seed            src/fuzz/fuzz_basic_seed_corpus
-$ ./build/fuzz_http_controller  build/fuzz_http_controller_seed  src/fuzz/fuzz_http_controller_seed_corpus
-$ ./build/fuzz_http_h1p         build/fuzz_http_h1p_seed         src/fuzz/fuzz_http_h1p_seed_corpus
-$ ./build/fuzz_http_h1p_peer    build/fuzz_http_h1p_peer_seed    src/fuzz/fuzz_http_h1p_peer_seed_corpus
-$ ./build/fuzz_json             build/fuzz_json_seed             src/fuzz/fuzz_json_seed_corpus
+$ ./build/fuzz_basic            build/fuzz_basic_seed            fuzzing/fuzz_basic_seed_corpus
+$ ./build/fuzz_http_controller  build/fuzz_http_controller_seed  fuzzing/fuzz_http_seed_corpus
+$ ./build/fuzz_http_h1p         build/fuzz_http_h1p_seed         fuzzing/fuzz_http_seed_corpus
+$ ./build/fuzz_http_h1p_peer    build/fuzz_http_h1p_peer_seed    fuzzing/fuzz_http_seed_corpus
+$ ./build/fuzz_json             build/fuzz_json_seed             fuzzing/fuzz_json_seed_corpus
 ```
 
 Here is more information about [LibFuzzer](https://llvm.org/docs/LibFuzzer.html).
--- a/fuzzing/build-fuzz.sh
+++ b/fuzzing/build-fuzz.sh
@@ -16,5 +16,5 @@ mkdir -p build/fuzz_http_h1p_peer_seed
 mkdir -p build/fuzz_json_seed
 
 echo ""
-echo "Run: ./build/\${fuzzer} build/\${fuzzer}_seed src/fuzz/\${fuzzer}_seed_corpus"
+echo "Run: ./build/\${fuzzer} build/\${fuzzer}_seed fuzzing/\${fuzzer}_seed_corpus"
 echo ""
--- a/fuzzing/nxt_http_controller_fuzz.c
+++ b/fuzzing/nxt_http_controller_fuzz.c
@@ -76,6 +76,14 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
         goto failed;
     }
 
+    r_controller->conn = nxt_mp_zget(mp, sizeof(nxt_conn_t));
+    if (r_controller->conn == NULL) {
+        goto failed;
+    }
+
+    nxt_main_log.level = NXT_LOG_ALERT;
+    r_controller->conn->log = nxt_main_log;
+
     nxt_http_fields_process(rp.fields, &nxt_controller_fields_hash,
                             r_controller);
 
--- a/fuzzing/nxt_http_h1p_fuzz.c
+++ b/fuzzing/nxt_http_h1p_fuzz.c
@@ -75,6 +75,8 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
         goto failed;
     }
 
+    r_h1p->mem_pool = mp;
+
     nxt_http_fields_process(rp.fields, &nxt_h1p_fields_hash, r_h1p);
 
 failed:
--- a/fuzzing/nxt_json_fuzz.c
+++ b/fuzzing/nxt_json_fuzz.c
@@ -4,7 +4,7 @@
 
 #include <nxt_main.h>
 #include <nxt_conf.h>
-
+#include <nxt_router.h>
 
 #define KMININPUTLENGTH 2
 #define KMAXINPUTLENGTH 1024
@@ -33,6 +33,8 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
     nxt_mp_t               *mp;
     nxt_str_t              input;
+    nxt_thread_t           *thr;
+    nxt_runtime_t          *rt;
     nxt_conf_value_t       *conf;
     nxt_conf_validation_t  vldt;
 
@@ -40,10 +42,20 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
         return 0;
     }
 
+    thr = nxt_thread();
+
     mp = nxt_mp_create(1024, 128, 256, 32);
     if (mp == NULL) {
         return 0;
     }
+
+    rt = nxt_mp_zget(mp, sizeof(nxt_runtime_t));
+    if (rt == NULL) {
+        goto failed;
+    }
+
+    thr->runtime = rt;
+    rt->mem_pool = mp;
 
     input.start = (u_char *)data;
     input.length = size;
@@ -63,6 +75,11 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
     vldt.conf = conf;
     vldt.conf_pool = mp;
     vldt.ver = NXT_VERNUM;
+
+    rt->languages = nxt_array_create(mp, 1, sizeof(nxt_app_lang_module_t));
+    if (rt->languages == NULL) {
+        goto failed;
+    }
 
     nxt_conf_validate(&vldt);
 


pkillarjun:
## FAQ

### Upload Crash

[Upload Crash](https://google.github.io/oss-fuzz/getting-started/continuous-integration/#understanding-results) uses [artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts) to save crashes found by fuzzing as artifacts.

```log
Artifacts: When the fuzzer crashes the input file that causes the crash is uploaded as an artifact.
To download the artifact, do the following steps:
```

#### Upload Sarif

There is no proper information available that I could find.
But, when a crash is found, the OSS-Fuzz infrastructure is alerted using [SARIF](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github) files. A nice detailed report is then generated for maintainers with a better UI.

My Souce: [Souce-1](https://github.com/google/oss-fuzz/issues/10090#issue-1668674254), [Souce-2](https://github.com/google/oss-fuzz/blob/master/infra/cifuzz/sarif_utils.py#L22) .

pkillarjun:
I doubt that.
```log
2024-06-18 02:24:35,946 - root - INFO - Fuzzer: fuzz_http_h1p_peer. Detected bug.
2024-06-18 02:24:35,946 - root - INFO - Trying to reproduce crash using: /tmp/tmp66kfxbhm/crash-293e3405de4ccb599b6e4f770ddf25362554727c.
```

Let's wait for 24 hours and then see.

pkillarjun:
Fix some harness bugs https://github.com/nginx/unit/pull/1327/commits/b80060f09cf2a52c05dc8f0e8f70acea637f8c09
* https://oss-fuzz.com/testcase-detail/5276662569172992
* https://oss-fuzz.com/testcase-detail/4581551606661120

src/nxt_conf_validation.c
```diff
@@ -1551,7 +1551,7 @@ static nxt_int_t
 nxt_conf_vldt_var(nxt_conf_validation_t *vldt, const nxt_str_t *name,
     nxt_str_t *value)
 {
-    u_char  error[NXT_MAX_ERROR_STR];
+    u_char  error[NXT_MAX_ERROR_STR] = {0};
 
     if (nxt_tstr_test(vldt->tstr_state, value, error) != NXT_OK) {
         return nxt_conf_vldt_error(vldt, "%s in the \"%V\" value.",
```
ac000, line 1557: Unless you have any more details, this looks like a false positive.

```c
    if (nxt_tstr_test(vldt->tstr_state, value, error) != NXT_OK) {              
        return nxt_conf_vldt_error(vldt, "%s in the \"%V\" value.",             
                                   error, name);                                
    }
```

When we return from `nxt_tstr_test()` there are two possibilities

1) The return value is NXT_OK & error hasn't been set.
2) The return value is NXT_ERROR & error **has** been set.

In the case of (2) we then format the error message.

pkillarjun: I know some simple things: static analyzers have false positives, dynamic analyzers don't. This is why I don't use static analyzers except for Cyclomatic Complexity.

You are right, the provided patch  https://github.com/nginx/unit/pull/1327/commits/576a18b70b43deb95a38d7bfd1ecb29132a60d5a is wrong.

I did some debugging,
When `nxt_tstr_test` is called, `nxt_tstr_test` calls `nxt_sprintf`.
```c
nxt_int_t
nxt_tstr_test(nxt_tstr_state_t *state, nxt_str_t *str, u_char *error)
{
    //...
#else
        nxt_sprintf(error, error + NXT_MAX_ERROR_STR,
                    "Unit is built without support of njs: "
                    "\"--njs\" ./configure option is missing.");
        return NXT_ERROR;
#endif

  //...

}
```

Which ends up calling `nxt_vsprintf`, This is the only code that is executed.
```c
u_char *
nxt_vsprintf(u_char *buf, u_char *end, const char *fmt, va_list args)
{
    // ...

    while (*fmt != '\0' && buf < end) {

        /*
         * "buf < end" means that we could copy at least one character:
         * a plain character, "%%", "%c", or a minus without test.
         */

        if (*fmt != '%') {
            *buf++ = *fmt++;
            continue;
        }


    // ...

    }

    return buf;
}
```

The `while` loop copies everything except `'\0'`.

pkillarjun: ### Patch to reproduce in controlled environment.

Well, I first thought I could be wrong, so I did some tests.

```patch
diff --git a/fuzzing/build-fuzz.sh b/fuzzing/build-fuzz.sh
index 04f080d9..2a11870e 100644
--- a/fuzzing/build-fuzz.sh
+++ b/fuzzing/build-fuzz.sh
@@ -2,8 +2,8 @@
 
 export CC=clang
 export CXX=clang++
-export CFLAGS="-g -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address,undefined -fsanitize=fuzzer-no-link"
-export CXXFLAGS="-g -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address,undefined -fsanitize=fuzzer-no-link"
+export CFLAGS="-O0 -fno-omit-frame-pointer -gline-tables-only -Wno-error=enum-constexpr-conversion -Wno-error=incompatible-function-pointer-types -Wno-error=int-conversion -Wno-error=deprecated-declarations -Wno-error=implicit-function-declaration -Wno-error=implicit-int -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=memory -fsanitize-memory-track-origins -fsanitize=fuzzer-no-link"
+export CXXFLAGS="-O0 -fno-omit-frame-pointer -gline-tables-only -Wno-error=enum-constexpr-conversion -Wno-error=incompatible-function-pointer-types -Wno-error=int-conversion -Wno-error=deprecated-declarations -Wno-error=implicit-function-declaration -Wno-error=implicit-int -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=memory -fsanitize-memory-track-origins -fsanitize=fuzzer-no-link -stdlib=libc++"
 export LIB_FUZZING_ENGINE="-fsanitize=fuzzer"
 
 ./configure --no-regex --no-pcre2 --fuzz=$LIB_FUZZING_ENGINE
diff --git a/src/nxt_sprintf.c b/src/nxt_sprintf.c
index 875f43a5..a606301f 100644
--- a/src/nxt_sprintf.c
+++ b/src/nxt_sprintf.c
@@ -160,6 +160,18 @@ nxt_vsprintf(u_char *buf, u_char *end, const char *fmt, va_list args)
                 continue;
             }
 
+            /*
+             * Just lazy work.
+             */
+            printf("\nPrint p at 75 as hax : %02x", (unsigned char)p[75]);
+            printf("\nPrint p at 75 as char: %c",   (unsigned char)p[75]);
+            printf("\nPrint value of error in nxt_vsprintf as p: %s \n", p);
+
+            /*
+             * There is no byte, this should crash here.
+             */
+            printf("\nPrint p at 76 as hax : %02x", (unsigned char)p[76]);
+
             while (*p != '\0' && buf < end) {
                 *buf++ = *p++;
             }
```

[clusterfuzz-testcase-minimized-fuzz_json-5545917827055616.txt](https://github.com/user-attachments/files/15935479/clusterfuzz-testcase-minimized-fuzz_json-5545917827055616.txt)

pkillarjun: Just for fun, you can try adding this hacky line in `nxt_conf_validation.c` to see if everything works as expected afterward.

```patch
--- a/src/nxt_conf_validation.c
+++ b/src/nxt_conf_validation.c
@@ -1554,6 +1554,7 @@ nxt_conf_vldt_var(nxt_conf_validation_t *vldt, const nxt_str_t *name,
     u_char  error[NXT_MAX_ERROR_STR] = {0};
 
     if (nxt_tstr_test(vldt->tstr_state, value, error) != NXT_OK) {
+        error[76] = '\0';
         return nxt_conf_vldt_error(vldt, "%s in the \"%V\" value.",
                                    error, name);
     }
```

pkillarjun: I think it would be better if I remove https://github.com/nginx/unit/pull/1327/commits/576a18b70b43deb95a38d7bfd1ecb29132a60d5a from this PR. You can fix it in a separate PR with `Issue: 69744`.

hongzhidao: Hi @pkillarjun,
Good catch. It's an issue that missed `\0` at the end of `nxt_sprintf()` in `nxt_tstr_test()`.
Your way to fix it is correct, but I'd suggest fixing it like this:
```
diff --git a/src/nxt_tstr.c b/src/nxt_tstr.c
index fde4822d..6f39cff2 100644
--- a/src/nxt_tstr.c
+++ b/src/nxt_tstr.c
@@ -159,7 +159,7 @@ nxt_tstr_test(nxt_tstr_state_t *state, nxt_str_t *str, u_char *error)
 #else
         nxt_sprintf(error, error + NXT_MAX_ERROR_STR,
                     "Unit is built without support of njs: "
-                    "\"--njs\" ./configure option is missing.");
+                    "\"--njs\" ./configure option is missing.%Z");
         return NXT_ERROR;
 #endif
```
There are more than one places to use `NXT_MAX_ERROR_STR`, and it's better to keep it just one way.
We usually add `%Z` on demand. In this case, it's an issue since it's used in the format `"%s in this \"%V\" value."`.
Thanks.

pkillarjun: `u_char  error[NXT_MAX_ERROR_STR] = {0};` This only fixes the symptoms, not the actual problem.

hongzhidao: Hmm, I mean we don't need to set it zero but append `%z` instead.

pkillarjun: Oh, sorry, missed it. I thought `%z` was for extra formatting.
I thought it was a note to yourself to add extra features instead of fixing this bug.

hongzhidao: Correct.

ac000 (commented):
This

```
conf: add %Z to resolve oss-fuzz issue 69758 use-of-uninitialized-value

```

Could do with a commit message, I wasn't even aware (or if I was I had completely forgotten) that nxt_sprintf() doesn't nul-terminate strings by default.

But then at the _very least_ it probably also needs a

```
Co-developed-by: Zhidao HONG <z.hong@f5.com>
Signed-off-by: Zhidao HONG <z.hong@f5.com>
```

@hongzhidao are you OK with that?

The commit subject prefix could probably be `tstr, conf:`. Perhaps something like

```
tstr, conf: Ensure error strings are nul-terminated
```

Are these bug reports publicly accessible?

If not I don't think we should mention specific bug numbers but just a general 'found with oss-fuss`, if so, it would be better to reference it with a link

```
Link: <https://oss-fuzz.com/testcase-detail/5545917827055616>
```

I think it would actually be better to open a separate pull-request for this where we can discuss the finer details and to get this fix in...

hongzhidao:
I'm fine with it.
It can be a separate fix.

>  I wasn't even aware (or if I was I had completely forgotten) that nxt_sprintf() doesn't nul-terminate strings by default.
Yep, it's not nul-temmiate by default.

> The commit subject prefix could probably be tstr, conf:. Perhaps something like
tstr, conf: Ensure error strings are nul-terminated

We could show the concrete function `nxt_tstr_test()` which is not used there.

Btw, I'd like to use `pkillarjun` as the signed-off instead of me. He reported the issue and fixed it with my suggestion.

ac000:
> I'm fine with it. It can be a separate fix.

It it's in a separate PR then it's not reliant on getting this whole thing in...
 
> > I wasn't even aware (or if I was I had completely forgotten) that nxt_sprintf() doesn't nul-terminate strings by default.
> 
> Yep, it's not nul-temmiate by default.
> 
> > The commit subject prefix could probably be tstr, conf:. Perhaps something like
> > tstr, conf: Ensure error strings are nul-terminated
> 
> We could show the concrete function `nxt_tstr_test()` which is not used there.

Yes, that would be part of the commit message explaining that you need to explicitly nul-terminate the string.

> Btw, I'd like to use `pkillarjun` as the signed-off instead of me. He reported the issue and fixed it with my suggestion.

You can have multiple `Signed-off-by:`'s and a `Co-developed-by:` should be followed by a `Signed-off-by:`

ac000:
@hongzhidao @pkillarjun 

If you're good with the commit message etc, then I'm happy to merge this patch in and then it can simply be dropped from this PR...

ac000:
PR for bug fix https://github.com/nginx/unit/pull/1342

ac000:
OK @pkillarjun you can remove this patch from this PR now, thanks!

```
conf: add %Z to resolve oss-fuzz issue 69758 use-of-uninitialized-value
```

pkillarjun:
This can be merge now.

ac000 (commented):
Hi, regarding

```
fuzzing: fix Null-dereference 69754, 69745, 69741 
```

I assume those numbers are some bug ticket references?

As a rule we don't put those kinds of things in the commit subject. If those things are publicly accessible then you can reference them via `Link:` tags...

However...

This is just in the fuzzing code, not in Unit right?

So is this fixing fuzzing code that was previously committed?

If so it could do with a `Fixes:` tag, e.g

```
Fixes: SHORT_HASH ("COMMIT_SUBJECT")
```

E.g

```
Fixes: c1e3f02f9 ("Compile with -fno-strict-overflow")
```
```

**Review by ac000 — COMMENTED** [review]
Nuts!

```
fuzzing: updated build-fuzz.sh help echo dir
```

This could do with a 

```
Fixes: 965fc94e4 ("fuzzing: add fuzzing infrastructure in build system")
```

I also notice the `./fuzzing/README.md` still has a bunch of references to `src/fuzz/`

**Comment by pkillarjun** [comment]
update: Will complete by next Sunday.

**Comment by pkillarjun** [comment]
Request for re-review.

**Comment by ac000** [comment]
Hi @pkillarjun 

The `Fixes:` tag is like any other tag and goes at the end of the commit message, not in the subject!

If those commits are fixing previous commits, please explain in the commit message what they are fixing.

File: .github/workflows/cifuzz.yml [hunk 2]
```diff
@@ -0,0 +1,35 @@
...
+        output-sarif: true
+    - name: Upload Crash
+      uses: actions/upload-artifact@v3
+      if: failure() && steps.build.outcome == 'success'
+      with:
+        name: artifacts
+        path: ./out/artifacts
+    - name: Upload Sarif
+      if: always() && steps.build.outcome == 'success'
+      uses: github/codeql-action/upload-sarif@v2
```
ac000, line 31: I notice this warning

> Warning: CodeQL Action v2 will be deprecated on December 5th, 2024. Please update all occurrences of the CodeQL Action in your workflow files to v3. For more information, see https://github.blog/changelog/2024-01-12-code-scanning-deprecation-of-codeql-action-v2/

.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,35 @@
+name: CIFuzz
+on: [pull_request]
```
ac000, line 2: I guess this only needs to run when stuff under `./src` changes, even then we may be able to limit it some more as we currently only cover some specific areas...

So at the very least

```yaml

  paths:
    - 'src/**'
```

.github/workflows/cifuzz.yml
pkillarjun: > I notice this warning
> 
> > Warning: CodeQL Action v2 will be deprecated on December 5th, 2024. Please update all occurrences of the CodeQL Action in your workflow files to v3. For more information, see https://github.blog/changelog/2024-01-12-code-scanning-deprecation-of-codeql-action-v2/

Ignore this, [cifuzz.yml](https://google.github.io/oss-fuzz/getting-started/continuous-integration/#integrating-into-your-repository) is still using `uses: github/codeql-action/upload-sarif@v2`.

.github/workflows/cifuzz.yml
pkillarjun: Will this [patch](https://github.com/nginx/unit/pull/1327/files#diff-cb0618c14c125bb79d60ba9a127ed18f4a3ce62cff9c49e26c8890afb6f35668R2-R6) work?

I have another follow-up patch on my disk. I will apply it as soon as this one looks good to you.

ac000: I wonder if it should also include the fuzzing stuff itself... `fuzzing/**`

pkillarjun:
> Hi @pkillarjun
> 
> The `Fixes:` tag is like any other tag and goes at the end of the commit message, not in the subject!
> 
> If those commits are fixing previous commits, please explain in the commit message what they are fixing.

Alright, thanks.
Ah, can you rereviw this now, i think these new comment will do the job.

ac000 (approved):
Thanks Arjun!

PR #1331, opened 2024-06-18
contrib: updated njs to 0.8.4

thresheek:
While at it, follow the njs move to github to fetch sources.


ac000:
We do have a patch pending to bring compatibility with njs 0.8.5 when it's released.

But yeah, no harm if we bump this again before our next release...

ac000 (approved):

PR #1337, opened 2024-06-21
ci: Fix tags on ad hoc unitctl releases

callahad:
- Adds `unitctl/` prefix to tags generated by manual workflow runs. Previously, only release titles (but not tags) were prefixed.

- Omits superfluous `name` field; falls back to `tag` when absent.

- Removes unnecessary conditional from `prelease` field.

This results in the following tagging / releasing behavior:

1. Running manually creates a pre-release and tags it `unitctl/VERSION`
2. Pushing a tag formatted like `x.y.z` creates a normal release

Refines: 3501a50ffb93756e145295021ff9313ac77f1ba9

### DIFF
--- a/.github/workflows/unitctl.yml
+++ b/.github/workflows/unitctl.yml
@@ -185,10 +185,8 @@ jobs:
         uses: ncipollo/release-action@v1
         with:
           artifacts: "unitctl-*"
-          # false if triggered by a tag
-          prerelease: ${{github.event_name == 'workflow_dispatch' && true}}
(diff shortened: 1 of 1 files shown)


callahad:
One other important part of this: Since previously, _all_ releases were unconditionally titled with the "unitctl/" prefix, pushing a release tag (e.g., `1.33.0`) would create a GitHub Release titled "unitctl/1.33.0".

Feels like when we make a release of Unit, the GitHub Release should be titled "1.33.0" without the "unitctl/" prefix.

avahahn (approved):
Looks good to me!

Issue #1318, opened 2024-06-14
CLI: deploy configuration to many instances


ava-affine:
`unitctl import` and `unitctl execute` should be able to send configuration snippets to a configurable list of Unit instances. Currently when multiple Unit instances exist on a host `unitctl` will ask for clarification on which control socket/endpoint to use. This should be extended to support a list. Under no circumstances should `unitctl` automatically broadcast to many Unit instances without a specific invocation from the user.

lcrilly:
Does this apply to multiple instances on the host where `unitctl` is executed, or to multiple instances over a network?

avahahn:
It should apply for any control socket that unitctl can reach.
At the moment `unitctl` knows how to access control sockets over TCP and via a file socket.
If the user sets up some number of proxies for instances over the network then we shouldn't refuse to operate on them.

avahahn:
Just want to drop a quick note here on implementation:

Currently the CLI has a global flag for specifying a control socket when there are multiple Unit instances detected. This needs to be refactored to accept either a single socket or a list of sockets. 

Additionally, code will be needed in main.rs to apply the specified cli command to each provided control socket. Perhaps socket should be an argument to each command in the cmd submodule. Existing commands should be checked to make sure that they arent manually handling the control socket flag.

PR #1350, opened 2024-07-03
Enable Multi Socket Support

avahahn:
# Description
This commit refactors the CLI code to accept
multiple instances of the control socket flag.
All subcommands except for edit and save now
support being run against multiple specified
instances of unitd.

# Changes
* control_socket_addresses CLI field is now a vector
* centralize error related logic into the error module
* wait_for_socket now returns a vector of sockets. all sockets in vector are waited upon and validated
* extraneous code is removed
* applications, execute, import, listeners, and status commands all run against N control sockets now
* edit and save commands return error when run against a single control socket

Fixes: #1318

### DIFF
--- a/tools/unitctl/unitctl/src/wait.rs
+++ b/tools/unitctl/unitctl/src/wait.rs
@@ -8,105 +8,83 @@ use unit_client_rs::unitd_instance::UnitdInstance;
 /// Waits for a socket to become available. Availability is tested by attempting to access the
 /// status endpoint via the control socket. When socket is available, ControlSocket instance
 /// is returned.
-pub async fn wait_for_socket(cli: &UnitCtl) -> Result<ControlSocket, UnitctlError> {
-    // Don't wait, if wait_time is not specified
-    if cli.wait_time_seconds.is_none() {
-        return cli.control_socket_address.instance_value_if_none().await.and_validate();
+pub async fn wait_for_sockets(cli: &UnitCtl) -> Result<Vec<ControlSocket>, UnitctlError> {
+    let socks: Vec<ControlSocket>;
+    match &cli.control_socket_addresses {
+        None => {
+            socks = vec![find_socket_address_from_instance().await?];
+        },
+        Some(s) => socks = s.clone(),
     }
 
-    let wait_time =
-        Duration::from_secs(cli.wait_time_seconds.expect("wait_time_option default was not applied") as u64);
-    let max_tries = cli.wait_max_tries.expect("max_tries_option default was not applied");
-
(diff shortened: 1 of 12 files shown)


ac000:
A few lines in the README with trailing whitespace.

View them with something like `git diff master.. --check` or `git show --check` etc...

dekobon (approved):
LGMT. Feedback submitted via an analog medium.

tools/unitctl/unitctl/src/wait.rs
```diff
@@ -8,113 +8,91 @@ use unit_client_rs::unitd_instance::UnitdInstance;
...
+            mapped.push(addr.to_owned().validate()?);
+            continue;
         }
 
-        attempt += 1;
-
-        let result = cli.control_socket_address.instance_value_if_none().await.and_validate();
+        let wait_time =
+            Duration::from_secs(cli.wait_time_seconds.expect("wait_time_option default was not applied") as u64);
+        let max_tries = cli.wait_max_tries.expect("max_tries_option default was not applied");
```
javorszky, line 29: is it customary to use `expect` in cli applications? Panicking here is not as big of a deal as in a service application. If the cli panics, are there any drawbacks to that?

tools/unitctl/unitctl/src/wait.rs
```diff
@@ -8,113 +8,91 @@ use unit_client_rs::unitd_instance::UnitdInstance;
...
+
+        let mut attempt = 0;
+        while attempt < max_tries {
+            if attempt > 0 {
+                eprintln!(
```
javorszky, line 34: I feel like this should not go to stderr 🤔

javorszky (changes requested):
Two tiny nitpicks. The waiting to stderr is the bigger one. I'm happy to leave code unchanged with explanations though 🙂

tools/unitctl/unitctl/src/wait.rs
avahahn: No, It basically just exits the flow and prints to stderr

tools/unitctl/unitctl/src/wait.rs
avahahn: In a previous review @dekobon made a really good point that any info printing should go to stderr as stdout should be reserved for serialized output formats.

avahahn: The idea here being that if the user is piping output to somewhere we should not prepend, interrupt, or trail valid formatted data with info logging

javorszky (approved):

PR #1351, opened 2024-07-04
contrib: make sha512sum check compatible with FreeBSD implementation

thresheek:
FreeBSD introduced sha512sum binary in version 14, but with slightly incompatible flags as compared to Linux version.  This change makes it work in both worlds.

### DIFF
--- a/pkg/contrib/Makefile
+++ b/pkg/contrib/Makefile
@@ -51,7 +51,7 @@ XZ ?= $(error XZ (LZMA) compressor not found)
 endif
 
 ifeq ($(shell sha512sum --version >/dev/null 2>&1 || echo FAIL),)
-SHA512SUM = sha512sum --check
+SHA512SUM = sha512sum -c -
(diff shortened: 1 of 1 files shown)


thresheek:
To add some context, this fixes the buildbot builds on FreeBSD14 that utilize njs from contribs

ac000 (commented):
I'd suggest to wrap the commit message such as (otherwise it'll look naff in an 80 col terminal)

```
FreeBSD introduced sha512sum binary in version 14, but with slightly
incompatible flags as compared to Linux version.  This change makes it
work in both worlds.
```

callahad:
Looks good to us pending commit message tweak

thresheek:
Patch updated, thanks!

ac000:
Hi @thresheek could you double check what you pushed up as the commit message doesn't seem to have changed (though the subject has)...

thresheek:
Ugh, I could have sworn I did a gg=G.  Did it now for sure.

ac000 (approved):
Bingo!

PR #1373, opened 2024-07-31
docs: Update unit-openapi.yaml

Jcahilltorre:
Update openapi for the modules status object

### DIFF
--- a/docs/unit-openapi.yaml
+++ b/docs/unit-openapi.yaml
@@ -6361,6 +6361,7 @@ components:
       properties:
         modules:
           $ref: "#/components/schemas/statusModules"
+
         connections:
           $ref: "#/components/schemas/statusConnections"
 
--- a/docs/unit-openapi.yaml
+++ b/docs/unit-openapi.yaml
@@ -6361,7 +6361,6 @@ components:
       properties:
         modules:
           $ref: "#/components/schemas/statusModules"
-          
         connections:
           $ref: "#/components/schemas/statusConnections"
 
--- a/docs/unit-openapi.yaml
+++ b/docs/unit-openapi.yaml
@@ -3906,7 +3906,7 @@ paths:
       operationId: getStatus
       summary: "Retrieve the status object"
       description: "Retrieves the entire `/status` section that represents
-        Unit's [usage statistics](https://unit.nginx.org/usagestats/)."
+        Unit's [usage statistics and list of loaded modules](https://unit.nginx.org/statusapi/)."
 
       tags:
         - status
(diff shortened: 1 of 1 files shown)


docs/unit-openapi.yaml
```diff
@@ -6331,6 +6371,11 @@ components:
         applications:
           $ref: "#/components/schemas/statusApplications"
 
+    # /status/modules
+    statusModules:
+      description: "Lists currently loaded language modules."
+      type: object
+
```
Jcahilltorre, line 6378: This section is incomplete

docs/unit-openapi.yaml
```diff
@@ -6319,9 +6356,12 @@ components:
 
     # /status
     status:
-      description: "Represents Unit's usage statistics."
+      description: "Represents Unit's loaded language modules and usage statistics."
       type: object
       properties:
+        modules:
+          $ref: "#/components/schemas/statusModules"
+
```
Jcahilltorre, line 6364:
```suggestion

```

docs/unit-openapi.yaml
```diff
@@ -3924,13 +3924,38 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/statusModules"
+
+              examples:
+                example1:
+                  $ref: "#/components/examples/statusModules"
+
```
Jcahilltorre, line 3951:
```suggestion

```

docs/unit-openapi.yaml
```diff
@@ -3924,13 +3924,37 @@ paths:
...
+            the configuration."
+
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/statusModules"
+
+              examples:
+                example1:
+                  $ref: "#/components/examples/statusModules"
```
Jcahilltorre, line 3950:
```suggestion
                  $ref: "#/components/examples/statusModules"

```

docs/unit-openapi.yaml
```diff
@@ -6319,9 +6356,11 @@ components:
 
     # /status
     status:
-      description: "Represents Unit's usage statistics."
+      description: "Represents Unit's loaded language modules and usage statistics."
       type: object
       properties:
+        modules:
+          $ref: "#/components/schemas/statusModules"
```
Jcahilltorre, line 6363:
```suggestion
          $ref: "#/components/schemas/statusModules"

```

ac000 (changes requested):
Hi @Jcahilltorre Just a heads up because we want to keep a clean commit history in Unit, without a bunch of unnecessary commits which just adds noise to the commit history.

Taking a quick look at the patches, it seems to me these should be done as two commits as these are two logical changes

1) A first commit that does the URL change
2) A second commit that adds the info for the new modules endpoint

Having the next version of the patches presented like that would help with reviewing!.

It also seems there are some [whitespace issues](https://github.com/nginx/unit/pull/1373/checks)...

Thanks for the patch! I'm still not sure what this is actually used for and if there is anyway to verify its correctness?

PR #1377, opened 2024-08-01
unitctl: clean up output format text

### LINKED ISSUE #1375: unitctl: fix up the output format tag's default
Currently the help text looks like this:

```
  -t, --output-format <OUTPUT_FORMAT>  Output format: text, yaml, json, json-pretty (default) [default: text] [possible values: yaml, json, json-pretty, text]
```

It's confusing to know whether the default is `text` or `json-pretty`.

javorszky:
After figuring out what and how clap deals with input enums and how those enums end up being serialised to the terminal, I reworded the output format message. Both the `default` and the `possible values` are programmatic and clap will print them anyways, there is no reason to include the same detail again in the help message.

This helps us avoid inconsistencies between functionality, code that clap makes use of, and freeform description that clap doesn't have knowledge about.

I alsoi changed the default output format of two other subcommands to be `json-pretty` rather than `text` to align with the rest of the subcommands.

Now every subcommand has the exact same output format help:

```
Options:
  -t, --output-format <OUTPUT_FORMAT>  Output format of the result [default: json-pretty] [possible values: yaml, json, json-pretty, text]
```

Closes #1375

### DIFF
--- a/tools/unitctl/unitctl/src/unitctl.rs
+++ b/tools/unitctl/unitctl/src/unitctl.rs
@@ -51,7 +51,7 @@ pub(crate) enum Commands {
             short = 't',
             long = "output-format",
             default_value = "json-pretty",
-            help = "Output format: yaml, json, json-pretty (default)"
+            help = "Output format: yaml, json, json-pretty"
         )]
         output_format: OutputFormat,
     },
@@ -68,7 +68,7 @@ pub(crate) enum Commands {
             short = 't',
             long = "output-format",
             default_value = "json-pretty",
-            help = "Output format: yaml, json, json-pretty (default)"
+            help = "Output format: yaml, json, json-pretty"
         )]
         output_format: OutputFormat,
         #[arg(
@@ -98,7 +98,7 @@ pub(crate) enum Commands {
             short = 't',
             long = "output-format",
             default_value = "json-pretty",
-            help = "Output format: yaml, json, json-pretty (default)"
+            help = "Output format: yaml, json, json-pretty"
         )]
         output_format: OutputFormat,
     },
@@ -110,7 +110,7 @@ pub(crate) enum Commands {
             short = 't',
             long = "output-format",
             default_value = "json-pretty",
-            help = "Output format: yaml, json, json-pretty (default)"
+            help = "Output format: yaml, json, json-pretty"
         )]
         output_format: OutputFormat,
     },
@@ -136,7 +136,7 @@ pub struct InstanceArgs {
         short = 't',
         long = "output-format",
         default_value = "text",
-        help = "Output format: text, yaml, json, json-pretty (default)"
+        help = "Output format: text, yaml, json, json-pretty"
     )]
     pub output_format: OutputFormat,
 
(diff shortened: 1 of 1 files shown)


ac000:
Should really be using `tools/unitctl:` as the commit subject prefix...

ac000:
Should really be using `tools/unitctl:` for the commit subject prefix...

javorszky:
> Should really be using `tools/unitctl:` for the commit subject prefix...

Fixed. An observation: `tools/unitctl: ` takes up a lot of space, and it's very hard to write a descriptive subject before running out of the 50 char length

lcrilly:
The previous convention was simply "Tools: …". `unitctl` changed it. See
```
git log | grep -i '    tools'
```

avahahn:
> The previous convention was simply "Tools: …". unitctl changed it. See

That was at the direct request of @ac000. I would be happy to revert to simply prefixing with "Tools:" if that sits well with all parties.

ac000:
> The previous convention was simply "Tools: …". `unitctl` changed it. See
> 
> ```
> git log | grep -i '    tools'
> ```

That's when the `tools` directory was simpler.

ac000:
> > Should really be using `tools/unitctl:` for the commit subject prefix...
> 
> Fixed. An observation: `tools/unitctl: ` takes up a lot of space, and it's very hard to write a descriptive subject before running out of the 50 char length

You can happily go to 65...

ac000:
> > The previous convention was simply "Tools: …". unitctl changed it. See
> 
> That was at the direct request of @ac000. I would be happy to revert to simply prefixing with "Tools:" if that sits well with all parties.

`Tools: setup-unit:` has been used from early on (once there was more than one thing in there), `tools/unitctl` simply aligns with that...

See:

```
$ git log --oneline tools/
```

avahahn (approved):
I like the change! Good catch :)

javorszky:
> > > Should really be using `tools/unitctl:` for the commit subject prefix...
> > 
> > 
> > Fixed. An observation: `tools/unitctl: ` takes up a lot of space, and it's very hard to write a descriptive subject before running out of the 50 char length
> 
> You can happily go to 65...

To triple check, I looked up the 72 char length, and the subject is usually 50 characters per https://www.gitkraken.com/learn/git/best-practices/git-commit-message and a bunch others and @callahad 's info. GitHub breaks the first line at 72 characters and moves the rest to the first line in the body.

Going forward I'm happy to keep to 65, mainly so we don't have any more conversations on how long the subject line should be, though ideally I'd hard cap it at 72 so GitHub doesn't wrap it.

ac000:
> > > > Should really be using `tools/unitctl:` for the commit subject prefix...
> > > 
> > > 
> > > Fixed. An observation: `tools/unitctl: ` takes up a lot of space, and it's very hard to write a descriptive subject before running out of the 50 char length
> > 
> > 
> > You can happily go to 65...
> 
> To triple check, I looked up the 72 char length, and the subject is usually 50 characters per https://www.gitkraken.com/learn/git/best-practices/git-commit-message and a bunch others and @callahad 's info. GitHub breaks the first line at 72 characters and moves the rest to the first line in the body.

50 characters for the subject is a general guide. For example, vim will highlight when you go over it,  But you still want to try and avoid the subject from wrapping (keep things git-shortlog(1) in mind).

> Going forward I'm happy to keep to 65, mainly so we don't have any more conversations on how long the subject line should be, though ideally I'd hard cap it at 72 so GitHub doesn't wrap it.

Yes, I make that point [here](https://gist.github.com/ac000/4925906720a771491c09748f40ef50ba)

> Again, don't worry too much about going over, just try to keep it 72 or less.

Issue #1042, opened 2023-12-30
How to manually set the number of router threads


pbromb:
Hello,
I am running an application using Nginx Unit in Kubernetes environment.
The `unit: router` process automatically runs in the number of threads, which equals the number of CPU cores on the physical server. 

### src/nxt_router.c
```
if (rtcf->threads == 0) {
    rtcf->threads = nxt_ncpu;
}

```
in the code I also found 
### src/nxt_router.c
```
static nxt_conf_map_t  nxt_router_conf[] = {
    {
        nxt_string("listeners_threads"),
        NXT_CONF_MAP_INT32,
        offsetof(nxt_router_conf_t, threads),
    },
};
```

but I was not able to set this from the configuration file

Is it possible to set this value ?

ac000:
On Sat, 30 Dec 2023 06:46:06 -0800
Paweł B ***@***.***> wrote:

> Hello,
> I am running an application using Nginx Unit in Kubernetes environment.
> The `unit: router` process automatically runs in the number of threads, which equals the number of CPU cores on the physical server. 
> 
> ### src/nxt_router.c
> ```
> if (rtcf->threads == 0) {
>     rtcf->threads = nxt_ncpu;
> }
> 
> ```
> in the code I also found 
> ### src/nxt_router.c
> ```
> static nxt_conf_map_t  nxt_router_conf[] = {
>     {
>         nxt_string("listeners_threads"),
>         NXT_CONF_MAP_INT32,
>         offsetof(nxt_router_conf_t, threads),
>     },
> };
> ```
> 
> 
> but I was not able to set this from the configuration file
> 
> Is it possible to set this value ?

Looks like it isn't currently hooked up to the configuration system...

juliantaylor:
There needs to be a way to configure the number of threads to use in particular as unit is using sched_yield using spinlocks with no sleep which have extremely bad performance when overcommitted.

Running unit on a host with 64 cores (including hyperthreads) with a cpu limit of 2 under load will cause the system to spend 90% or more of its time rescheduling its threads waiting for the lock while the process which is actually holding the lock only has a very small chance of running and releasing it.

A sample profile of an application under this load looks like this:

```
-   92.81%     0.00%  unitd    [unknown]                                         [k] 0x0000000000000004                                                                                                           ◆
   - 0x4                                                                                                                                                                                                          ▒
      - 91.42% 0x7f844e03f527                                                                                                                                                                                     ▒
[...]

PR #1383, opened 2024-08-12
Better available CPU count determination on Linux

ac000:
This pull-request enhances how we determine the number of CPU's available to Unit to run on under Linux.

In a recent [discussion](https://github.com/nginx/unit/issues/1042#issuecomment-2282308293) it became clear there are cases whereby we can end up creating way too many router threads (these are what handle client connections).

We currently try to create a number of router threads (not incl the main one) that match the number of on-line CPU's. This is currently determined with

```c
n = sysconf(_SC_NPROCESSORS_ONLN);
```

While that generally works, there are cases where it can give the wrong answer, for us anyway. I'll use the nproc(1) utility to demonstrate

```sh
$ nproc 
4
# 4 cpu system, 4 on-line cpu's
# Let's limit nproc to just 2 cpu's
$ taskset -c 0-1 nproc 
2
# Let's ignore the cpu allowed mask
$ taskset -c 0-1 nproc --all
4
```

nproc with no arguments will (on Linux) use sched_getaffinity(2) to determine the number of CPU's that it's allowed to run on. Passing `--all` makes it use sysconf(3) as above.

That last example is what currently happens with Unit, even if it's been limited in the number of CPU's it's actually allowed to run on it will still create a number of router threads matching the result of the sysconf(3) call.

This pull-request thus comprises two patches to remedy this situation

The first adds a new auto check for the sched_getaffinity(2) system call on Linux.

The second uses this to determine the number of router threads to create, on Linux we still use sysconf(3) as a fallback.

This could also help with containers, whereby they are being limited in the number of CPU's they can use on high cpu count machines.

### DIFF
--- a/src/nxt_lib.c
+++ b/src/nxt_lib.c
@@ -32,7 +32,7 @@ const char *malloc_conf = "junk:true";
 nxt_int_t
 nxt_lib_start(const char *app, char **argv, char ***envp)
 {
-    int           n;
+    int           n = 0;
     nxt_int_t     flags;
     nxt_bool_t    update;
     nxt_thread_t  *thread;
@@ -87,12 +87,31 @@ nxt_lib_start(const char *app, char **argv, char ***envp)
 #ifdef _SC_NPROCESSORS_ONLN
     /* Linux, FreeBSD, Solaris, MacOSX. */
     n = sysconf(_SC_NPROCESSORS_ONLN);
+#endif
+
+#if (NXT_HAVE_LINUX_SCHED_GETAFFINITY)
+    if (n > 0) {
+        int        err;
+        size_t     size;
+        cpu_set_t  *set;
+
+        set = CPU_ALLOC(n);
+        if (set == NULL) {
+            return NXT_ERROR;
+        }
+
+        size = CPU_ALLOC_SIZE(n);
+
+        err = sched_getaffinity(0, size, set);
+        if (err == 0) {
+            n = CPU_COUNT_S(size, set);
+        }
+
+        CPU_FREE(set);
+    }
 
 #elif (NXT_HPUX)
     n = mpctl(MPC_GETNUMSPUS, NULL, NULL);
-
-#else
-    n = 0;
 
 #endif
 
--- a/auto/sched
+++ b/auto/sched
@@ -1,0 +1,19 @@
+# Copyright (C) Andrew Clayton
+# Copyright (C) NGINX, Inc.
+
+
+nxt_feature="Linux sched_getaffinity()"
+nxt_feature_name=NXT_HAVE_LINUX_SCHED_GETAFFINITY
+nxt_feature_run=no
+nxt_feature_incs=
+nxt_feature_libs=
+nxt_feature_test="#define _GNU_SOURCE
+                  #include <sched.h>
+
+                  int main(void) {
+                      cpu_set_t set;
+
+                      sched_getaffinity(0, sizeof(set), &set);
+                      return 0;
+                  }"
+. auto/feature
--- a/configure
+++ b/configure
@@ -135,6 +135,7 @@ fi
 . auto/cgroup
 . auto/isolation
 . auto/capability
+. auto/sched
 
 
 case "$NXT_SYSTEM_PLATFORM" in


ac000:
@hongzhidao Welcome back!

hongzhidao:
Hi @ac000,
It looks like a new feature but does not change the existing behavior.

> $ taskset -c 0-1 nproc 

It's possible to change the available CPU numbers in runtime, right?
If yes, I feel we can take account into adding the feature on runtime, I mean users can change it on the fly.
It seems it has anything to do with the feature you're doing. https://github.com/nginx/unit/pull/1385

ac000:
> Hi @ac000, It looks like a new feature but does not change the existing behavior.
> 
> > $ taskset -c 0-1 nproc
> 
> It's possible to change the available CPU numbers in runtime, right? If yes, I feel we can take account into adding the feature on runtime, I mean users can change it on the fly. It seems it has anything to do with the feature you're doing. #1385

Yes, you can change the processes CPU allowed mask at runtime.

However, doing automatic dynamic runtime reconfiguration is a much larger topic.

It may not be _too_ much work to hook it up so it gets run a reconfigure time, currently when doing a reconfigure

` nxt_router_conf_create()` is called which uses `rtcf->threads` or nxt_ncpu` to create the desired number of threads. However  `nxt_lib_start()` is not called which is where we determine the number of available CPUs.

It's probably not too much effort to factor out the CPU detection code into its own function and have it run from nxt_router_conf_create() or something...

But none of that really changes this patch which just makes the current situation better and would still be needed for doing the dynamic reconfiguration.

ac000:
> > But none of that really changes this patch which just makes the current situation better
> 
> Agree.
> 
> My thought is more conservative. I consider the following points.
> 
>     1. If in the future we implement this.
> 
> 
> > However, doing automatic dynamic runtime reconfiguration is a much larger topic.
> 
> It means it will be more powerful and can do the same thing as the current change.

It would still want this change though... so we do the right thing by default in more cases...

>2. Support the CPU affinity make it sense to extend the "threads" option in the "settings" object.

Possibly, or it might make sense for a different setting... However having the ability to specify cpu affinity within Unit is a separate unrelated change from this.

This patch isn't really about CPU affinity per se...

The traditional way to determine the number of available CPUs was with `_SC_NPROCESSORS_ONLN`.

On Linux sched_getaffinity(2) is a better solution as it _does_ take the per-process (well it's actually per-thread) CPU allowed masks into account.

There is also pthread_getaffinity_np(3) which does a similar thing and could perhaps be used on the BSDs to achieve the same. 

Say you have a 128 CPU machine and you have a number of docker instances running Unit, with each being limited to two CPUs via `--cpuset-cpus="2"`

What will currently happen is that each instance of unit will start 128 router threads, after this patch, each instance will start only 2.

> 3. It's not clear yet for me that what happens if the available CPU numbers is 0.

That would be an invalid setting

```
$ taskset 0x0 bash
taskset: failed to set pid 518501's affinity: Invalid argument
```

ac000:
Personally I think both this and #1385 are both separate valid changes with each stadning up on their own merit.

1) #1383 It's good to try and do the right thing by default where possible.
2) #1385 However that will not always be possible so let people fine tune things.

ac000:
Put it this way. at start up we create `nr router threads == nr runnable cpus`, however there are currently cases where this equation breaks down.

Currently we have the following situation

128 CPU box, Unit has the whole thing

```
nr router threads == nr runnable cpus

              128 == 128
```

128 CPU box, Unit is constrained to just 2 CPUs

```
nr router threads == nr runnable cpus

              128 == 2
```

There's the problem... forget about setting CPU affinities or things changing at runtime (that problem is the same with or without this patch).

hongzhidao:
> Personally I think both this and #1385 are both separate valid changes with each stadning up on their own merit.
> 
> 1. [Better available CPU count determination on Linux #1383](https://github.com/nginx/unit/pull/1383) It's good to try and do the right thing by default where possible.
> 2. [router: Make the number of router threads configurable #1385](https://github.com/nginx/unit/pull/1385) However that will not always be possible so let people fine tune things.

Makes sense.

src/nxt_lib.c
```diff
@@ -87,13 +87,31 @@ nxt_lib_start(const char *app, char **argv, char ***envp)
...
+    if (n > 0) {
+        int        err;
+        size_t     setsize;
+        cpu_set_t  *set;
+
+        set = CPU_ALLOC(n);
+        if (set == NULL) {
+            return NXT_ERROR;
+        }
+        setsize = CPU_ALLOC_SIZE(n);
```
hongzhidao, line 102: You could consider using `size` instead and adding a blank link.
Others look good.

ac000:
Couple of minor code-style tweaks

```
$ git range-diff 4b7e5136...446db7ac
1:  4b7e5136 ! 1:  446db7ac lib: Better available cpu count determination on Linux
    @@ src/nxt_lib.c: nxt_lib_start(const char *app, char **argv, char ***envp)
     +#if (NXT_HAVE_LINUX_SCHED_GETAFFINITY)
     +    if (n > 0) {
     +        int        err;
    -+        size_t     setsize;
    ++        size_t     size;
     +        cpu_set_t  *set;
     +
     +        set = CPU_ALLOC(n);
     +        if (set == NULL) {
     +        }
    -+        setsize = CPU_ALLOC_SIZE(n);
     +
    -+        err = sched_getaffinity(0, setsize, set);
    ++        size = CPU_ALLOC_SIZE(n);
    ++
    ++        err = sched_getaffinity(0, size, set);
     +        if (err == 0) {
    -+            n = CPU_COUNT_S(setsize, set);
    ++            n = CPU_COUNT_S(size, set);
     +        }
     +
     +        CPU_FREE(set);
```

src/nxt_lib.c
ac000: Sure.

hongzhidao (approved):

ac000:
Rebased with master

PR #1388, opened 2024-08-14
conf, router: Make the listen(2) backlog configurable

ac000:
@oopsoop2 on GitHub reported a performance issue related to the default listen(2) backlog size of 511 on nginx. They found that increasing it helped, nginx has a config option to configure this.

They would like to be able to do the same on Unit (which also defaults to 511 on some systems, incl Linux). This seems reasonable.

This adds a new per-listener 'backlog' config option, e.g

```
  {
      "listeners": {
          "[::1]:8080": {
              "pass": "routes",
              "backlog": 1024
          },
      }

      ...
  }
```

This doesn't effect the control socket.

### DIFF
--- a/src/nxt_router.c
+++ b/src/nxt_router.c
@@ -40,6 +40,7 @@ typedef struct {
 typedef struct {
     nxt_str_t         pass;
     nxt_str_t         application;
+    int               backlog;
 } nxt_router_listener_conf_t;
 
 
@@ -166,7 +167,7 @@ static void nxt_router_app_prefork_ready(nxt_task_t *task,
 static void nxt_router_app_prefork_error(nxt_task_t *task,
     nxt_port_recv_msg_t *msg, void *data);
 static nxt_socket_conf_t *nxt_router_socket_conf(nxt_task_t *task,
-    nxt_router_temp_conf_t *tmcf, nxt_str_t *name);
+    nxt_router_temp_conf_t *tmcf, nxt_str_t *name, int backlog);
 static nxt_int_t nxt_router_listen_socket_find(nxt_router_temp_conf_t *tmcf,
     nxt_socket_conf_t *nskcf, nxt_sockaddr_t *sa);
 
@@ -1494,6 +1495,12 @@ static nxt_conf_map_t  nxt_router_listener_conf[] = {
         NXT_CONF_MAP_STR_COPY,
         offsetof(nxt_router_listener_conf_t, application),
     },
+
+    {
+        nxt_string("backlog"),
+        NXT_CONF_MAP_INT32,
+        offsetof(nxt_router_listener_conf_t, backlog),
+    },
 };
 
 
@@ -1968,12 +1975,9 @@ nxt_router_conf_create(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
                 break;
             }
 
-            skcf = nxt_router_socket_conf(task, tmcf, &name);
-            if (skcf == NULL) {
-                goto fail;
-            }
-
             nxt_memzero(&lscf, sizeof(lscf));
+
+            lscf.backlog = -1;
 
             ret = nxt_conf_map_object(mp, listener, nxt_router_listener_conf,
                                       nxt_nitems(nxt_router_listener_conf),
@@ -1984,6 +1988,11 @@ nxt_router_conf_create(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
             }
 
             nxt_debug(task, "application: %V", &lscf.application);
+
+            skcf = nxt_router_socket_conf(task, tmcf, &name, lscf.backlog);
+            if (skcf == NULL) {
+                goto fail;
+            }
 
             // STUB, default values if http block is not defined.
             skcf->header_buffer_size = 2048;
@@ -2688,7 +2697,7 @@ nxt_router_application_init(nxt_router_conf_t *rtcf, nxt_str_t *name,
 
 static nxt_socket_conf_t *
 nxt_router_socket_conf(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
-    nxt_str_t *name)
+    nxt_str_t *name, int backlog)
 {
     size_t               size;
     nxt_int_t            ret;
@@ -2732,7 +2741,7 @@ nxt_router_socket_conf(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
         nxt_listen_socket_remote_size(ls);
 
         ls->socket = -1;
-        ls->backlog = NXT_LISTEN_BACKLOG;
+        ls->backlog = backlog > -1 ? backlog : NXT_LISTEN_BACKLOG;
         ls->flags = NXT_NONBLOCK;
         ls->read_after_accept = 1;
     }
@@ -2879,7 +2888,7 @@ nxt_router_listen_socket_ready(nxt_task_t *task, nxt_port_recv_msg_t *msg,
 
     nxt_socket_defer_accept(task, s, rpc->socket_conf->listen->sockaddr);
 
-    ret = nxt_listen_socket(task, s, NXT_LISTEN_BACKLOG);
+    ret = nxt_listen_socket(task, s, rpc->socket_conf->listen->backlog);
     if (nxt_slow_path(ret != NXT_OK)) {
         goto fail;
     }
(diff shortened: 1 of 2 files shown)


hongzhidao (commented):
I feel the prefix can be removed from the commit subject, "conf, router" is a bit strange.

src/nxt_router.c
```diff
@@ -1959,12 +1959,22 @@ nxt_router_conf_create(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
         next = 0;
 
         for ( ;; ) {
+            int               listen_backlog = -1;
+            nxt_conf_value_t  *backlog;
+
+            static const nxt_str_t  backlog_path = nxt_string("backlog");
```
hongzhidao, line 1965: Hi @ac000,
What about moving it into `nxt_router_listener_conf`?
For example:
```
diff --git a/src/nxt_router.c b/src/nxt_router.c
index 43209451..fe4bbabc 100644
--- a/src/nxt_router.c
+++ b/src/nxt_router.c
@@ -1964,13 +1964,10 @@ nxt_router_conf_create(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
                 break;
             }

-            skcf = nxt_router_socket_conf(task, tmcf, &name);
-            if (skcf == NULL) {
-                goto fail;
-            }
-
             nxt_memzero(&lscf, sizeof(lscf));

+            lscf.backlog = -1;
+
             ret = nxt_conf_map_object(mp, listener, nxt_router_listener_conf,
                                       nxt_nitems(nxt_router_listener_conf),
                                       &lscf);
@@ -1981,6 +1978,13 @@ nxt_router_conf_create(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,

             nxt_debug(task, "application: %V", &lscf.applicatio
n);

+            /* use lscf.backlog */
+
+            skcf = nxt_router_socket_conf(task, tmcf, &name);
+            if (skcf == NULL) {
+                goto fail;
+            }
+
             // STUB, default values if http block is not define
d.
```
I usually prefer to get object property by `nxt_conf_map_object()` rather than explicitly `nxt_conf_get_object_member()`.

Others look good.

ac000:
> I feel the prefix can be removed from the commit subject, "conf, router" is a bit strange.

It's simply because sometimes a change effects multiple areas, e.g

```
b5fe3eaf auto, wasm-wc: Copy the .so into build/lib/unit/modules/
8e254a4d tstr, conf: Ensure error strings are nul-terminated
5d1ce5c4 auto, perl: Fix building the Perl language module with clang
```

In this case it adds a new config option that effects the router. If I removed it, I would need to change the subject as well as this change _only_ effects the routers listen sockets...

src/nxt_router.c
ac000: I did try that originally but then it started getting messy... let me take another look...

ac000: OK, so it means

1) Adding a new 'backlog' entry to nxt_router_listener_conf[]

2) Adding a new 'backlog' member to nxt_router_listener_conf_t

2) Adding a pointer to nxt_router_listener_conf_t in nxt_router_temp_conf_t

So I took the simpler route, but sure, I can do the above instead...

src/nxt_router.c
```diff
@@ -1494,6 +1489,12 @@ static nxt_conf_map_t  nxt_router_listener_conf[] = {
         NXT_CONF_MAP_STR_COPY,
         offsetof(nxt_router_listener_conf_t, application),
     },
+
+    {
```
hongzhidao, line 1493: That's what I mean.

src/nxt_router.c
```diff
@@ -1981,6 +1979,13 @@ nxt_router_conf_create(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
 
             nxt_debug(task, "application: %V", &lscf.application);
 
+            tmcf->listener_conf = &lscf;
+
+            skcf = nxt_router_socket_conf(task, tmcf, &name);
```
hongzhidao, line 1984: It looks like your previous way is better, could you try it again?
```
nxt_router_socket_conf(task, tmcf, &name, lscf.backlog);
```
That we don't need to add `listener_conf` to `tmcf`.
I guess it's more concise.

ac000: Ah, a mixture of the original _and_ this way, yeah, that maybe makes sense.

ac000:
This basically just reverts the last change to make the next change much clearer...

ac000:
- Use `nxt_conf_map_object()` to get the backlog from the configuration and pass it directly into `nxt_router_socket_conf()`

```
$ git range-diff c6d5699c...4f99b044
1:  c6d5699c ! 1:  4f99b044 conf, router: Make the listen(2) backlog configurable
    @@ src/nxt_conf_validation.c: nxt_conf_vldt_forwarded(nxt_conf_validation_t *vldt,
          nxt_conf_value_t *value)
     
      ## src/nxt_router.c ##
    +@@ src/nxt_router.c: typedef struct {
    + typedef struct {
    +     nxt_str_t         pass;
    +     nxt_str_t         application;
    ++    int               backlog;
    + } nxt_router_listener_conf_t;
    + 
    + 
     @@ src/nxt_router.c: static void nxt_router_app_prefork_ready(nxt_task_t *task,
      static void nxt_router_app_prefork_error(nxt_task_t *task,
          nxt_port_recv_msg_t *msg, void *data);
    @@ src/nxt_router.c: static void nxt_router_app_prefork_ready(nxt_task_t *task,
      static nxt_int_t nxt_router_listen_socket_find(nxt_router_temp_conf_t *tmcf,
          nxt_socket_conf_t *nskcf, nxt_sockaddr_t *sa);
      
    -@@ src/nxt_router.c: nxt_router_conf_create(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
    -         next = 0;
    +@@ src/nxt_router.c: static nxt_conf_map_t  nxt_router_listener_conf[] = {
    +         NXT_CONF_MAP_STR_COPY,
    +         offsetof(nxt_router_listener_conf_t, application),
    +     },
    ++
    ++    {
    ++        nxt_string("backlog"),
    ++        NXT_CONF_MAP_INT32,
    ++        offsetof(nxt_router_listener_conf_t, backlog),
    ++    },
    + };
      
    -         for ( ;; ) {
    -+            int               listen_backlog = -1;
    -+            nxt_conf_value_t  *backlog;
    -+
    -+            static const nxt_str_t  backlog_path = nxt_string("backlog");
    -+
    -             listener = nxt_conf_next_object_member(listeners, &name, &next);
    -             if (listener == NULL) {
    + 
    +@@ src/nxt_router.c: nxt_router_conf_create(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
                      break;
                  }
      
     -            skcf = nxt_router_socket_conf(task, tmcf, &name);
    -+            backlog = nxt_conf_get_object_member(listener, &backlog_path, NULL);
    -+            if (backlog != NULL) {
    -+                listen_backlog = nxt_conf_get_number(backlog);
    +-            if (skcf == NULL) {
    +-                goto fail;
    +-            }
    +-
    +             nxt_memzero(&lscf, sizeof(lscf));
    + 
    ++            lscf.backlog = -1;
    ++
    +             ret = nxt_conf_map_object(mp, listener, nxt_router_listener_conf,
    +                                       nxt_nitems(nxt_router_listener_conf),
    +                                       &lscf);
    +@@ src/nxt_router.c: nxt_router_conf_create(nxt_task_t *task, nxt_router_temp_conf_t *tmcf,
    + 
    +             nxt_debug(task, "application: %V", &lscf.application);
    + 
    ++            skcf = nxt_router_socket_conf(task, tmcf, &name, lscf.backlog);
    ++            if (skcf == NULL) {
    ++                goto fail;
     +            }
     +
    -+            skcf = nxt_router_socket_conf(task, tmcf, &name, listen_backlog);
    -             if (skcf == NULL) {
    -                 goto fail;
    -             }
    +             // STUB, default values if http block is not defined.
    +             skcf->header_buffer_size = 2048;
    +             skcf->large_header_buffer_size = 8192;
     @@ src/nxt_router.c: nxt_router_application_init(nxt_router_conf_t *rtcf, nxt_str_t *name,
      
      static nxt_socket_conf_t *
```

hongzhidao (approved):

ac000:
Update the commit regarding the commit that changed the default listen backlog on Linux.

```
$ git range-diff 4f99b044...9c7c57ca
1:  4f99b044 ! 1:  9c7c57ca conf, router: Make the listen(2) backlog configurable
    @@ Commit message
         helped, nginx has a config option to configure this.
     
         They would like to be able to do the same on Unit (which also defaults
    -    to 511 on some systems, incl Linux). This seems reasonable.
    +    to 511 on some systems). This seems reasonable.
    +
    +    NOTE: On Linux before commit 97c15fa38 ("socket: Use a default listen
    +    backlog of -1 on Linux") we defaulted to 511. Since that commit we
    +    default to the Kernels default, which before 5.4 is 128 and after is
    +    4096.
     
         This adds a new per-listener 'backlog' config option, e.g
     
```

ac000:
Rebased with master

PR #1397, opened 2024-08-21
docs/openapi: Add new config options

ac000:
Add entries for the new 'backlog' and 'listen_threads' config options introduced in commits 57c88fd40 ("router: Make the number of router threads configurable") and 76489fb7e ("conf, router: Make the listen(2) backlog configurable").


avahahn (approved):
Looks great to me!

PR #1403, opened 2024-08-27
Remove .hgtags

ac000:
This is no longer needed since GitHub is our primary repository now.


thresheek:
LGTM

thresheek (approved):

thresheek:
On a second thought, maybe also remove .hgignore?  I cant imagine it being useful anymore too.

ac000:
Sure I can do that, the only reason I didn't is because I didn't know if people were still using mercurial but then using some hg->git bridge to push things... but yeah, probably not...

If anyone complains we can always add it back...

PR #1405, opened 2024-08-28
Added .gitignore for pkg/contrib/tarballs

thresheek:
This directory is needed for contribs to function.


thresheek:
We could go away with .gitkeep there, but let's mimic what we had with hgignores in this directory.

ac000:
Heh, yeah...

ac000 (approved):

PR #1415, opened 2024-09-06
Test fixes

ac000:
Two small fixes.

1) Fix an issue with routing tests when njs is not enabled

2) Suppress cargo-component output when we check for its existence

### DIFF
--- a/test/test_routing.py
+++ b/test/test_routing.py
@@ -2011,10 +2011,11 @@ def test_routes_match_destination_proxy():
     assert client.get()['status'] == 200, 'proxy'
 
 
-def test_routes_match_if():
+def set_if(condition):
+    assert 'success' in client.conf(f'"{condition}"', 'routes/0/match/if')
 
-    def set_if(condition):
-        assert 'success' in client.conf(f'"{condition}"', 'routes/0/match/if')
+
+def test_routes_match_if():
 
     def try_if(condition, status):
         set_if(condition)
@@ -2055,7 +2056,21 @@ def test_routes_match_if():
     assert client.get(url='/foo_empty?foo')['status'] == 200
     assert client.get(url='/foo?foo=1')['status'] == 404
 
-    # njs
+def test_routes_match_if_njs(require):
+    require({'modules': {'njs': 'any'}})
+
+    assert 'success' in client.conf(
+        {
+            "listeners": {"*:8080": {"pass": "routes"}},
+            "routes": [
+                {
+                    "match": {"method": "GET"},
+                    "action": {"return": 200},
+                }
+            ],
+            "applications": {},
+        }
+    )
 
     set_if('`${args.foo == \'1\'}`')
     assert client.get(url='/foo_1?foo=1')['status'] == 200
--- a/test/unit/applications/lang/wasm_component.py
+++ b/test/unit/applications/lang/wasm_component.py
@@ -11,7 +11,10 @@ class ApplicationWasmComponent(ApplicationProto):
     @staticmethod
     def prepare_env(script):
         try:
-            subprocess.check_output(['cargo', 'component', '--help'])
+            subprocess.check_output(
+                ['cargo', 'component', '--help'],
+                stderr=subprocess.STDOUT,
+            )
         except (subprocess.CalledProcessError, FileNotFoundError):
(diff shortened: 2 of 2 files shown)


hongzhidao (approved):

test/test_routing.py
```diff
@@ -2057,6 +2057,9 @@ def try_if(condition, status):
 
     # njs
 
+    if not option.available['modules']['njs']:
```
andrey-zelenkov, line 2060: `require({'modules': {'njs': 'any'}})` should mark test as skipped rather than terminating silently

ac000: Doesn't that make it skip the whole test?

Maybe we need to split that test up...

ac000:
Split njs test out into its own function so we now get

```
collected 115 items / 113 deselected / 2 selected                              

test/test_routing.py::test_routes_match_if PASSED
test/test_routing.py::test_routes_match_if_njs SKIPPED (Unit has no njs
module(s))

================= 1 passed, 1 skipped, 113 deselected in 0.71s =================
```

```
$ git range-diff 63fd0c4f...2a06b4bb
1:  fae7b575 < -:  -------- tests: Fix routing tests in the no njs case
-:  -------- > 1:  8e6c4041 tests: Fix routing tests in the no njs case
2:  63fd0c4f = 2:  2a06b4bb tests: Suppress cargo-component output
andrew@kappa unit:test-fixes $ git range-diff --creation-factor=100 63fd0c4f...2a06b4bb
1:  fae7b575 ! 1:  8e6c4041 tests: Fix routing tests in the no njs case
    @@ Commit message
     
      ## test/test_routing.py ##
    -@@ test/test_routing.py: def test_routes_match_if():
    +@@ test/test_routing.py: def test_routes_match_destination_proxy():
    +     assert client.get()['status'] == 200, 'proxy'
      
    -     # njs
      
    -+    if not option.available['modules']['njs']:
    -+        return
    +-def test_routes_match_if():
    ++def set_if(condition):
    ++    assert 'success' in client.conf(f'"{condition}"', 'routes/0/match/if')
    + 
    +-    def set_if(condition):
    +-        assert 'success' in client.conf(f'"{condition}"', 'routes/0/match/if')
     +
    ++def test_routes_match_if():
    + 
    +     def try_if(condition, status):
    +         set_if(condition)
    +@@ test/test_routing.py: def test_routes_match_if():
    +     assert client.get(url='/foo_empty?foo')['status'] == 200
    +     assert client.get(url='/foo?foo=1')['status'] == 404
    + 
    +-    # njs
    ++def test_routes_match_if_njs(require):
    ++    require({'modules': {'njs': 'any'}})
    ++
    ++    assert 'success' in client.conf(
    ++        {
    ++            "listeners": {"*:8080": {"pass": "routes"}},
    ++            "routes": [
    ++                {
    ++                    "match": {"method": "GET"},
    ++                    "action": {"return": 200},
    ++                }
    ++            ],
    ++            "applications": {},
    ++        }
    ++    )
    + 
          set_if('`${args.foo == \'1\'}`')
          assert client.get(url='/foo_1?foo=1')['status'] == 200
    -     assert client.get(url='/foo_2?foo=2')['status'] == 404
2:  63fd0c4f = 2:  2a06b4bb tests: Suppress cargo-component output
```

test/test_routing.py
andrey-zelenkov: > Doesn't that make it skip the whole test?

No, only the tests after the `require()` call are skipped, but the entire test suite will be marked as skipped.

> Maybe we need to split that test up...

Ideally, yes. It would make the tests more explicit.

ac000: Which is what I did, so it looks good to go...

ac000:
Rebased with master.

ac000:
Rebased with master

```
$ git range-diff 53d726a1...cff5e092
-:  -------- > 1:  c5846ba3 ci: Fix wasmtime paths in ci.yml
-:  -------- > 2:  46ddb010 ci: Trigger ci.yml for changes under pkg/contrib
1:  22125366 = 3:  6976a614 tests: Fix routing tests in the no njs case
2:  53d726a1 = 4:  cff5e092 tests: Suppress cargo-component output
```

Issue #828, opened 2023-01-17
unitd router process crash whilst using proxy


mattxtaz:
I can make the unitd router process crash with a signal 11 on FreeBSD by proxying to an NGINX webserver, downloading a large file through it using curl, and interrupting it with control-C a few times. Unit is listening with TLS and I'm matching a uri and proxying out to 127.0.0.1.

If I change "proxy" to "share" and share it locally instead then I can't get it to crash. Without pressing control-c and leaving it to complete, I also can't get it to crash.

Is there anything I can do to try and diagnose this further?

**On the client:**
% curl --output /dev/null https://example.com/temp/Bank.mp4
**-wait a few seconds and then press control-c-**
% curl --output /dev/null https://example.com/temp/Bank.mp4
**-wait a few seconds and then press control-c-**
% curl --output /dev/null https://example.com/temp/Bank.mp4
curl: (56) OpenSSL SSL_read: error:0A000126:SSL routines::unexpected eof while reading, errno 0

**On the server:
-this is logged when i press control-c-**
2023/01/17 17:01:41 [info] 38055#229573 *70 SSL_write(51, 8072AD848, 4096) failed (32: Broken pipe)
2023/01/17 17:01:41 [alert] 38055#229573 kevent(33) error (9: Bad file descriptor) on ident:-1 filter:-1        
**-this is logged when i press control-c> -**                                    
2023/01/17 17:01:51 [info] 38055#229575 *69 SSL_write(51, 8034D1448, 4096) failed (32: Broken pipe)
2023/01/17 17:01:51 [alert] 38055#229575 kevent(35) error (9: Bad file descriptor) on ident:-1 filter:-1
**-this is logged when i press control-c and then it crashes-**
2023/01/17 17:02:04 [info] 38055#229574 *67 SSL_write(50, 807484C48, 4096) failed (32: Broken pipe)     
2023/01/17 17:02:04 [alert] 38055#229574 kevent(34) error (9: Bad file descriptor) on ident:-1 filter:-1
2023/01/17 17:02:05 [alert] 33800#100289 process 38055 exited on signal 11                                                                            
2023/01/17 17:02:05 [alert] 33800#100289 sendmsg(14, -1, -1, 1) failed (32: Broken pipe)
2023/01/17 17:02:05 [alert] 33800#100289 sendmsg(17, -1, -1, 1) failed (32: Broken pipe)
[...]

PR #1418, opened 2024-09-09
http: Fix router process crash whilst using proxy

hongzhidao:
When the client closes the connection before the upstream, the proxy's error handler was calling cleanup operation like peer close and requestclose twice, this fix ensures the cleanup is performed only once, improving proxy stability.

Closes: https://github.com/nginx/unit/issues/828

### DIFF
--- a/src/nxt_h1proto.c
+++ b/src/nxt_h1proto.c
@@ -2869,6 +2869,11 @@ nxt_h1p_peer_body_process(nxt_task_t *task, nxt_http_peer_t *peer,
     } else if (h1p->remainder > 0) {
         length = nxt_buf_chain_length(out);
         h1p->remainder -= length;
+
+        if (h1p->remainder == 0) {
+            nxt_buf_chain_add(&out, nxt_http_buf_last(peer->request));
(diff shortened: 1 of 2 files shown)


ac000 (commented):
Hi @hongzhidao 

Good fix!

```
http: Fix router process crash whilst using proxy

When the client closes the connection before the upstream,
the proxy's error handler was calling cleanup operation like
peer close and requestclose twice, this fix ensures the cleanup
is performed only once, improving proxy stability.
```

Maybe requestclose should be `request close`?

hongzhidao:
> Hi @hongzhidao
> 
> Good fix!
> 
> ```
> http: Fix router process crash whilst using proxy
> 
> When the client closes the connection before the upstream,
> the proxy's error handler was calling cleanup operation like
> peer close and requestclose twice, this fix ensures the cleanup
> is performed only once, improving proxy stability.
> ```
> 
> Maybe requestclose should be `request close`?

Fixed, thanks.

ac000 (approved):
LGTM.

Issue #1439, opened 2024-09-23
Unitctl: openssl linking issues


avahahn:
Unitctl fails to open when the expected libssl version is not installed on the host system.
It seems on aarch64 systems libssl 1 is expected and on x86 systems libssl 3 is expected.

An example from Ubuntu on aarch64
```
$ unitctl
unitctl: error while loading shared libraries: libssl.so.1.0.0: cannot open shared object file: No such file or directory
```

Ideally unitctl should work with a wide berth of openssl/libssl versions on a variety of platforms.

ac000:
Seems to be an artefact of cross compiling on Ubuntu...


```
[target.aarch64-unknown-linux-gnu]
pre-build = [
    "dpkg --add-architecture $CROSS_DEB_ARCH",
    "apt-get update && apt-get install --assume-yes libssl-dev:$CROSS_DEB_ARCH"
]
```

```
#5 [2/2] RUN eval "dpkg --add-architecture $CROSS_DEB_ARCH
apt-get update && apt-get install --assume-yes libssl-dev:$CROSS_DEB_ARCH"
...
#5 6.335 The following additional packages will be installed:
#5 6.335   gcc-6-base:arm64 libc6:arm64 libc6-dev:arm64 libgcc1:arm64 libssl1.0.0:arm64
...
#5 6.337 The following NEW packages will be installed:
#5 6.338   gcc-6-base:arm64 libc6:arm64 libc6-dev:arm64 libgcc1:arm64 libssl-dev:arm64
#5 6.338   libssl1.0.0:arm64 linux-libc-dev:arm64 zlib1g:arm64 zlib1g-dev:arm64
...
#5 7.507 Get:5 http://ports.ubuntu.com/ubuntu-ports xenial-updates/main arm64 libssl1.0.0 arm64 1.0.2g-1ubuntu4.20 [693 kB]
...
#5 8.136 Preparing to unpack .../libssl1.0.0_1.0.2g-1ubuntu4.20_arm64.deb ...
#5 8.138 Unpacking libssl1.0.0:arm64 (1.0.2g-1ubuntu4.20) ...
...
#5 8.498 Selecting previously unselected package libssl-dev:arm64.
#5 8.499 Preparing to unpack .../libssl-dev_1.0.2g-1ubuntu4.20_arm64.deb ...
...
#5 9.719 Setting up libssl1.0.0:arm64 (1.0.2g-1ubuntu4.20) ...
...
[...]

Issue #1438, opened 2024-09-23
Error in the hyper legacy client: client error (Connect)


zwn:
When using `unitctl` I always see this message: `Error in the hyper legacy client: client error (Connect)`
Afterwards everything seems to be working.

I am working with version 1.33 of both `unitd` and `unitctl` on Debian 12. Default install, no changes.

avahahn:
Hello there! 
We just merged a fix for this. If you are interested in trying a nightly build fresh from our CI pipelines to verify that our fix is working in your environment I can link you to one here:
https://github.com/nginx/unit/actions/runs/11019141121/artifacts/1973044785

I stress that this is not a supported release

avahahn:
Closing issue for now, but if you continue to experience this once you are running the new code feel free to reopen, thank you!

zwn:
> Hello there! We just merged a fix for this. If you are interested in trying a nightly build fresh from our CI pipelines to verify that our fix is working in your environment I can link you to one here: https://github.com/nginx/unit/actions/runs/11019141121/artifacts/1973044785

Thanks for the fix. I tested the linked artifact but it is straight 1.33.0 release. I tried to find the correct link, but the [build job](https://github.com/nginx/unit/actions/runs/11019141121/job/30600998116) contains the same link so I am lost.

ac000:
@zwn 

Are you sure?

```
$ ls -l unitctl-*
-rwxr-xr-x 1 andrew andrew 12214560 Sep 17 02:26 unitctl-1.33.0-x86_64-unknown-linux-gnu
-rwxr-xr-x 1 andrew andrew 15447840 Sep 24 17:48 unitctl-merge-x86_64-unknown-linux-gnu
```

The second binary is from the above link.

Let's take a closer look

```console
$ readelf -a unitctl-1.33.0-x86_64-unknown-linux-gnu | grep NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libssl.so.3]
 0x0000000000000001 (NEEDED)             Shared library: [libcrypto.so.3]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]
[...]

PR #1440, opened 2024-09-23
unitctl: fixes for openssl and docker client

avahahn:
This PR contains commits to fix both https://github.com/nginx/unit/issues/1439 and https://github.com/nginx/unit/issues/1438.

The OpenSSL issue is solved by moving to hyper-rustls from hyper-tls. 

The docker client error (hyper legacy client connect error) is solved by bumping the bollard library version and also just in case I have elaborated on the error print text.

Both issues were tested on Debian latest docker image (x86_64) as well as M1 (aarch64) OSX.
Help verifying on an aarch64 Linux box is welcome, but I expect similar results.

### DIFF
--- a/tools/unitctl/unit-client-rs/src/unit_client.rs
+++ b/tools/unitctl/unit-client-rs/src/unit_client.rs
@@ -191,7 +191,7 @@ impl UnitClient {
                    .unwrap_or_else(|_| HttpsConnectorBuilder::default()
                                    .with_webpki_roots())
                    .https_or_http()
-                   .enable_http1()
+                   .enable_all_versions()
                    .wrap_connector(HttpConnector::new()));
         Self {
             control_socket,
--- a/tools/unitctl/unit-client-rs/src/unit_client.rs
+++ b/tools/unitctl/unit-client-rs/src/unit_client.rs
@@ -9,7 +9,7 @@ use hyper::body::{Buf, HttpBody};
 use hyper::client::{HttpConnector, ResponseFuture};
 use hyper::Error as HyperError;
 use hyper::{http, Body, Client, Request};
-use hyper_tls::HttpsConnector;
+use hyper_rustls::{HttpsConnectorBuilder, HttpsConnector};
 use hyperlocal::{UnixClientExt, UnixConnector};
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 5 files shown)


thresheek:
Does the move to hyper-rustls mean that the resulting binary will not be linked to openssl anymore?

avahahn:
> Does the move to hyper-rustls mean that the resulting binary will not be linked to openssl anymore?

Yes. I verified this with ldd on x86_64.

tools/unitctl/unit-client-rs/src/unit_client.rs
```diff
@@ -185,7 +185,14 @@ impl UnitClient {
...
 
     pub fn new_http(control_socket: ControlSocket) -> Self {
-        let remote_client = Client::builder().build(HttpsConnector::new());
+        let remote_client = Client::builder()
+            .build(HttpsConnectorBuilder::default()
+                   .with_native_roots()
+                   .unwrap_or_else(|_| HttpsConnectorBuilder::default()
+                                   .with_webpki_roots())
+                   .https_or_http()
+                   .enable_http1()
```
javorszky, line 194: not a blocker, more of a curiosity question: is [http2 something we might want to enable](https://docs.rs/hyper-rustls/latest/hyper_rustls/struct.HttpsConnectorBuilder.html#method.enable_http2-1) here as well?

tools/unitctl/unit-client-rs/src/unit_client.rs
```diff
@@ -185,7 +185,14 @@ impl UnitClient {
...
+                   .wrap_connector(HttpConnector::new()));
```
javorszky, line 195: this one apparently is only available with the http2 crate feature: https://docs.rs/hyper-rustls/latest/hyper_rustls/struct.HttpsConnectorBuilder.html#method.wrap_connector-1

The http2 flag isn't enabled by default per https://docs.rs/crate/hyper-rustls/latest/features, and it's also not enabled in the cargo.toml file 🤔 

How does this work? 😄 It built completely fine, but docs say it shouldn't have...?

(this is a rust thing, not a you thing)

javorszky (approved):
Left 2 questions, more of a curiosity than anything else. The code looks good, implementation makes sense, builds locally on aarch64, lgtm!

tools/unitctl/unit-client-rs/src/unit_client.rs
javorszky: Oh, there are two `wrap_connector` methods, one is on the `ConnectorBuilder<WantsProcols2>` impl, the other is on the `ConnectorBuilder<WantsProtocols3>` impl.

The second one needs the http2 crate, the first one does not.

tools/unitctl/unit-client-rs/src/unit_client.rs
avahahn: The only case where that would be used that I can imagine is in the case of a load balancer that is set up to serve HTTP2 to clients, but to relay HTTP1 back to the Unit control API. 

Since its easy and we dont have to support the logic directly Ill enable it, but it is a funky little corner case.

javorszky (approved):

Issue #1458, opened 2024-10-14
Access log entries lack newline characters when using JS configuration.


hongzhidao:
When using `format` in `access_log` with JS, each log doesn't end with `\n`.
For example:
```
{
        "listeners":
                {"*:80":
                        {"pass":"routes"}
                },
        "routes":[
                {"action":{"return":204}}
        ],
        "access_log": {
                "path": "/tmp/access.log",
                "format": "`test`"
        }
}
```
But it works well with non-JS `format` option.

PR #1459, opened 2024-10-14
Fix missing newlines in access logs for JS configuration

hongzhidao:
When using JS configuration for the "format" option, access log entries
    were being written without newline characters. This commit adds the
    missing newline character to each log entry.

    Closes: https://github.com/nginx/unit/issues/1458

### DIFF
--- a/src/nxt_js.c
+++ b/src/nxt_js.c
@@ -230,7 +230,7 @@ nxt_js_add_module(nxt_js_conf_t *jcf, nxt_str_t *name, nxt_str_t *text)
 
 
 nxt_js_t *
-nxt_js_add_tpl(nxt_js_conf_t *jcf, nxt_str_t *str, nxt_bool_t strz)
+nxt_js_add_tpl(nxt_js_conf_t *jcf, nxt_str_t *str, nxt_uint_t flags)
 {
     size_t     size;
     u_char     *p, *start;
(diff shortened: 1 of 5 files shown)


ac000 (commented):
```
Fix missing newlines in access logs for JS configuration

When using JS configuration for the "format" option, access log entries
were being written without newline characters. This commit adds the
missing newline character to each log entry.

Closes: <https://github.com/nginx/unit/issues/1458>
```

Remember, until GH fixes their parsing we should not enclose `Closes:` URLs in `<>`.

src/nxt_js.c
```diff
@@ -243,13 +243,19 @@ nxt_js_add_tpl(nxt_js_conf_t *jcf, nxt_str_t *str, nxt_bool_t strz)
                                            "    return ");
 
     /*
+     * Appending a newline character if newline is true.
```
ac000, line 246: I would rather s/Appending/Append/, because... English... but then this matches the other one...

src/nxt_tstr.c
```diff
@@ -80,16 +80,17 @@ nxt_tstr_compile(nxt_tstr_state_t *state, const nxt_str_t *str,
 {
     u_char      *p;
     nxt_tstr_t  *tstr;
-    nxt_bool_t  strz;
+    nxt_bool_t  strz, newline;
 
     strz = (flags & NXT_TSTR_STRZ) != 0;
+    newline = (flags & NXT_TSTR_NEWLINE) != 0;
```
ac000, line 86: I don't know why we do the test like this (!= 0) it can _only_ be either set or not.

We don't do it like that in the if () statements...

src/nxt_js.c
hongzhidao: Good to know.

src/nxt_tstr.c
hongzhidao: Makes sense.

ac000:
Looks like the tests will need tweaking <https://github.com/nginx/unit/actions/runs/11337821193/job/31530024356?pr=1459>.

src/nxt_tstr.c
ac000: Yes, indeed... sorry brain fart on my part...

hongzhidao: No worries, tests can help in this case.

ac000 (approved):
Yeah, look good, tests are happy now ;) (apart from wasm-wasi-component which I've noticed has started failing recently, but that's a separate issue).

PR #1461, opened 2024-10-15
ci: Various fixes for update to Ubuntu 24.04

ac000:
This pull-request compreises three patches fixing various issues with the CI due to the runners moving to Ubuntu 24.04 (though they seem to have sinnce reverted back, but we may as well keep these fixes...)

The first patch fixes the disabling of the mono service which may or may not be installed.

The second patch fixes the installation of pytest by installing it via the apt(8) package manager.

The third drops PHP 8.1 from our tests. See the commit message for the gory details.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,12 +101,18 @@ jobs:
       # https://github.com/actions/runner-images/issues/2821
       - name: Kill mono process
         run: |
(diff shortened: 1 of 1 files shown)


javorszky (approved):
Reviewed all separate commits. I really appreciate the explanation on the php issue, that was an exhaustive reasoning!

This patch is g2g from my end.

PR #1479, opened 2024-10-25
http: JSON format access logging support

hongzhidao:
Allow format to be an object to generate JSON logs. The object keys
become JSON field names, and values support string, variable, and JS.

Note that when there is no JS in the format values, the object will
be pre-serialized to a JSON template string at configuration phase
for better performance.

Example config:
```
  {
    "access_log": {
      "path": "/tmp/access.log",
      "format": {
        "remote_addr": "$remote_addr",
        "time_local": "$time_local",
        "request_line": "$request_line",
        "status": "$status",
        "body_bytes_sent": "$body_bytes_sent",
        "header_referer": "$header_referer",
        "header_user_agent": "$header_user_agent"
      }
    }
  }
```

### DIFF
--- a/src/nxt_conf_validation.c
+++ b/src/nxt_conf_validation.c
@@ -216,6 +216,11 @@ static nxt_int_t nxt_conf_vldt_server_weight(nxt_conf_validation_t *vldt,
     nxt_conf_value_t *value, void *data);
 static nxt_int_t nxt_conf_vldt_access_log(nxt_conf_validation_t *vldt,
     nxt_conf_value_t *value, void *data);
+static nxt_int_t nxt_conf_vldt_access_log_format(nxt_conf_validation_t *vldt,
+    nxt_conf_value_t *value, void *data);
+static nxt_int_t nxt_conf_vldt_access_log_format_field(
+    nxt_conf_validation_t *vldt, const nxt_str_t *name,
+    nxt_conf_value_t *value);
 
 static nxt_int_t nxt_conf_vldt_isolation(nxt_conf_validation_t *vldt,
     nxt_conf_value_t *value, void *data);
@@ -1465,7 +1470,8 @@ static nxt_conf_vldt_object_t  nxt_conf_vldt_access_log_members[] = {
         .type       = NXT_CONF_VLDT_STRING,
     }, {
         .name       = nxt_string("format"),
-        .type       = NXT_CONF_VLDT_STRING,
+        .type       = NXT_CONF_VLDT_STRING | NXT_CONF_VLDT_OBJECT,
+        .validator  = nxt_conf_vldt_access_log_format,
     }, {
         .name       = nxt_string("if"),
         .type       = NXT_CONF_VLDT_STRING,
@@ -3620,6 +3626,49 @@ nxt_conf_vldt_access_log(nxt_conf_validation_t *vldt, nxt_conf_value_t *value,
 
     if (nxt_is_tstr(&conf.format)) {
         return nxt_conf_vldt_var(vldt, &format_str, &conf.format);
+    }
+
+    return NXT_OK;
+}
+
+
+static nxt_int_t
+nxt_conf_vldt_access_log_format(nxt_conf_validation_t *vldt,
+    nxt_conf_value_t *value, void *data)
+{
+    static const nxt_str_t  format = nxt_string("format");
+
+    if (nxt_conf_type(value) == NXT_CONF_OBJECT) {
+        return nxt_conf_vldt_object_iterator(vldt, value,
+                                         nxt_conf_vldt_access_log_format_field);
+    }
+
+    /* NXT_CONF_STRING */
+
+    return nxt_conf_vldt_access_log_format_field(vldt, &format, value);
+}
+
+
+static nxt_int_t
+nxt_conf_vldt_access_log_format_field(nxt_conf_validation_t *vldt,
+    const nxt_str_t *name, nxt_conf_value_t *value)
+{
+    nxt_str_t  str;
+
+    if (name->length == 0) {
+        return nxt_conf_vldt_error(vldt, "In the access log format, the name "
+                                         "must not be empty.");
+    }
+
+    if (nxt_conf_type(value) != NXT_CONF_STRING) {
+        return nxt_conf_vldt_error(vldt, "In the access log format, the value "
+                                         "must be a string.");
+    }
+
+    nxt_conf_get_string(value, &str);
+
+    if (nxt_is_tstr(&str)) {
+        return nxt_conf_vldt_var(vldt, name, &str);
     }
 
     return NXT_OK;
--- a/src/nxt_router_access_log.c
+++ b/src/nxt_router_access_log.c
@@ -23,9 +23,16 @@ typedef struct {
 } nxt_router_access_log_ctx_t;
 
 
+struct nxt_router_access_log_format_s {
+    nxt_tstr_t                *tstr;
+};
+
+
+static nxt_router_access_log_format_t *nxt_router_access_log_format_create(
+    nxt_task_t *task, nxt_router_conf_t *rtcf, nxt_conf_value_t *value);
 static void nxt_router_access_log_writer(nxt_task_t *task,
     nxt_http_request_t *r, nxt_router_access_log_t *access_log,
-    nxt_tstr_t *format);
+    nxt_router_access_log_format_t *format);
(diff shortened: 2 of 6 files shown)


src/nxt_conf_validation.c
```diff
@@ -3523,3 +3529,46 @@ nxt_conf_vldt_access_log(nxt_conf_validation_t *vldt, nxt_conf_value_t *value,
...
+    nxt_str_t  str;
+
+    if (name->length == 0) {
+        return nxt_conf_vldt_error(vldt, "In the access log format, the name "
+                                         "must not be empty.");
+    }
+
+    if (nxt_conf_type(value) != NXT_CONF_STRING) {
+        return nxt_conf_vldt_error(vldt, "In the access log format, the value "
+                                         "must be only string.");
```
ac000, line 3564: Maybe that could be re-worded slightly to "must be a string."

src/nxt_router_access_log.c
```diff
@@ -190,25 +274,105 @@ nxt_router_access_log_writer(nxt_task_t *task, nxt_http_request_t *r,
...
+nxt_router_access_log_text(nxt_task_t *task, nxt_http_request_t *r,
+    nxt_router_access_log_ctx_t *ctx, nxt_tstr_t *tstr)
+{
+    nxt_int_t          ret;
+    nxt_router_conf_t  *rtcf;
+
+    if (nxt_tstr_is_const(tstr)) {
+        nxt_tstr_str(tstr, &ctx->text);
 
     } else {
```
ac000, line 300: Yep, Mr Reduce Scope again!

```c
} else {
    nxt_int_t          ret;
    nxt_router_conf_t  *rtcf;
```

src/nxt_router_access_log.c
```diff
@@ -190,25 +274,105 @@ nxt_router_access_log_writer(nxt_task_t *task, nxt_http_request_t *r,
...
+    nxt_router_access_log_member_t  *member;
+
+    rtcf = r->conf->socket_conf->router_conf;
+
+    value = nxt_conf_create_object(r->mem_pool, format->nmembers);
+    if (nxt_slow_path(value == NULL)) {
+        return NXT_ERROR;
+    }
+
+    for (i = 0; i < format->nmembers; i++) {
```
ac000, line 339: `i` could be for loop initialised...

src/nxt_router_access_log.c
```diff
@@ -190,25 +274,105 @@ nxt_router_access_log_writer(nxt_task_t *task, nxt_http_request_t *r,
...
+        member = &format->member[i];
+
+        if (nxt_tstr_is_const(member->tstr)) {
+            nxt_tstr_str(member->tstr, &str);
+
+        } else {
```
ac000, line 345: Looks like `ret` could be declared here...

ac000: `member` could be declared in the for loop...

hongzhidao: Good to know, thanks for the suggestions.
I'm fine with both styles. And I'm used to declaring variables at the beginning of functions, so I'd like to keep this way.

ac000: Well, it is more than about style, it's good programming [practice](https://wiki.sei.cmu.edu/confluence/display/c/DCL19-C.+Minimize+the+scope+of+variables+and+functions), which is why I won't stop suggesting it ;)

Note: declaring variables at start of scope is _not_ the same as declaring variables randomly throughout the function, which I certainly _wouldn't_ suggest to do!

hongzhidao: I think both ways have pros and cons. Declaring at the start makes all variables visible at a glance and makes reuse easier.

Take this function as an example:
```
static nxt_int_t
nxt_router_access_log_json(nxt_task_t *task, nxt_http_request_t *r,
    nxt_router_access_log_ctx_t *ctx, nxt_router_access_log_format_t *format)
{
    nxt_uint_t                      i;
    nxt_conf_value_t                *value;
    nxt_router_access_log_member_t  *member;

    value = nxt_conf_create_object(r->mem_pool, format->nmembers);
    if (nxt_slow_path(value == NULL)) {
        return NXT_ERROR;
    }

    for (i = 0; i < format->nmembers; i++) {
        member = &format->member[i];

        ...

        nxt_conf_set_member_string(value, &member->name, &str, i);
    }

    ...

    return NXT_OK;
}
```
I feel it's common to write like it. If we asked someone to totally follow something like the above "practice", it looks like that:
```
static nxt_int_t
nxt_router_access_log_json(nxt_task_t *task, nxt_http_request_t *r,
    nxt_router_access_log_ctx_t *ctx, nxt_router_access_log_format_t *format)
{
    nxt_conf_value_t  *value;

    value = nxt_conf_create_object(r->mem_pool, format->nmembers);
    if (nxt_slow_path(value == NULL)) {
        return NXT_ERROR;
    }

    for (nxt_uint_t i = 0; i < format->nmembers; i++) {
        nxt_router_access_log_member_t  *member;

        member = &format->member[i];

        ...

        nxt_conf_set_member_string(value, &member->name, &str, i);
    }

    ...

    return NXT_OK;
}
```
From the point of code style, both ways look fine.
It doesn't seem to make a difference in terms of whether it's more buggy or not.

ac000: For me the second version is more concise.

It also has the advantage that suppose you _mistakenly_ tried to do something with `member` after the for loop, the compiler will shout at you... unlike the first version where all kinds of mayhem may ensue...

So I will continue pushing it (even if I do sound like a broken record) as in the current climate anything that helps write _safer_ code is a _good_ thing...

hongzhidao: Thanks for caring about code safety. I'm comfortable with my current style of declaring at the beginning. Both ways can be safe when written carefully, especially with tests, so I'd like to keep my approach.

ac000: > Thanks for caring about code safety. I'm comfortable with my current style of declaring at the beginning. Both ways can be safe when written carefully

Careful you don't fall into the fallacy that you won't make mistakes no matter how careful you are.

I don't care _who_ you are, if you say you _never_ write buggy code, then you are either outright lying or delusional!

> , especially with tests, so I'd like to keep my approach.

Tests are good but not infallible...

Your choice, I'll just keep bugging you ;) (though I won't bring it up again in this PR)

hongzhidao: > I don't care who you are, if you say you never write buggy code, then you are either outright lying or delusional!

I discussed it based on the specific code which I listed above. Please do not expand the scope and apply one "principle" to all situations.
I didn't say I never write buggy code. I mean the way I used in `nxt_router_access_log_json()` didn't introduce issues, let me know if it's possible.

Furthur example:
```
int array_sum(int a[], int n)
{
    int i, sum;

    sum = 0;

    for (i = 0; i < n; i++) {
        sum += a[i];
    }

    return sum;
}
```
I don't think it should be like that, but I'm fine if anyone likes the way.
```
int array_sum(int a[], int n)
{
   int sum;

    sum = 0;

    for (int i = 0; i < n; i++) {
        sum += a[i];
    }

    return sum;
}
```

ac000: > I discussed it based on the specific code which I listed above. Please do not expand the scope and apply one "principle" to all situations.

I _am_ talking generally (not specific examples)...

Reducing variable scope is just good programming practice... ipso facto...

hongzhidao: Take `nxt_router_access_log_json()` as an example, `i` and `member` are only used in the `for` sentence, I can't see benefits to force their declaration within it.
I would be confused if it can introduce issues in this way, I'm only talking about the function since we are reviewing it.

ac000: > Take `nxt_router_access_log_json()` as an example, `i` and `member` are only used in the `for` sentence,

Yes, that's exactly the point! They are _only_ used in the for loop so that is where they should be _scoped_.

> I can't see benefits to force their declaration within it.

I already gave an example of a benefit...

> I would be confused if it can introduce issues in this way, I'm only talking about the function since we are reviewing it.

Never underestimate the propensity for introducing bugs.

hongzhidao: I wonder if a function is as small as array_sum(), should it follow the practice?

ac000: Certainly no harm... and it's more concise to read...

ac000 (approved):
Assuming this doesn't change wildly...

PR #1489, opened 2024-11-13
Decast nxt_cpymem()

ac000:
```
    Decast nxt_cpymem()
    
    nxt_cpymem() is basically mempcpy(3)
    
    Like mempcpy() nxt_cpymem() returns a void *.
    
    nxt_cpymem() is implemented as a wrapper around memcpy(3), however
    before returning the new pointer value we cast the return of memcpy(3)
    to a u_char *, then add the length parameter to it.
    
    I guess this was done to support compilers that do not support
    arithmetic on void pointers as the C standard forbids it.
    
    However since we removed support for compilers other than GCC and Clang
    (ending in commit 9cd11133 ("Remove support for Sun's Sun Studio/SunPro
    C compiler")) this is no longer an issue as both GCC and Clang support
    arithmetic on void pointers (without the -pedantic option).
    
    While removing the unnecessary casting in this case doesn't necessarily
    improve type-safety (as we're dealing with void *'s in and out), it does
    just make the code that little more readable.
    
    Oh and for interest we have actually already been relying on this
    extension
    
      src/nxt_array.c:143:40: warning: arithmetic on a pointer to void is a GNU extension [-Wgnu-pointer-arith]
        143 |             nxt_memcpy(data, src->elts + (i * size), size);
            |                              ~~~~~~~~~ ^
      src/nxt_string.h:45:24: note: expanded from macro 'nxt_memcpy'
         45 |     (void) memcpy(dst, src, length)
            |                        ^~~
    
    which was introduced in e2b53e16 ("Added "rootfs" feature.") back in
    2020.
    
    Link: <https://gcc.gnu.org/onlinedocs/gcc/Pointer-Arith.html>
    Link: <https://clang.llvm.org/docs/LanguageExtensions.html#introduction>
```

### DIFF
--- a/src/nxt_string.h
+++ b/src/nxt_string.h
@@ -58,7 +58,7 @@ NXT_EXPORT void nxt_memcpy_upcase(u_char *dst, const u_char *src,
 nxt_inline void *
 nxt_cpymem(void *dst, const void *src, size_t length)
 {
-    return ((u_char *) memcpy(dst, src, length)) + length;
+    return memcpy(dst, src, length) + length;
 }
 
 


callahad:
@hongzhidao @avahahn Would either of you like to review?

hongzhidao:
Hi,
I have no strong preference either way. If the testing results look good, feel free to do it.
Thanks.

hongzhidao (approved):

hongzhidao (approved):

ac000:
Note in the commit message that GCC et al treat sizeof(void) as 1.

```
$ git range-diff 2bcfd5dc...2fe744cd
1:  2bcfd5dc ! 1:  2fe744cd Decast nxt_cpymem()
    @@ Commit message
         However since we removed support for compilers other than GCC and Clang
         (ending in commit 9cd11133 ("Remove support for Sun's Sun Studio/SunPro
         C compiler")) this is no longer an issue as both GCC and Clang support
    -    arithmetic on void pointers (without the -pedantic option).
    +    arithmetic on void pointers (without the -pedantic option) by treating
    +    the size of a void as 1.
     
         While removing the unnecessary casting in this case doesn't necessarily
         improve type-safety (as we're dealing with void *'s in and out), it does
```

ac000:
Rebased with master

```
$ git range-diff 2fe744cd...4f041328
-:  -------- > 1:  2d4624f0 Make nxt_tstr_is_js() macro public in header
-:  -------- > 2:  a92d8149 http: Refactor format field in nxt_router_access_log_conf_t
-:  -------- > 3:  a760e24a http: Introduce nxt_router_access_log_format_t structure
-:  -------- > 4:  bc838c5e http: Support JSON format in access log
-:  -------- > 5:  a3551790 tests: Add tests for JSON format access log
1:  2fe744cd = 6:  4f041328 Decast nxt_cpymem()
```

PR #1493, opened 2024-11-22
Fix build with wasmtime 27.0.0

osokin:


thresheek:
Does that still work with e.g. wasmtime 24?

thresheek:
> Does that still work with e.g. wasmtime 24?

Ah, it looks like CI tells us it does not: https://github.com/nginx/unit/actions/runs/11979528350/job/33401907630?pr=1493

ac000:
I guess we also want write perms to match current behaviour...

osokin:
> I guess we also want write perms to match current behaviour...

Updated.

ac000 (approved):

ac000:
Merged. Thanks Sergey!

PR #1495, opened 2024-11-26
Packages: support rpm 4.20+

thresheek:
In 4.20, rpm changed the way build roots are named.  Our packaging relies on the symlinks being created as a part of a rule to make sure targets are being built correctly.

This patch implements (a hacky) way to check what the build root should be named on a currently running rpm version and adjusts the symlinks accordingly.


ac000 (approved):
If it works...

thresheek:
It seems to work at least on Amazon Linux 2, RHEL 9 and Fedora 41 😅

PR #1496, opened 2024-11-26
Packages: added Fedora 41 support.

thresheek:
While at it, removed support for Fedora 39 as it's EOL.


ac000 (approved):
Minor nit

s/Packages: added Fedora 41 support./pkgs: add  Fedora 41 support/

other than that...

thresheek:
Thank you for the review!

This actually depends on https://github.com/nginx/unit/pull/1495 to be fully correct :-)

thresheek:
...As Fedora 41 is the first distro that ships with a newer rpm.

Issue #1453, opened 2024-10-09
python3.13 support + docker


imkebe:
Python 3.13 had been released. There is no build for such version of the docker images

callahad:
@thresheek Would it be possible for SysEng to bump our Python language versions without a Unit release, or would that be a bit of a pain? 🙏

thresheek:
I think we can do the following:

- Bump Python version to 3.13+3.12 in the `master` branch
- Backport those commits to the `packaging` branch
- Do a new PR to Docker Library to pick up those changes.

A couple of things that are potentially breaking someone's workflows, as we typically publish two latest language runtime versions:
- we'd drop 3.11 from the library definition (published images like `unit:1.33.0-python3.11` and `unit:python3.11` will stay as is, albeit not easily discoverable)
- we'd move `unit:python3` and `unit:python` tags to 3.13 instead of a current 3.12

imkebe:
One thing to consider in general 3.13 release. Default Python 3.13 has similar behaviour like 3.12 however there is new 3.13t version that has changes to GIL (threading mechanism). This version is supported by Unit should be tested.

callahad:
@thresheek Would it be possible to avoid the potential of breakage by instead:

- Keeping 3.11 on the `packaging` branch (so Docker Library publishes 3.11, 3.12, and 3.13)
- Keeping `unit:python` at its current 3.12, so it only revs when we do our next full Unit release?

@imkebe I'm super excited about the GIL removal, but we'll probably not post images to Docker Hub while it's still experimental. I do encourage you to try it locally, though, and report back!

thresheek:
The [current implementation](https://github.com/nginx/unit/blob/master/pkg/docker/Makefile#L155) of the ["library" definition](https://github.com/docker-library/official-images/blob/master/library/unit) sorts the versions and assigns non-versioned variants to whatever the latest is.

We can cheat a bit and provide the manually edited file, of course, however it's not very nice from etical POV since we also publish "this file is generated via ..." line in there...

callahad:
Fair enough. I'm fine with splitting the difference: publish 3.11, 3.12, and 3.13; and allow `unit:python` to point to 3.13.

If you're implicitly pointing at the latest version of Python, *you get* the latest version of Python.

thresheek:
Python 3.13 is now supported via Docker.

PR #1497, opened 2024-11-26
Docker: update Python and PHP versions

thresheek:
Note that this updates the version in the master branch; the commits to add (and not replace the older versions) will be in a separate push to `packaging` branch as discussed in https://github.com/nginx/unit/issues/1453


ac000 (approved):

luchaninov (approved):

thresheek:
I suppose it doesnt make sense to update 1.33 similarly since we're going to do 1.34 any time soon.

Issue #1510, opened 2024-12-13
otel-enabled build fails on systems without openssl.pc


thresheek:
Hello!

I've noticed the following on an operating system that doesn't have an openssl.pc:

```
checking for OTEL requirements:
  - checking for rust compiler ... found
  - checking for cargo ... found
  - checking for OpenSSL library ... found
Package openssl was not found in the pkg-config search path.
Perhaps you should add the directory containing `openssl.pc'
to the PKG_CONFIG_PATH environment variable
No package 'openssl' found
```

On this OS, the .pc file is named openssl11.pc - but that doesnt really matter in this case I think.

Looking at the code, it comes the following part of `auto/otel`:

```
    nxt_feature="OpenSSL library"
    nxt_feature_run=yes
    nxt_feature_incs=
    nxt_feature_libs="-lssl -lcrypto"
    nxt_feature_test="#include <openssl/ssl.h>

                      int main(void) {
                          SSL_library_init();
                          return 0;
                      }"
    . auto/feature

    if [ ! $nxt_found = yes ]; then
        $echo
        $echo $0: error: OpenTelemetry support requires OpenSSL.
        $echo
        exit 1;
    fi

    NXT_OTEL_LIBS="-lssl -lcrypto"
    if [ $(which pkg-config) ]; then
        NXT_OTEL_LIBS="$(pkg-config openssl --cflags --libs)"
    fi
    cat << END >> $NXT_AUTO_CONFIG_H

```

Why do we even need this check when we have `auto/ssltls` ?  Given the fact that `auto/ssltls` is getting called in `configure` much earlier, cant we just reuse the knowledge we already gathered?

ac000:
> Why do we even need this check when we have auto/ssltls ? Given the fact that auto/ssltls is getting called in configure much earlier, cant we just reuse the knowledge we already gathered?

They're really for two different things and you may want OTEL but not TLS/https support...

What OS was this?

thresheek:
That's Amazon Linux 2 with openssl11-devel.

ac000:
I think the simplest thing to do is to just drop the pkg-config check. If we got there we already successfully built wirh `-lssl -lcrypto`, so just go with that and it matches what we do in `auto/ssltls`.

PR #1513, opened 2024-12-13
auto/otel: Remove pkg-config check

ac000:
```
    auto/otel: Remove pkg-config check

    @thresheek reported an issue trying to configure OTEL support on Amazon
    Linux 2
    
      checking for OTEL requirements:
        - checking for rust compiler ... found
        - checking for cargo ... found
        - checking for OpenSSL library ... found
      Package openssl was not found in the pkg-config search path.
      Perhaps you should add the directory containing `openssl.pc'
      to the PKG_CONFIG_PATH environment variable
      No package 'openssl' found
    
    We successfully built the test program with '-lssl -lcrypto', but then
    tried to use 'pkg-config openssl --cflags --libs' to override
    NXT_OTEL_LIBS.
    
    On Amazon Linux2 there is no openssl.pc, they have a openssl11.pc.
    
    Let's just remove the pkg-config check, if we get here, we have
    successfully built with '-lssl -lcrypto', so just go with that (it also
    matches what we do in auto/ssltls).
    
    Closes: https://github.com/nginx/unit/issues/1510
```

### DIFF
--- a/auto/otel
+++ b/auto/otel
@@ -41,9 +41,7 @@ if [ $NXT_OTEL = YES ]; then
     fi
 
     NXT_OTEL_LIBS="-lssl -lcrypto"
-    if [ $(which pkg-config) ]; then
-        NXT_OTEL_LIBS="$(pkg-config openssl --cflags --libs)"
-    fi
+
     cat << END >> $NXT_AUTO_CONFIG_H
 
 #ifndef NXT_HAVE_OTEL


thresheek:
I agree with reasoning in https://github.com/nginx/unit/issues/1510#issuecomment-2541875665 - this pkg-config call actually shouldnt be required if we already linked with openssl.  And if we failed, a caller needs to provide proper `--cc-opt` and `--ld-opt`.

LGTM

thresheek (approved):

PR #1520, opened 2024-12-20
auto/otel: Fix breakage on non-Linux platforms

ac000:
auto/otel: Fix breakage on non-Linux platforms
    
    When building with --otel on macOS for example I was seeing compile
    failures with the cpu_set_t stuff which should only be used under Linux.
    
    It turned out that despite
    
      checking for Linux sched_getaffinity() ... not found
    
    we were getting
    
      #ifndef NXT_HAVE_LINUX_SCHED_GETAFFINITY
      #define NXT_HAVE_LINUX_SCHED_GETAFFINITY  1
      #endif
    
    in build/include/nxt_auto_config.h
    
    It seems this was due to the
    
        . auto/feature
    
    in auto/otel, this check happens right after the above. Without having
    
        nxt_feature_name=NXT_HAVE_OTEL
    
    set.
    
    Instead we were adding the define for that manually.
    
    Doing auto/feature without having a nxt_feature_name must have used the
    last set one and enabled it.
    
    Set nxt_feature_name and remove the manual editing of nxt_auto_config.h
    
### DIFF
--- a/auto/otel
+++ b/auto/otel
@@ -22,6 +22,7 @@ if [ $NXT_OTEL = YES ]; then
     $echo -n "  - "
 
     nxt_feature="OpenSSL library"
+    nxt_feature_name=NXT_HAVE_OTEL
     nxt_feature_run=yes
     nxt_feature_incs=
     nxt_feature_libs="-lssl -lcrypto"
@@ -42,11 +43,4 @@ if [ $NXT_OTEL = YES ]; then
 
     NXT_OTEL_LIBS="-lssl -lcrypto"
 
-    cat << END >> $NXT_AUTO_CONFIG_H
-
-#ifndef NXT_HAVE_OTEL
-#define NXT_HAVE_OTEL 1
-#endif
-
-END
 fi


thresheek:
This patch seems correct to address this particular issue.

My fix for macOS build uses a similar trick, but we also need to link with additional frameworks on that OS: https://github.com/thresheek/unit/commit/660e8535944c4cd7e44098e6a9ca739be05749cc

Also, another fix is required to be able to build on macOS: https://github.com/thresheek/unit/commit/18381b103f3e6e5ffceb122d6f0fe74d749ef5e5

We probably want to incorporate those changes in this PR for the sake of completeness.

ac000:
> This patch seems correct to address this particular issue.

It would affect any system that didn't have a Linux compatible sched_getaffinity(2), not that I was aware of any that did, but it seems at least FreeBSD does, so I'll need to tweak the commit subject...

> My fix for macOS build uses a similar trick, but we also need to link with additional frameworks on that OS: [thresheek@660e853](https://github.com/thresheek/unit/commit/660e8535944c4cd7e44098e6a9ca739be05749cc)

Weird, I wonder what is doing that... it certainly uses regular OpenSSL for general TLS support...

> Also, another fix is required to be able to build on macOS: [thresheek@18381b1](https://github.com/thresheek/unit/commit/18381b103f3e6e5ffceb122d6f0fe74d749ef5e5)

This one makes sense.
 
> We probably want to incorporate those changes in this PR for the sake of completeness.

But then there is some other breakage also with `--otel`

(in auto/make)
```make
 87 ifeq (\$D,1)
 88         NXT_OTEL_LIB_LOC = $NXT_OTEL_LIB_DIR/target/debug/libotel.a
 89 else
 90         NXT_OTEL_LIB_LOC = $NXT_OTEL_LIB_DIR/target/release/libotel.a
 91 endif
```

Prevents the use of `make` on at least FreeBSD, however, it also breaks with gmake as `D` has not been set as that only happens when

```shell
 44 if [ -n "$NXT_GNU_MAKE" ] || [ $NXT_OS = "SunOS" ]; then
```

And then above that we have

```make
 64 ifeq (\$D,1)                                                                
 65         CFLAGS += -O0                                                       
 66         RUST_FLAGS += --debug                                               
 67 else                                                                        
 68         RUST_FLAGS += --release                                             
 69 endif
```

However that is guarded by the

```shell
 44 if [ -n "$NXT_GNU_MAKE" ] || [ $NXT_OS = "SunOS" ]; then                 
```

So on the likes of FreeBSD RUST_FLAGS isn't set all...

I don't mind making GNU make a build dependency... however at the moment it seems --otel is fubar on anything non-Linux...  (between needing sched_getaffinity(2) and GNU make being the default make(1))

ac000:
Some of this would be simpler if we could just tell `rustc` to write the library to `build/lib/libotel.a`, oh right, we can...

`cargo build ... --emit link=../../build/lib/libotel.a`

and then just do debug builds by default and enable release builds by setting a cargo environment variable, but that seems to be a stumbling block as unless I missed it, that seems to be the _one_ thing you can't set...

ac000:
While we're at it could we

`s/NXT_OTEL_LIB_LOC/NXT_OTEL_LIB_STATIC`

ac000:
Closing this PR as this is now a part of https://github.com/nginx/unit/pull/1527

Issue #1526, opened 2024-12-31
Segmentation faults in tests with otel enabled


thresheek:
Hello,

Running the testsuite (`make test`) causes segmentation faults on a variety of systems, e.g. Ubuntu 22.04 arm64, with Unit configured with ```./configure --otel --openssl```:

```
thresh@ubuntu22:~/unit$ env python3 -m pytest --user=nobody --save-log --restart test/test_static_share.py
================================================================================================================== test session starts ===================================================================================================================
platform linux -- Python 3.10.12, pytest-6.2.5, py-1.10.0, pluggy-0.13.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /home/thresh/unit/test, configfile: pytest.ini
collected 3 items

test/test_static_share.py::test_share_array FAILEDPath to unit.log:
/tmp/unit-test-q9g7yzkj/unit.log

2024/12/31 23:01:00 [warn] 57741#57741 Unit is running unprivileged, then it cannot use arbitrary user and group.
2024/12/31 23:01:00 [info] 57741#57741 unit 1.35.0 started
2024/12/31 23:01:00 [info] 57742#57742 discovery started
2024/12/31 23:01:00 [notice] 57742#57742 no modules matching: "/home/thresh/unit/build/lib/unit/modules/*.unit.so" found
2024/12/31 23:01:00 [info] 57741#57741 controller started
2024/12/31 23:01:00 [notice] 57741#57741 process 57742 exited with code 0
2024/12/31 23:01:00 [info] 57744#57744 router started
2024/12/31 23:01:00 [info] 57744#57744 OpenSSL 3.0.2 15 Mar 2022, 30000020
2024/12/31 23:01:00 [alert] 57741#57741 process 57744 exited on signal 11 (core dumped)
2024/12/31 23:01:00 [info] 57753#57753 router started
2024/12/31 23:01:00 [info] 57753#57753 OpenSSL 3.0.2 15 Mar 2022, 30000020
2024/12/31 23:01:00 [notice] 57741#57741 process 57743 exited with code 0
2024/12/31 23:01:00 [notice] 57741#57741 process 57753 exited with code 0
Path to unit.log:
/tmp/unit-test-q9g7yzkj/unit.log

2024/12/31 23:01:00 [warn] 57741#57741 Unit is running unprivileged, then it cannot use arbitrary user and group.
2024/12/31 23:01:00 [info] 57741#57741 unit 1.35.0 started
[...]

PR #1531, opened 2025-01-07
otel: fix segfaults when otel not configured

### LINKED ISSUE #1523: Segfault when receiving the traceparent header and otel is not configured
We have encountered this problem with Nginx Unit after upgrading to 1.34. Our setup is:

- No otel configuration block in Nginx Unit configuration;
- Our system has tracing enable on some other components of the system, this result in a `traceparent` header being present on the request.

In this case, the router of Nginx Unit segfault every time when receiving the request. This seems to be due to a missing check on whereas the otel structure is initialized in `nxt_otel_parse_traceparent` (not verified but probably too in `nxt_otel_parse_tracestate`).

 We will test, as a workaround, to configure otel in Nginx Unit which should make the problem disappear.

avahahn:
Fixes #1523 and #1526 by adding null checks to cover unconfigured OpenTelemetry cases.

Both tests and traffic were tested on both x86_64 Linux and arm64 Linux.

### DIFF
--- a/src/nxt_otel.c
+++ b/src/nxt_otel.c
@@ -67,7 +67,7 @@ nxt_otel_propagate_header(nxt_task_t *task, nxt_http_request_t *r)
      * if we didn't inherit a trace id then we need to add the
      * traceparent header to the request
      */
-    } else if (r->otel->trace_id == NULL) {
+    } else {
 
         nxt_otel_rs_copy_traceparent(traceval, r->otel->trace);
 
@@ -92,15 +92,6 @@ nxt_otel_propagate_header(nxt_task_t *task, nxt_http_request_t *r)
 
         nxt_otel_rs_add_event_to_trace(r->otel->trace,
                                        &traceparent_name, &traceparent);
-
-    /*
-     * potentially nxt_http_request_error called before headers
-     * finished parsing
-     */
-    } else {
-        nxt_log(task, NXT_LOG_DEBUG,
-                "not propagating tracing headers for missing trace");
-        return;
     }
 
     f = nxt_list_add(r->resp.fields);
--- a/src/nxt_otel.c
+++ b/src/nxt_otel.c
@@ -311,17 +311,13 @@ nxt_otel_test_and_call_state(nxt_task_t *task, nxt_http_request_t *r)
 void
 nxt_otel_request_error_path(nxt_task_t *task, nxt_http_request_t *r)
 {
-    if (r->otel->trace == NULL) {
+    if (r->otel == NULL || r->otel->trace == NULL) {
         return;
     }
 
     // response headers have been cleared
     nxt_otel_propagate_header(task, r);
-
-    // collect span immediately
-    if (r->otel) {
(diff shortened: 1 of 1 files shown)


ac000 (commented):
```
otel: fix segfaults when otel not configured

This commit adds null checks for the request->otel object that
were missed in the Traceparent and Tracestate routines. This
fix solves issues 1526 and 1523

Signed-off-by: Ava Hahn <a.hahn@f5.com>
```

s/null/NULL/

This could also do with some closes and a fixes tag i.e.

```
...

Closes: https://github.com/nginx/unit/issues/1523
Closes: https://github.com/nginx/unit/issues/1526
Fixes: 9d3dcb800 ("otel: add build tooling to include otel code")
Signed-off-by: Ava Hahn <a.hahn@f5.com>
```

and then remove the `The fix solves...` line.

Did you audit the other functions that assume `r->otel` is _not_ `NULL`? E.g.

- nxt_otel_request_error_path()
- nxt_otel_parse_traceparent()
- nxt_otel_parse_tracestate()

Also taking a closer look at `nxt_otel_request_error_path()`

```c
void                                                                            
nxt_otel_request_error_path(nxt_task_t *task, nxt_http_request_t *r)            
{                                                                               
    if (r->otel->trace == NULL) {                                               
        return;                                                                 
    }                                                                           
                                                                                
    // response headers have been cleared                                       
    nxt_otel_propagate_header(task, r);                                         
                                                                                
    // collect span immediately                                                 
    if (r->otel) {                                                              
        nxt_otel_state_transition(r->otel, NXT_OTEL_COLLECT_STATE);             
    }                                                                           
    nxt_otel_test_and_call_state(task, r);                                      
} 
```

The `if (r->otel)` check (aside from the fact it should do an explicit NULL check), looks superfluous.

We've already checked `r->otel->trace == NULL`, so assuming we didn't crash, then we know r->otel isn't NULL. But maybe we should be checking that at the top of the function?

For the avoidance of any doubt. If it's superfluous it should be removed in a separate commit, OTOH if we should be checking `r->otel == NULL` at the top of the function then just do it in this commit...

While we're at it and whatever the fix here is it should be done as a separate commit (and then we can decide if it warrants going into 1.34.1), in `nxt_otel_propagate_header()` we have

```c
if (r->otel->trace_id != NULL) {
...
} else if (r->otel->trace_id == NULL) {
...
} else {
...
}
```

`->trace_id` can _only_ be NULL or !NULL, you see where this is going...

avahahn:
> Did you audit the other functions that assume r->otel is not NULL? E.g.
> 
> nxt_otel_request_error_path()
> nxt_otel_parse_traceparent()
> nxt_otel_parse_tracestate()

...

>  OTOH if we should be checking r->otel == NULL at the top of the function then just do it in this commit...

I am not sure what commit you are looking at but that is what this commit does. Those three specific functions and your OTOH comment are all directly addressed in the patch that github shows. Did you end up with something wildly different somehow?

ac000:
No, you're right (in my mind I'd applied the patch locally)

Looks good then...

ac000 (approved):

PR #1541, opened 2025-01-11
[nxt_http_static] add `.mjs` extension to default_types

kedartal:
Associate file extension `.mjs` with `application/javascript`.

Context: common output of static site generators. There's little risk of ambiguity for this extension, so might as well support it out of the box.

### DIFF
--- a/src/nxt_http_static.c
+++ b/src/nxt_http_static.c
@@ -957,6 +957,7 @@ nxt_http_static_mtypes_init(nxt_mp_t *mp, nxt_lvlhsh_t *hash)
         { nxt_string("text/x-rst"),     ".rst"   },
 
         { nxt_string("application/javascript"),  ".js"   },
+        { nxt_string("application/javascript"),  ".mjs"  },
         { nxt_string("application/json"),        ".json" },
         { nxt_string("application/xml"),         ".xml"  },
         { nxt_string("application/rss+xml"),     ".rss"  },


ac000:
I'm in two minds about this.

There are several other _obvious_ missing entries...

But then you can always add your own (and modify existing ones) like

```json
"settings": {
    "http": {
        "static": {
            "mime_types": {
                "application/javascript": [
                    ".js",
                    ".mjs"
                ]
            }
        }
    }
},
```

Each _mime_types_ entry can either be an array or a string.

kedartal:
@ac000 Of course, but that requires a call to the control API -- in practice, often a _separate_ call from the one for configuring listeners, applications, routes et cetera. IMHO we should add all or most of the standard types: the impacts on the binary size and runtime complexity are negligible, standard MIME changes very slowly so little maintenance cost, and obviously it's easier for users so good for adoption. As you said, users can always append / replace, so including all or most types by default seems natural.
An alternative might be to offload the types list into a file processed by a build `make` script, but I guess most users prefer installing from a package manager rather than build from sources, and I'm not sure it would make life easier for anyone.

ac000:
I'm not against adding it... let's see what others think...

Cc: @hongzhidao 
Cc @avahahn :

hongzhidao:
Hi,
I’m fine with it.

ac000 (approved):

ac000:
Thanks Tal, applied (with a minor tweak to the commit subject).

kedartal:
@ac000 Thanks!

PR #1543, opened 2025-01-21
Fix GCC 15 Wunterminated-string-initialization issues

ac000:
```
    Fix GCC 15 Wunterminated-string-initialization issues
    
    When building with the upcoming GCC 15, we are hit by the new
    Wunterminated-string-initialization compiler warning (however, due to
    -Werror, this is a build failure). E.g.
    
      src/nxt_string.c: In function ‘nxt_encode_uri’:
      src/nxt_string.c:601:36: error: initializer-string for array of ‘unsigned char’ is too long [-Werror=unterminated-string-initialization]
        601 |     static const u_char  hex[16] = "0123456789ABCDEF";
            |                                    ^~~~~~~~~~~~~~~~~~
    
    We have created a character array containing 16 elements, but is not NUL
    terminated. While this type of (lookup/translation table) use case is
    perfectly valid and we could simply disable the warning, it is probably
    still a useful warning to have enabled.
    
    There is a 'nonstring' attribute, but that doesn't influence this
    warning. Well not yet at least... but that looks to change!
    
    Lets just do what nginx does and create them NUL terminated. Except in
    src/nxt_http_parse.c where they *really* want to be non-NUL
    terminated and we use *arrays* of chars.
    
    Link: <https://gcc.gnu.org/onlinedocs/gcc-14.2.0/gcc/Common-Variable-Attributes.html>
```

### DIFF
--- a/src/nxt_http_parse.c
+++ b/src/nxt_http_parse.c
@@ -164,8 +164,10 @@ nxt_http_parse_request_line(nxt_http_request_parse_t *rp, u_char **pos,
     nxt_http_ver_t           ver;
     nxt_http_target_traps_e  trap;
 
-    static const nxt_http_ver_t  http11 = { "HTTP/1.1" };
-    static const nxt_http_ver_t  http10 = { "HTTP/1.0" };
+    static const nxt_http_ver_t  http11 =
+                            {{ 'H', 'T', 'T', 'P', '/', '1', '.', '1' }};
+    static const nxt_http_ver_t  http10 =
+                            {{ 'H', 'T', 'T', 'P', '/', '1', '.', '0' }};
 
     p = *pos;
 
@@ -516,7 +518,8 @@ nxt_http_parse_field_name(nxt_http_request_parse_t *rp, u_char **pos,
     size_t    len;
     uint32_t  hash;
 
-    static const u_char  normal[256]  nxt_aligned(64) =
+    /* The last '\0' is not needed because the string is NUL terminated */
+    static const u_char  normal[]  nxt_aligned(64) =
         "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"
     /*   \s ! " # $ % & ' ( ) * + ,        . /                 : ; < = > ?   */
         "\0\1\0\1\1\1\1\1\0\0\1\1\0" "-" "\1\0" "0123456789" "\0\0\0\0\0\0"
@@ -529,7 +532,7 @@ nxt_http_parse_field_name(nxt_http_request_parse_t *rp, u_char **pos,
         "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"
         "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"
         "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"
-        "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0";
+        "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0";
 
     p = *pos + rp->field_name.length;
     hash = rp->field_hash;
--- a/src/nxt_sprintf.c
+++ b/src/nxt_sprintf.c
@@ -112,8 +112,8 @@ nxt_vsprintf(u_char *buf, u_char *end, const char *fmt, va_list args)
     nxt_sprintf_t        spf;
     nxt_file_name_t      *fn;
 
-    static const u_char  hexadecimal[16] = "0123456789abcdef";
-    static const u_char  HEXADECIMAL[16] = "0123456789ABCDEF";
+    static const u_char  hexadecimal[] = "0123456789abcdef";
+    static const u_char  HEXADECIMAL[] = "0123456789ABCDEF";
     static const u_char  nan[] = "[nan]";
     static const u_char  null[] = "[null]";
     static const u_char  infinity[] = "[infinity]";
--- a/src/nxt_string.c
+++ b/src/nxt_string.c
@@ -592,13 +592,14 @@ nxt_decode_uri_plus(u_char *dst, u_char *src, size_t length)
 }
 
 
+static const u_char  nxt_hex[] = "0123456789ABCDEF";
+
+
 uintptr_t
 nxt_encode_uri(u_char *dst, u_char *src, size_t length)
 {
     u_char      *end;
     nxt_uint_t  n;
-
-    static const u_char  hex[16] = "0123456789ABCDEF";
 
     end = src + length;
 
@@ -624,8 +625,8 @@ nxt_encode_uri(u_char *dst, u_char *src, size_t length)
 
         if (nxt_uri_escape[*src >> 5] & (1U << (*src & 0x1f))) {
             *dst++ = '%';
-            *dst++ = hex[*src >> 4];
-            *dst++ = hex[*src & 0xf];
+            *dst++ = nxt_hex[*src >> 4];
+            *dst++ = nxt_hex[*src & 0xf];
 
         } else {
             *dst++ = *src;
@@ -643,8 +644,6 @@ nxt_encode_complex_uri(u_char *dst, u_char *src, size_t length)
 {
     u_char      *reserved, *end, ch;
     nxt_uint_t  n;
-
-    static const u_char  hex[16] = "0123456789ABCDEF";
 
     reserved = (u_char *) "?#\0";
 
@@ -689,8 +688,8 @@ nxt_encode_complex_uri(u_char *dst, u_char *src, size_t length)
 
             } else {
                 *dst++ = '%';
-                *dst++ = hex[ch >> 4];
-                *dst++ = hex[ch & 0xf];
+                *dst++ = nxt_hex[ch >> 4];
+                *dst++ = nxt_hex[ch & 0xf];
                 continue;
             }
         }


ac000:
- Highlight that this is a build failure...

```
$ git range-diff 72ea5ce9...98f11e65
1:  72ea5ce9 ! 1:  98f11e65 Fix GCC 15 Wunterminated-string-initialization issues
    @@ Commit message
         Fix GCC 15 Wunterminated-string-initialization issues
     
         When building with the upcoming GCC 15, we are hit by the new
    -    Wunterminated-string-initialization compiler warning. E.g.
    +    Wunterminated-string-initialization compiler warning (however, due to
    +    -Werror, this is a build failure). E.g.
     
           src/nxt_string.c: In function ‘nxt_encode_uri’:
           src/nxt_string.c:601:36: error: initializer-string for array of ‘unsigned char’ is too long [-Werror=unterminated-string-initialization]
```

hongzhidao (commented):
Sorry that I’m unable to check if the change is reasonable.
As I understand, we introduce nxt_str_t to handle string with a non traditional way, that it explicitly specify its length. In this case, the stuff like hex is an array of u_char/uint8 type, I think they are not entirely strings whose ending char ‘\0’ is used.
I’d like to ask team’s review.
Thanks.

ac000:
Taking this as an example

```c
    static const u_char  hex[16] = "0123456789ABCDEF";                          
```

That declares a non-NUL terminated string (by loosing the terminating `\0` from the string literal), which in this case is OK, but could often be an error, which this new compiler warning warns about. See [here](https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=44c9403ed1833ae71a59e84f9e37af3182be0df5) for details (Yes, this was introduced by Alex!).

You know that these are usually used as translation table type things and is really being treated as just an array of characters.

So assuming we want to keep this new warning enabled. The two main options are

Make it NUL terminated, which this patch does. (As you say ) It doesn't matter that it's NUL terminated or not (other than using an extra byte) as we only ever access the array indices 0-15 and never the NUL byte.

This is what [nginx](https://github.com/nginx/nginx/blob/5d5d9adccfeaff7d5926737ee5dfa43937fe5899/src/core/ngx_string.c#L495) does.

The other option would be to define it as a character array like

```c
    static const u_char  hex[16] = { '0', '1', '2', '3', '4', '5', '6', '7',
                                     '8', '9', 'A', 'B', 'C', 'D', 'E', 'F' }; 
```

I chose the first option for most of the cases as it's simpler and they don't matter being NUL terminated.

You can see I took the second option for a couple of them where it does matter that they are non-NUL terminated.

I suppose you _could_ maybe declare it as a nxt_str_t... OK, lets stop there... (besides that's just going to give you a pointer to a NUL terminated string literal, which is no better than char [] and char [] better describes its purpose...)

ac000:
- There is work under way to make the `nonstring` attribute silence the `Wunterminated-string-initialization` warning. Although I think it's too late for that work to make it into GCC 15 as it's now in the regression and documentation fixes stage...

```
 $ git range-diff 98f11e65...36d69013
1:  98f11e65 ! 1:  36d69013 Fix GCC 15 Wunterminated-string-initialization issues
    @@ Commit message
         still a useful warning to have enabled.
     
         There is a 'nonstring' attribute, but that doesn't influence this
    -    warning.
    +    warning. Well not yet at least... but that looks to change!
     
         Lets just do what nginx does and create them NUL terminated. Except in
         src/nxt_http_parse.c where they *really* want to be non-NUL
         terminated and we use *arrays* of chars.
     
    +    Link: <https://gcc.gnu.org/onlinedocs/gcc-14.2.0/gcc/Common-Variable-Attributes.html>
    +    Link: <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117178#c21>
     
      ## src/nxt_http_parse.c ##
```

hongzhidao:
To minimize it, it’s d discussion about if the below code is good or not.
```c
static const u_char hex[16] = "0123456789ABCDEF";
```

Personally I prefer to keep the explicit [16] because:
1. It clearly expresses the intent of a hexadecimal lookup table
2. Provides explicit boundary checking
3. Self-documenting code - size is known without looking at the initialization value

Regarding string initialization: The original syntax fits this lookup table use case. GCC 15's warning is a bit overzealous here. We don't actually need this string to be NUL-terminated since it's only accessed through 0-15 indices.

Removing [16] just to satisfy the compiler warning actually obscures the code's intent. If we really need to handle this warning, I'd prefer to suppress it with a compiler directive rather than change the code structure.

Welcome to others suggestions on it. It’s also a discussion if we need to support GCC 15 and the new compile option.

ac000:
> To minimize it, it’s d discussion about if the below code is good or not.
> 
> ```c
> static const u_char hex[16] = "0123456789ABCDEF";
> ```

 > Personally I prefer to keep the explicit [16] because:

This fails to build (due to -Werror) with GCC 15 (I.e. the Fedora Rawhide GCC CI check, the Clang check is failing due to Ruby 3.4 API deprecations)... which is the whole point of this PR...

> Regarding string initialization: The original syntax fits this lookup table use case. GCC 15's warning is a bit overzealous here. We don't actually need this string to be NUL-terminated since it's only accessed through 0-15 indices.

No, I know we don't _need_ it to be NUL terminated, if we did then we had bigger problems. 

Making it NUL terminated is harmless however... and is what the nginx folks did. nginx actually builds fine with GCC 15 (with -Wextra).

> Removing [16] just to satisfy the compiler warning actually obscures the code's intent. If we really need to handle this

Yes, I agree with that...

>warning, I'd prefer to suppress it with a compiler directive rather than change the code structure.
> 
> Welcome to others suggestions on it. It’s also a discussion if we need to support GCC 15 and the new compile option.

Do we need to support GCC 15?, err yes... why wouldn't we? We support all versions of Clang/GCC that support -std=gnu11

Seeing as there is work in progress to make the `nonstring` variable attribute quell this warning, that will hopefully make GCC 16. We could temporarily disable this warning on GCC and the re-enable it for GCC 16 (hopefully) and later.

hongzhidao:
I thought the PR is only because of the option -Wu-s-i* support in GCC 15.

ac000:
Correct.

hongzhidao:
While I'm comfortable with the current approach, I'll stay neutral as others might prefer the new GCC
 15 option.

src/nxt_http_parse.c
```diff
@@ -516,7 +518,8 @@ nxt_http_parse_field_name(nxt_http_request_parse_t *rp, u_char **pos,
     size_t    len;
     uint32_t  hash;
 
-    static const u_char  normal[256]  nxt_aligned(64) =
+    /* The last '\0' is not needed because the string is NUL terminated */
+    static const u_char  normal[]  nxt_aligned(64) =
```
alejandro-colomar, line 522: Hi @ac000 ,

That comment is incorrect:

```
/* The last '\0' is not needed because the string is NUL terminated */
```

Even though this is a "string literal", it is **not** a string.  The standard prohibits having NUL bytes in the middle of a string, and this is thus not a string.  See footnote 75 in C23 (n3220): <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3220.pdf#subsection.6.4.5>.

> 75)A string literal may not be a string (see 7.1.1), because a null character can be embedded in it by a \0 escape sequence.

So I would disable the diagnostic.  This change (I'm referring specifically to this hunk, and not the entire patch; most of the patch, I like it) is bad for readability, IMO.  Instead, I'd take the opportunity to fix the few places where it wouldn't hurt to adapt to the warning, but then disable it until `[[gnu::nonstring()]]` allows you to re-enable it.

Cheers,
Alex

Cc: @thesamesam

src/nxt_sprintf.c
```diff
@@ -112,8 +112,8 @@ nxt_vsprintf(u_char *buf, u_char *end, const char *fmt, va_list args)
     nxt_sprintf_t        spf;
     nxt_file_name_t      *fn;
 
-    static const u_char  hexadecimal[16] = "0123456789abcdef";
-    static const u_char  HEXADECIMAL[16] = "0123456789ABCDEF";
+    static const u_char  hexadecimal[] = "0123456789abcdef";
+    static const u_char  HEXADECIMAL[] = "0123456789ABCDEF";
```
alejandro-colomar, line 116: This one I like it.  Everything that need not be a nonstring, should be a string.  Everybody knows hexadecimal enough that the lost information is negligible.

src/nxt_http_parse.c
ac000: > Hi @ac000 ,
> 
> That comment is incorrect:
> 
> ```
> /* The last '\0' is not needed because the string is NUL terminated */
> ```

Admittedly I just yanked that comment out of nginx, but In this case I would take string to just mean "array of characters".

src/nxt_http_parse.c
```diff
@@ -164,8 +164,10 @@ nxt_http_parse_request_line(nxt_http_request_parse_t *rp, u_char **pos,
     nxt_http_ver_t           ver;
     nxt_http_target_traps_e  trap;
 
-    static const nxt_http_ver_t  http11 = { "HTTP/1.1" };
-    static const nxt_http_ver_t  http10 = { "HTTP/1.0" };
+    static const nxt_http_ver_t  http11 =
+                            {{ 'H', 'T', 'T', 'P', '/', '1', '.', '1' }};
+    static const nxt_http_ver_t  http10 =
+                            {{ 'H', 'T', 'T', 'P', '/', '1', '.', '0' }};
```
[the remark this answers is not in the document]
ac000: Oh, it looks hideous, but it meant we could keep the warning enabled... but then that whole nxt_http_ver_t thing is _only_ used in _that_ function and it seems to be purely to allow for doing a string comparison using unit64_t's...

src/nxt_http_parse.c
alejandro-colomar: > > Hi @ac000 ,
> > That comment is incorrect:
> > ```
> > /* The last '\0' is not needed because the string is NUL terminated */
> > ```
> 
> Admittedly I just yanked that comment out of nginx,

Ughhh.  Maybe tell them that their comment is actively harmful?

> but In this case I would take string to just mean "array of characters".

Yeah, but I this one is really too obscure.  The new code feels like a hybrid between a string and a nonstring.  And I'd say the `[256]` was actually informative, since it's not obvious to see how many zeros there are.

"the string is NUL terminated" reminds just so much to a proper string, that it will most likely confuse someone at some point.

I'd keep the old version here.

ac000: https://github.com/nginx/nginx/blob/5d5d9adccfeaff7d5926737ee5dfa43937fe5899/src/http/ngx_http_parse.c#L832 (I tweaked it slightly...)

alejandro-colomar: git-blame(1) points to this commit from Igor: <https://github.com/nginx/nginx/commit/bb8bbb7c0bfe2171c52318435deaece357a95c0c>

Heh, it looks like one of those cases of saving one byte of memory at the expense of the sanity of the reader.
This reminds me of the day I looked into
<https://github.com/nginx/unit/blob/b4201ab371011e94123de5895c0b40943c644a2a/src/nxt_malloc.h#L45>.
I mean, it probably made sense back in 2006.  But today, I would just nuke it all.  :-)

alejandro-colomar: Prior related commits:

<https://github.com/nginx/nginx/commit/e5efadb60ee795e96a51116cc81451eeb3e397cc>

<https://github.com/nginx/nginx/commit/3338cfdfbc5522ca96319ece6d21ddc4d1dce45d>

(sounds like this confused Igor himself.)

ac000: In this context, I would assume "string" just means "string of characters" where string is synonymous with 'series', like you might say "string of robberies".

alejandro-colomar: > In this context, I would assume "string" just means "string of characters" where string is synonymous with 'series', like you might say "string of robberies".

I'd use "array" instead of string.  Or "string literal", if you want.  But using string seems inducing confusion.

Maybe I'm too much of a pedantic regarding wording issues; that's my job.  :-)

ac000: Heh, if we're being _pedantic_ then I think the word would be _pedant_.

Anyway, I'm probably just going to close this PR and simply disable the warning for now, and then look at sprinkling `__attribute__((nonstring))`'s (nicer named) around when Kees' patch makes it in, maybe GCC 16, maybe even GCC 15, which would be prefect, or we may be sitting here in two years time still waiting for it to land. Never can tell... (No, think positive thoughts...)

alejandro-colomar: > Heh, if we're being _pedantic_ then I think the word would be _pedant_.

Lol.  True.

> Anyway, I'm probably just going to close this PR and simply disable the warning for now,

I do like that one, however: <https://github.com/nginx/unit/pull/1543#discussion_r1925605148>

I'd do that change, and then turn the diagnostic off.

> and then look at sprinkling `__attribute__((nonstring))`'s (nicer named) around when Kees' patch makes it in, maybe GCC 16, maybe even GCC 15, which would be prefect, or we may be sitting here in two years time still waiting for it to land. Never can tell... (No, think positive thoughts...)

:-)

hongzhidao:
> That declares a non-NUL terminated string (by loosing the terminating \0 from the string literal), which in this case is OK, but could often be an error, which this new compiler warning warns about. See [here](https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=44c9403ed1833ae71a59e84f9e37af3182be0df5) for details (Yes, this was introduced by Alex!).

I read the related threads, personally I’m not a fan of this new compile option.

Especially when I read one of them
https://lists.gnu.org/archive/html/groff/2022-11/msg00021.html

> Let's say there's some software with cowboy programmers, which has things like:
    typedef struct {
        size_t  length;
        u_char  *start;
    } str_t;
    #define length(s)   (sizeof(s) - 1)
    #define str_set(str, text)  do       \
    {                                    \
        (str)->length = length(test);    \
        (str)->start = (u_char *) text;  \
    }
(Of course, cowboy programmers don't need terminating NUL bytes, that's for newbies, but that's not today's rant.)
And the use of that macro is typically for things like:
…

The labeled 'cowboy programmer' as one of the most respected engineers created a oss with his team powers 40% of the world's active websites. Perhaps we should be more thoughtful about judging code quality based on personal preferences?

hongzhidao:
I mean your comment about the bias “cowboy programmer”.
People can like or dislike that compiler flags, that's personal preference. But calling someone a 'cowboy programmer', especially someone like professional programmers, is rarely seen in professional discussions.

ac000:
I'm going to disable the warning _for now_... (at least until the `nonstring` attribute is able to quell it).

But I do like it in general. Anything which can help catch "mistakes" is a good thing IHMO. Especially in today's climate where every man and his dog likes to rag on C...

ac000:
This has been supplanted by https://github.com/nginx/unit/pull/1548

PR #1550, opened 2025-01-27
http: Fix WebSockets with Firefox

ac000:
```
    http: Fix WebSockets with Firefox
    
    Firefox (going back a couple of years at least) was unable to open a
    WebSocket connection to Unit due to it sending a Connection header of
    
      Connection: keep-alive, Upgrade
    
    However in Unit we were expecting only a single value in the header.
    
    Fix the 'Connection' parsing in nxt_h1p_connection() to address this.
    
    Closes: https://github.com/nginx/unit/issues/772
```

### DIFF
--- a/src/nxt_h1proto.c
+++ b/src/nxt_h1proto.c
@@ -759,24 +759,23 @@ nxt_h1p_header_buffer_test(nxt_task_t *task, nxt_h1proto_t *h1p, nxt_conn_t *c,
 static nxt_int_t
 nxt_h1p_connection(void *ctx, nxt_http_field_t *field, uintptr_t data)
 {
+    const u_char        *end;
     nxt_http_request_t  *r;
 
     r = ctx;
     field->hopbyhop = 1;
 
-    if (field->value_length == 5
-        && nxt_memcasecmp(field->value, "close", 5) == 0)
-    {
+    end = field->value + field->value_length;
+
+    if (nxt_memcasestrn(field->value, end, "close", 5) != NULL) {
         r->proto.h1->keepalive = 0;
+    }
 
-    } else if (field->value_length == 10
-               && nxt_memcasecmp(field->value, "keep-alive", 10) == 0)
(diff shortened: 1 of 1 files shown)


hongzhidao:
Hi @ac000,
It seems it only works for WS, then do we need to check if the header “Upgrader: websocket” exists?

ac000:
True, as the code currently stands we are assuming that we only upgrade to websockets.

I guess strictly speaking we should check that's the case... but I would probably do that as a separate patch to this as this patch simply fixes the problem of determining if we are *upgrading* the connection.

ac000:
@hongzhidao 

Actually we do already check the `Upgrade` header for `websocket` in `nxt_h1p_upgrade()` where we set `r->proto.h1->upgrade_websocket = 1` if it's found.

If we find `upgrade` in the `Connection` header we set `r->proto.h1->connection_upgrade = 1`

This combination is then checked for in `nxt_h1p_header_process()`

```c
if (h1p->connection_upgrade && h1p->upgrade_websocket) {
```

So I think we're all good, right?

hongzhidao:
Correct.
Btw, is there any other case like “upgrade, close”?

ac000:
Not sure it makes a lot of sense, but sure, we can do that...

hongzhidao (approved):

ac000:
Rebased with master

```
$ git range-diff aede3ed6...df62a2a2
-:  -------- > 1:  7b7b29fc ruby: Fix build failures with Ruby 3.4
-:  -------- > 2:  15037822 Fix build with GCC 15
-:  -------- > 3:  c2279c0d python: Fix Litestar WebSockets compatibility
-:  -------- > 4:  b2866409 otel, tools/unitctl: bump the openssl crate from 0.10.68 to 0.10.70
1:  aede3ed6 = 5:  df62a2a2 http: Fix WebSockets with Firefox
```

PR #1560, opened 2025-02-14
ensure sbindir is proper value for RPM

remicollet:
Starting with Fedora 42 /usr/sbin is no more used

So %{_sbindir} is defined as /usr/bin


thresheek:
Hello!  Thank you for your contribution!

LGTM except for a few minor nits:

 - commit message should be something like:
   `Packaging: Ensure sbindir is properly set on newer Fedora (42+)`

- probably squash into one commit?

- shall we use a `%{__sed}` macro instead of a direct invocation?  I'm not really following the contemporary stance on using macros vs directly calling executables in the specs, but at least it would be inline with pretty much all the other calls in the spec file.

remicollet:
> shall we use a %{__sed} 

No.

remicollet:
Notice: this is only the **mandatory** change required to build F42 package

BTW current spec file is outdated

Especially for
- systemd 
- sysusers

Switch to systemd packaging work for all Fedora and EL >= 8
Switch to sysusers packaging work for all Fedora and EL >= 8, and is mandatory for F42

Ex recent change done for my repo
https://git.remirepo.net/cgit/rpms/unit.git/commit/?id=22a5372cf24b0d2e26aa74948b2e797ac880dfc3

ac000:
@thresheek

Unless you have any objections, I'm going to go ahead and merge this...

thresheek:
Yeah I think it's fine.

thresheek (approved):

PR #1567, opened 2025-03-03
Packaging: update docker golang version. Remove unsupported 1.22, add newer 1.24

oxpa:
Official images check for outdated references in images so we have to update golang image we use.
https://github.com/docker-library/official-images/pull/18559

While at it: git desc doesn't work well with lightweight tags. So replaced it with rev-parse which gives proper results.
IIRC we did similar substitution for other docker projects somewhere.


thresheek (approved):
Looks good!

PR #1575, opened 2025-03-17
Tag some char arrays with __attribute__ ((__nonstring__))

ac000:
As previously promised, this PR comprises three commits that allow us to re-enable the new GCC 15 warning `Wunterminated-string-initialization`.

* The first commit adds a wrapper around `__attribute__ ((__nonstring__))`

As a reminder, this attribute is used to mark character arrays that are intentionally not NUL terminated.

* The second commit tags various character arrays with this attribute.

* The third commit removes the disabling of this new warning.

```
The following changes since commit d7afeb2b94f1cd72ed02403609e5484f9514e5eb:

  java: websocket: Additional payload length validation (2025-02-21 22:49:15 +0000)

are available in the Git repository at:

  git@github.com:ac000/unit.git wusi

for you to fetch changes up to ca60536172916b7c3e4ff62e20be1ae2f3de039d:

  auto/cc: gcc: Don't disable -Wunterminated-string-initialization (2025-03-19 17:58:37 +0000)

----------------------------------------------------------------
Andrew Clayton (3):
      auto/clang: Add a NXT_NONSTRING macro
      Tag various character arrays with NXT_NONSTRING
      auto/cc: gcc: Don't disable -Wunterminated-string-initialization

 auto/cc/test         |  4 ----
 auto/clang           | 14 ++++++++++++++
 src/nxt_clang.h      | 11 +++++++++++
 src/nxt_http_parse.c |  2 +-
 src/nxt_http_parse.h |  2 +-
 src/nxt_sprintf.c    |  4 ++--
 src/nxt_string.c     |  4 ++--
 7 files changed, 31 insertions(+), 10 deletions(-)
```

### DIFF
--- a/auto/cc/test
+++ b/auto/cc/test
@@ -93,10 +93,6 @@ case $NXT_CC_NAME in
 
         NXT_CFLAGS="$NXT_CFLAGS -Wmissing-prototypes"
 
-        # Disable Wunterminated-string-initialization temporarily.
-        # See <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117178#c21>
-        NXT_CFLAGS="$NXT_CFLAGS -Wno-unterminated-string-initialization"
-
         # Stop on warning.
         NXT_CFLAGS="$NXT_CFLAGS -Werror"
 
--- a/auto/clang
+++ b/auto/clang
@@ -194,3 +194,17 @@ nxt_feature_test="static void f(void) __attribute__ ((__unused__));
                       return 0;
                   }"
 . auto/feature
+
+
+nxt_feature="GCC __attribute__ nonstring"
+nxt_feature_name=NXT_HAVE_GCC_ATTRIBUTE_NONSTRING
+nxt_feature_run=
+nxt_feature_incs=
+nxt_feature_libs=
+nxt_feature_test="int main(void) {
+                      static const char str[3] __attribute__ ((__nonstring__)) =
+                          \"ABC\";
+
+                      return !!str[0];
+                  }"
+. auto/feature
--- a/src/nxt_http_parse.c
+++ b/src/nxt_http_parse.c
@@ -516,7 +516,7 @@ nxt_http_parse_field_name(nxt_http_request_parse_t *rp, u_char **pos,
     size_t    len;
     uint32_t  hash;
 
-    static const u_char  normal[256]  nxt_aligned(64) =
+    static const u_char  normal[256]  NXT_NONSTRING nxt_aligned(64) =
         "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"
     /*   \s ! " # $ % & ' ( ) * + ,        . /                 : ; < = > ?   */
         "\0\1\0\1\1\1\1\1\0\0\1\1\0" "-" "\1\0" "0123456789" "\0\0\0\0\0\0"
--- a/src/nxt_http_parse.h
+++ b/src/nxt_http_parse.h
@@ -21,7 +21,7 @@ typedef struct nxt_http_fields_hash_s    nxt_http_fields_hash_t;
 
 
 typedef union {
-    u_char                    str[8];
+    u_char                    str[8] NXT_NONSTRING;
     uint64_t                  ui64;
 
     struct {
--- a/src/nxt_clang.h
+++ b/src/nxt_clang.h
@@ -131,6 +131,17 @@
 #endif
 
 
+#if (NXT_HAVE_GCC_ATTRIBUTE_NONSTRING)
+
+#define NXT_NONSTRING      __attribute__((__nonstring__))
+
+#else
+
+#define NXT_NONSTRING
+
+#endif
+
+
 #if (NXT_HAVE_BUILTIN_POPCOUNT)
 
 #define nxt_popcount       __builtin_popcount
--- a/src/nxt_sprintf.c
+++ b/src/nxt_sprintf.c
@@ -112,8 +112,8 @@ nxt_vsprintf(u_char *buf, u_char *end, const char *fmt, va_list args)
     nxt_sprintf_t        spf;
     nxt_file_name_t      *fn;
 
-    static const u_char  hexadecimal[16] = "0123456789abcdef";
-    static const u_char  HEXADECIMAL[16] = "0123456789ABCDEF";
+    static const u_char  hexadecimal[16] NXT_NONSTRING = "0123456789abcdef";
+    static const u_char  HEXADECIMAL[16] NXT_NONSTRING = "0123456789ABCDEF";
     static const u_char  nan[] = "[nan]";
     static const u_char  null[] = "[null]";
     static const u_char  infinity[] = "[infinity]";
--- a/src/nxt_string.c
+++ b/src/nxt_string.c
@@ -598,7 +598,7 @@ nxt_encode_uri(u_char *dst, u_char *src, size_t length)
     u_char      *end;
     nxt_uint_t  n;
 
-    static const u_char  hex[16] = "0123456789ABCDEF";
+    static const u_char  hex[16] NXT_NONSTRING = "0123456789ABCDEF";
 
     end = src + length;
 
@@ -644,7 +644,7 @@ nxt_encode_complex_uri(u_char *dst, u_char *src, size_t length)
     u_char      *reserved, *end, ch;
     nxt_uint_t  n;
 
-    static const u_char  hex[16] = "0123456789ABCDEF";
+    static const u_char  hex[16] NXT_NONSTRING = "0123456789ABCDEF";
 
     reserved = (u_char *) "?#\0";
 


ac000:
Cc: @alejandro-colomar

src/nxt_http_parse.c
```diff
@@ -164,8 +164,10 @@ nxt_http_parse_request_line(nxt_http_request_parse_t *rp, u_char **pos,
     nxt_http_ver_t           ver;
     nxt_http_target_traps_e  trap;
 
-    static const nxt_http_ver_t  http11 = { "HTTP/1.1" };
-    static const nxt_http_ver_t  http10 = { "HTTP/1.0" };
```
hongzhidao, line 168: If we don't touch them, can it work?

ac000: No, because you can't apply the attribute to a union.

It needs the attribute because the char member of the union is 8 bytes and it doesn't want the `\0`. We rely on the fact it's 8 bytes as that also happens to be the size of a `uint64_t` which is another member of that union and is what we use to do an *optimised* string compare.

If this was being done in multiple places, I'd consider hiding it in a macro, but it's just this one location we do this...

In some ways splitting it out into the two parts does make it clearer what's going on and then you don't need to grasp the link between `char[8]` and `uint64_t` and mistakenly thinking to make it `char[9]` which will break things.

Other options are

```c
    static const nxt_http_ver_t  http11 =
                            {{ 'H', 'T', 'T', 'P', '/', '1', '.', '1' }};
    static const nxt_http_ver_t  http10 =
                            {{ 'H', 'T', 'T', 'P', '/', '1', '.', '0' }};
```

```c
nxt_http_ver_t  http11, http10;

memcpy(&http11, "HTTP/1.1", 8);
memcpy(&http10, "HTTP/1.0", 8);
```

Think I still prefer what's in the patch...

alejandro-colomar: Can't you tag the union member as being a nonstring?

```c
alx@devuan:~/tmp$ cat union.c 
union u {
	[[gnu::nonstring]] char  ns[8];
	long  l;
};
alx@devuan:~/tmp$ gcc -Wall -Wextra -S union.c 
alx@devuan:~/tmp$ 
```

auto/clang
```diff
@@ -194,3 +194,17 @@ nxt_feature_test="static void f(void) __attribute__ ((__unused__));
...
+nxt_feature_name=NXT_HAVE_GCC_ATTRIBUTE_NONSTRING
+nxt_feature_run=
+nxt_feature_incs=
+nxt_feature_libs=
+nxt_feature_test="int main(void) {
+                      static const char str[3] __attribute__ ((__nonstring__)) =
+                          \"ABC\";
+
+                      return !!str[0];
+                  }"
```
alejandro-colomar, line 209: For commit 1:

```
Reviewed-by: Alejandro Colomar <alx@kernel.org>
```

As a minor note, for consistency with nginx style, there should be two spaces between `char` and `str`, I think.

Another minor note:  it is more common not having a space after `__attribute__`, although the project would benefit of some more consistency.

```c
$ grep -Trno '__attribute__ *('
auto/clang: 122:	__attribute__ (
auto/clang: 135:	__attribute__ (
auto/clang: 150:	__attribute__ (
auto/clang: 173:	__attribute__ (
auto/clang: 186:	__attribute__ (
src/nxt_clang.h:  11:	__attribute__(
src/nxt_clang.h:  12:	__attribute__(
src/nxt_clang.h:  81:	__attribute__(
src/nxt_clang.h:  92:	__attribute__(
src/nxt_clang.h: 103:	__attribute__(
src/nxt_clang.h: 114:	__attribute__(
src/nxt_clang.h: 125:	__attribute__(
src/nxt_unit.h:  344:	__attribute__(
```

auto/cc/test
```diff
@@ -93,10 +93,6 @@ case $NXT_CC_NAME in
 
         NXT_CFLAGS="$NXT_CFLAGS -Wmissing-prototypes"
 
-        # Disable Wunterminated-string-initialization temporarily.
-        # See <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117178#c21>
-        NXT_CFLAGS="$NXT_CFLAGS -Wno-unterminated-string-initialization"
-
```
alejandro-colomar, line 99: For commit 3:

```
Reviewed-by: Alejandro Colomar <alx@kernel.org>
```

Nice!

auto/clang
ac000: > As a minor note, for consistency with nginx style, there should be two spaces between char and str, I think.

Then again for these tests we do things like

```c
int main(void) {
```

(Curly brace placement)

So I'm not too bothered for these wee things.... and as it is it just fits in 80 chars...

> Another minor note: it is more common not having a space after __attribute__

Yeah.

> although the project would benefit of some more consistency.

I just tried to be consistent with the surrounding code.

alejandro-colomar: > > As a minor note, for consistency with nginx style, there should be two spaces between char and str, I think.
> 
> Then again for these tests we do things like
> 
> ```c
> int main(void) {
> ```
> 
> (Curly brace placement)
> 
> So I'm not too bothered for these wee things.... and as it is it just fits in 80 chars...

Makes sense.

> 
> > Another minor note: it is more common not having a space after **attribute**
> 
> Yeah.
> 
> > although the project would benefit of some more consistency.
> 
> I just tried to be consistent with the surrounding code.

Ok.  :-)

src/nxt_http_parse.c
alejandro-colomar: :-}

ac000:
@hongzhidao 

Thanks to @alejandro-colomar pointing out my idiocy, your qualms about the `nxt_http_ver_t` union initialisation changes should be neutralised, I.e. they don't change now.

```
$ git range-diff 6a272400...ca605361
1:  52529aa2 ! 1:  764ad73f auto/clang: Add a NXT_NONSTRING macro
    @@ Commit message
         Link: <https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=622968990beee7499e951590258363545b4a3b57>
         Cc: Alejandro Colomar <alx@kernel.org>
    +    Reviewed-by: Alejandro Colomar <alx@kernel.org>
     
      ## auto/clang ##
2:  34ae3a0f ! 2:  b53d03c2 Tag various character arrays with NXT_NONSTRING
    @@ Commit message
         happening.
     
         Cc: Alejandro Colomar <alx@kernel.org>
    +    Co-authored-by: Alejandro Colomar <alx@kernel.org>
    +    Signed-off-by: Alejandro Colomar <alx@kernel.org>
     
      ## src/nxt_http_parse.c ##
    -@@ src/nxt_http_parse.c: nxt_http_parse_request_line(nxt_http_request_parse_t *rp, u_char **pos,
    -     nxt_http_ver_t           ver;
    -     nxt_http_target_traps_e  trap;
    - 
    --    static const nxt_http_ver_t  http11 = { "HTTP/1.1" };
    --    static const nxt_http_ver_t  http10 = { "HTTP/1.0" };
    -+    static const u_char          http11_str[8] NXT_NONSTRING = "HTTP/1.1";
    -+    static const u_char          http10_str[8] NXT_NONSTRING = "HTTP/1.0";
    -+    static const nxt_http_ver_t  *http11 = (nxt_http_ver_t *)http11_str;
    -+    static const nxt_http_ver_t  *http10 = (nxt_http_ver_t *)http10_str;
    - 
    -     p = *pos;
    - 
    -@@ src/nxt_http_parse.c: nxt_http_parse_request_line(nxt_http_request_parse_t *rp, u_char **pos,
    - 
    -     nxt_memcpy(ver.str, &p[1], 8);
    - 
    --    if (nxt_fast_path(ver.ui64 == http11.ui64
    --                      || ver.ui64 == http10.ui64
    -+    if (nxt_fast_path(ver.ui64 == http11->ui64
    -+                      || ver.ui64 == http10->ui64
    -                       || (memcmp(ver.str, "HTTP/1.", 7) == 0
    -                           && ver.s.minor >= '0' && ver.s.minor <= '9')))
    -     {
     @@ src/nxt_http_parse.c: nxt_http_parse_field_name(nxt_http_request_parse_t *rp, u_char **pos,
          size_t    len;
          uint32_t  hash;
    @@ src/nxt_http_parse.c: nxt_http_parse_field_name(nxt_http_request_parse_t *rp, u_
          /*   \s ! " # $ % & ' ( ) * + ,        . /                 : ; < = > ?   */
              "\0\1\0\1\1\1\1\1\0\0\1\1\0" "-" "\1\0" "0123456789" "\0\0\0\0\0\0"
     
    + ## src/nxt_http_parse.h ##
    +@@ src/nxt_http_parse.h: typedef struct nxt_http_fields_hash_s    nxt_http_fields_hash_t;
    + 
    + 
    + typedef union {
    +-    u_char                    str[8];
    ++    u_char                    str[8] NXT_NONSTRING;
    +     uint64_t                  ui64;
    + 
    +     struct {
    +
      ## src/nxt_sprintf.c ##
     @@ src/nxt_sprintf.c: nxt_vsprintf(u_char *buf, u_char *end, const char *fmt, va_list args)
          nxt_sprintf_t        spf;
3:  6a272400 ! 3:  ca605361 auto/cc: gcc: Don't disable -Wunterminated-string-initialization
    @@ Commit message
     
         Fixes: 150378224 ("Fix build with GCC 15")
         Cc: Alejandro Colomar <alx@kernel.org>
    +    Reviewed-by: Alejandro Colomar <alx@kernel.org>
     
      ## auto/cc/test ##
```

src/nxt_http_parse.h
```diff
@@ -21,7 +21,7 @@ typedef struct nxt_http_fields_hash_s    nxt_http_fields_hash_t;
 
 
 typedef union {
-    u_char                    str[8];
+    u_char                    str[8] NXT_NONSTRING;
     uint64_t                  ui64;
 
     struct {
```
alejandro-colomar, line 27: Hi @ac000 ,

Another detail:

C23 attributes are more strict regarding the placement of the attribute than the old GNU attributes.  GNU is being benevolent, and figuring out where you want the attribute to apply.

Here's an example program:

```c
alx@devuan:~/tmp$ cat attr.c | grep -Tn ^
                  1:	int
                  2:	main(void)
                  3:	{
                  4:	[[gnu::nonstring]] char                    n0                    [8];
                  5:	                   char [[gnu::nonstring]] n1                    [8];
                  6:	                   char                    n2 [[gnu::nonstring]] [8];
                  7:	                   char                    n3                    [8] [[gnu::nonstring]];
                  8:
                  9:	[[gnu::nonstring]] char                    m0                    [8],                    x0;
                 10:	                   char [[gnu::nonstring]] m1                    [8],                    x1;
                 11:	                   char                    m2 [[gnu::nonstring]] [8],                    x2;
                 12:	                   char                    m3                    [8] [[gnu::nonstring]], x3;
                 13:	}
```

The attribute can be placed in 4 positions, and each of them means a different thing.

In lines 4 and 9, it means it applies to all variables declared in the line (so, `n0`, `m0`, and `x0`).
In lines 5 and 10, it means it applies to the thing immediately to the left, which is the type `char`.
In lines 6 and 11, it means it applies to the thing immediately to the left, which are the arrays `n2` and `m2`.
In lines 7 and 12, it means it applies to the thing immediately to the left, which is the type `char [8]`.

In general, the rule is that if there's something immediately to the left of the attribute, it applies to it.  If the attribute is the left-most thing, it applies to everything.

```
alx@devuan:~/tmp$ gcc -Wattributes -S attr.c 
attr.c: In function ‘main’:
attr.c:5:20: warning: ‘nonstring’ attribute does not apply to types [-Wattributes]
    5 |                    char [[gnu::nonstring]] n1                    [8];
      |                    ^~~~
attr.c:7:20: warning: ‘nonstring’ attribute does not apply to types [-Wattributes]
    7 |                    char                    n3                    [8] [[gnu::nonstring]];
      |                    ^~~~
attr.c:9:1: warning: ‘nonstring’ attribute ignored on objects of type ‘char’ [-Wattributes]
    9 | [[gnu::nonstring]] char                    m0                    [8],                    x0;
      | ^
attr.c:10:20: warning: ‘nonstring’ attribute does not apply to types [-Wattributes]
   10 |                    char [[gnu::nonstring]] m1                    [8],                    x1;
      |                    ^~~~
attr.c:12:20: warning: ‘nonstring’ attribute does not apply to types [-Wattributes]
   12 |                    char                    m3                    [8] [[gnu::nonstring]], x3;
      |                    ^~~~
```

It might be interesting to be pedantic about these rules, for easing portability in the future to C23 attribute syntax.
For readability, I tend to add a newline after the attribute, to avoid having to move everything too far to the right.

alejandro-colomar: This is (confusingly) described in <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3467.pdf#subsubsection.0.6.7.7.1>, I think.

ac000: TBH, I'm not too concerned about C23 right now, it will be some time (if ever) we move to that...

Seeing as we're using __attribute__, I'll just stick to how we're currently doing placements. IFF we ever do move to C23 we can look at doing a wholesale cleanup of the attributes then...

The NXT_NOPNSTRING placements match how we currently do them **and** the GCC documentation... e.g. for the "nonstring" example

```c
struct Data
{
  char name [32] __attribute__ ((nonstring));
};
```

ac000: Oh and gcc with -std=gnu23 seems happy with the above also...

ac000: I guess the stricter placements only apply to the C++ style attributes?

ac000:
```c
$ cat nonstring.c
#if (defined(__GNUC__) && !defined(__clang__))
#define __nonstring	__attribute__((nonstring))
#else
#define __nonstring
#endif

union u {
	char s[8] __nonstring;
	long long l;
};

int hexit(int c)
{
	static const char hex[16] __nonstring = "0123456789ABCDEF";

	return hex[c % 16];
}

int f(void)
{
	static const union u u = { "HTTP.1.1" };

	return u.l == 0;
}
```

```console
$ gcc -std=gnu23 -Wall -Wextra -Wattributes -c nonstring.c 
$
```

All good in the hood...

alejandro-colomar: > I guess the stricter placements only apply to the C++ style attributes?

Yes, GNU attributes are older and very arbitrary/obscure/benevolent on the placement rules.  They didn't change them, to avoid breaking old code.

alejandro-colomar: BTW, never mix GNU attributes and C23 attributes.  That's another problem of its own.  (But I guess it's not a real issue in this project, since you will not use C23 in a decade at least.)  :)

alejandro-colomar (approved):

hongzhidao (approved):

src/nxt_http_parse.h
alejandro-colomar: <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=108796>

ac000:
Fix a commit message.

```
$ git range-diff ca605361...ec1608b4
1:  b53d03c2 ! 1:  d9c2fd79 Tag various character arrays with NXT_NONSTRING
    @@ Commit message
     
         we no longer get the warning.
     
    -    We need to split the funky union initialisation in
    -    nxt_http_parse_request_line() into two parts, initialise the HTTP
    -    version string arrays then use them to initialise the unions, as you
    -    can't apply the attribute to non char base types.
    -
    -    It could be argued that this actually makes it clearer what is
    -    happening.
    -
         Cc: Alejandro Colomar <alx@kernel.org>
         Co-authored-by: Alejandro Colomar <alx@kernel.org>
2:  ca605361 = 2:  ec1608b4 auto/cc: gcc: Don't disable -Wunterminated-string-initialization
```

PR #1598, opened 2025-04-21
http: Compress application responses

ac000:
```
    http: Compress application responses
    
    This adds initial support for compressing application responses.
    
    A couple of things to note
    
    1) Compressed responses are sent 'chunked' as we don't know beforehand
       how large the compressed response will be.
    
    2) We only compress responses where we know the Content-Length as we
       need to check with the 'min_length' config parameter. It's also
       currently how we track when we need to close the compression stream
       off.
    
    Co-authored-by: Alejandro Colomar <alx@kernel.org>
```


hongzhidao (approved):

ac000:
Thanks Zhidao!

PR #1619, opened 2025-05-21
Use NULL instead of 0

ac000:
Use NULL instead of 0
    
    GCC 15 introduced a new warning "-Wzero-as-null-pointer-constant" which
    checks for places where '0' has been used in place of the NULL macro.
    
    This showed a few places in Unit where we were using '0' instead of the
    more correct NULL. E.g.
    
      $ make -j4 EXTRA_CFLAGS=-Wzero-as-null-pointer-constant
      ...
      src/nxt_buf.c: In function ‘nxt_buf_mmap_alloc’:
      src/nxt_buf.h:192:21: error: zero as null pointer constant [-Werror=zero-as-null-pointer-constant]
        192 |         (bm)->start = 0;                                                      \
            |                     ^
      src/nxt_buf.c:135:9: note: in expansion of macro ‘nxt_buf_mem_set_size’
        135 |         nxt_buf_mem_set_size(&b->mem, size);
            |         ^~~~~~~~~~~~~~~~~~~~
    

ac000:
Ignore the failing the Clang-AST test, it needs https://github.com/nginx/unit/pull/1620

PR #1632, opened 2025-08-11
A couple of HTTP compression fixes

ac000:
```
This pull-request comprises two commits

1) http: compression: Set the temporary file name in n_h_c_c_s_r()

The sets the temporary compressed file name in the nxt_file_t structure.

While we don't actually need it ourselves, it's used in the case that
nxt_file_close() fails.

2) http: compression: Add a missed nxt_http_comp_compress() return check

This adds a missing check for compression errors when compressing static
responses.

----------------------------------------------------------------
Andrew Clayton (2):
      http: compression: Set the temporary file name in n_h_c_c_s_r()
      http: compression: Add a missed nxt_http_comp_compress() return check

 src/nxt_http_compression.c | 26 ++++++++++++++++----------
 src/nxt_http_compression.h |  4 ++--
 src/nxt_http_static.c      |  2 +-
 3 files changed, 19 insertions(+), 13 deletions(-)
```

### DIFF
--- a/src/nxt_http_compression.c
+++ b/src/nxt_http_compression.c
@@ -305,6 +305,12 @@ nxt_http_comp_compress_static_response(nxt_task_t *task, nxt_http_request_t *r,
 
         cbytes = nxt_http_comp_compress(out + *out_total, out_size - *out_total,
                                         in + in_size - rest, n, last);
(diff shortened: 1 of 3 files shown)


hongzhidao (approved):
Look good.

PR #1635, opened 2025-08-12
Don't leak file descriptor in nxt_main_port_access_log_handler()

ac000:
```
Don't leak file descriptor in nxt_main_port_access_log_handler()

After opening a file and setting file.fd we _may_ call
nxt_port_socket_write(). If so then the file is eventually closed via
something like

  nxt_port_socket_write()
    nxt_port_socket_write2()
      nxt_port_write_handler()
        nxt_port_msg_close_fd()
          nxt_port_close_fds()

Alternatively we may just return from the function and never close(2)
file.fd.

In which case we should call nxt_file_close().

This was reported by coverity.

```


hongzhidao (approved):
LGTM.

